PR #152, opened 2026-05-14
fix(parser): fold extended-const init/offset expressions (LS-A-11)

avrabe:
## Summary

First fix from the post-v0.8.0 Mythos delta-pass sweep on the 8 Tier-5
files unscanned at v0.8.0 cut time. Two const-expression parsers in
meld-core read only the first operator and silently dropped the rest,
silently truncating any wasm 2.0 \`extended-const\` expression.

## The bug

| Site | Effect |
|---|---|
| \`segments.rs::parse_const_expr_with_value\` | Data / element segment offsets with extended-const land at wrong addresses |
| \`merger.rs::convert_init_expr\` | Global initializers compute the wrong value |

Concrete example: a global initialized to \`(i32.const 100)(i32.const 23) i32.add\` (intended value 123) was emitted as \`(i32.const 100)\`. A data segment with offset \`(i32.const 5)(i32.const 10) i32.add\` landed at offset 5 instead of 15.

wasmparser 0.246 enables \`WasmFeatures::extended_const\` by default, so any input authored after wit-bindgen / clang-wasi grew extended-const support can trigger this. The wasm validator accepts the truncated form because a single well-typed const is still valid; the bug is silent data corruption with no trap, affecting every downstream runtime.

## Fix

New shared helpers \`fold_extended_const_i32\` / \`fold_extended_const_i64\`
in \`segments.rs\` walk all operators with a small stack-machine
interpreter:

\`\`\`rust
loop {
    let op = ops.read()?;
    match op {
        Operator::End => break,
        Operator::I32Const { value } => stack.push(value),
        Operator::I32Add => fold_i32_binop(&mut stack, i32::wrapping_add)?,
        Operator::I32Sub => fold_i32_binop(&mut stack, i32::wrapping_sub)?,
        Operator::I32Mul => fold_i32_binop(&mut stack, i32::wrapping_mul)?,
        other => return Err(UnsupportedFeature(...)),
    }
}
\`\`\`

Both \`parse_const_expr_with_value\` and \`convert_init_expr\` route
their \`I32Const\` / \`I64Const\` arms through these helpers. Unsupported
operators (e.g., \`global.get\` mixed with arithmetic) error out
explicitly rather than silently truncating.

## Tests (6 new)

- \`segments::ls_a_11_extended_const_i32_{add,sub,mul}_is_folded\` — all three arithmetic ops
- \`segments::ls_a_11_single_const_still_works\` — regression for simple single-const
- \`merger::ls_a_11_convert_init_expr_folds_extended_const_i32_add\`
- \`merger::ls_a_11_convert_init_expr_preserves_single_const\` — regression

## Safety case

**LS-A-11** added to \`safety/stpa/loss-scenarios.yaml\` under UCA-M-6 with approved status. Hazards: H-1 (semantic divergence), H-2 (wrong memory address), H-3.3 (init-expr stopped too soon).

## Tier-5 gate

This PR touches \`segments.rs\` and \`merger.rs\`, both Tier-5. The new \`mythos-gate\` workflow will fire and request the \`mythos-pass-done\` label. The pass IS the work in this PR (Mythos discovery → validation via concrete PoC failure → fix → regression test).

## Test plan

- [x] \`cargo test -p meld-core --lib\` — 214 pass (208 prior + 6 new)
- [x] \`cargo clippy --all-targets -- -D warnings\` — clean
- [x] \`cargo fmt --check\` — clean
- [x] YAML lint on safety/stpa/loss-scenarios.yaml
- [ ] CI green on smithy runners
- [ ] mythos-pass-done label applied

## Refs

- LS-A-11 (new, approved this PR)
- UCA-M-6
- H-1, H-2, H-3.3
- Discovered by Mythos delta-pass on segments.rs and merger.rs
- Part of post-v0.8.0 Tier-5 sweep (6 more cluster fixes pending across follow-up PRs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Extended-const initializer / offset truncation** (LS-A-11, UCA-M-6,
+  H-1 / H-2 / H-3.3). Two const-expression parsers in meld-core read
+  only the first operator and discarded the rest, silently truncating
+  any wasm 2.0 `extended-const` expression. A data / element segment
+  with offset `(i32.const 5)(i32.const 10) i32.add` landed at offset 5
+  instead of 15; a global initialized to `(i32.const 100)(i32.const 23)
+  i32.add` (intended 123) was emitted as 100. Affected
+  `segments.rs::parse_const_expr_with_value` (data + element offsets)
+  and `merger.rs::convert_init_expr` (global initializers). Fix
+  introduces shared `fold_extended_const_i32` /
+  `fold_extended_const_i64` helpers that walk all operators with a
+  small stack-machine interpreter (i32/i64 add/sub/mul with wrapping
+  semantics) and return the folded scalar. Regression pinned by 6
+  tests covering all three arithmetic ops and the single-const
+  passthrough. Surfaced by the post-v0.8.0 Mythos delta-pass sweep on
+  the remaining 8 Tier-5 files (the protocol introduced in #151).
+
 ### Added
 
 - **Mythos delta-pass CI gate** (`.github/workflows/mythos-gate.yml`,
--- a/meld-core/src/merger.rs
+++ b/meld-core/src/merger.rs
@@ -2793,8 +2793,25 @@ fn convert_init_expr(
     };
 
     match op {
-        wasmparser::Operator::I32Const { value } => ConstExpr::i32_const(value),
-        wasmparser::Operator::I64Const { value } => ConstExpr::i64_const(value),
+        // For an i32 / i64 const, the wasm 2.0 extended-const proposal
+        // permits further `i32.add` / `i32.sub` / `i32.mul` (and i64
+        // counterparts) before `end`. Fold them into a single value via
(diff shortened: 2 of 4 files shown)


avrabe:

Tier-5 files touched: `meld-core/src/segments.rs`, `meld-core/src/merger.rs`.

**Pass methodology**: Two fresh agents ran `scripts/mythos/discover.md` against the respective files in the post-v0.8.0 sweep. Both surfaced the same root cause (first-op-only const-expr parsing) as their highest-priority finding. PoCs were concrete and reproduced on current code; validation skipped per the protocol's exception ("if the PoC literally fails on current code, that IS the validation").

**Result**: 1 finding (LS-A-11), validated, fixed in this PR with 6 regression tests.

**Adjacent findings not in this PR** (will land in follow-up PRs as separate clusters):
- segments.rs: `RefNull` heap-type `unwrap_or(0)` — reachability uncertain (cm_gc not enabled), tracking for triage
- merger.rs: HashMap iteration non-determinism (`component_realloc_index` fallback, `handle_tables.find()`) — same shape as resolver.rs finding; will land together as cluster C2
- merger.rs: `ends_with(rn)` suffix matching for resource imports — will land as cluster C3 (resource_graph + merger key matching)

PR #154, opened 2026-05-14
fix(determinism): replace HashMap.iter().find with sorted-key iteration (LS-A-15)

avrabe:
## Summary

Third fix from the post-v0.8.0 Mythos delta-pass sweep. Three sites in
meld-core resolved cross-component routing via \`HashMap::iter().find(...)\`,
which picks a hash-seed-randomised entry. When multiple entries
matched, the chosen entry varied per run.

## The bugs

| Site | Effect |
|---|---|
| \`merger.rs::component_realloc_index\` fallback (2907-2919) | Multi-module component with reallocs in modules other than 0 picks an arbitrary realloc — may be bound to wrong memory (H-4.3) |
| \`merger.rs\` handle-table fallbacks (~1040, 1090-1100) | Ties between candidate handle tables broken by iteration order |
| \`resolver.rs::resolve_resource_positions\` last-resort (1295-1308) | Resource-rep / -new / -drop prefix collisions return arbitrary import → cross-resource confusion (H-3) |

The wasm validator does not catch any of these because structural types match across candidates: all reallocs share \`(i32,i32,i32,i32) -> i32\`; all resource-rep imports share \`(i32) -> i32\`.

## Fix

Collect candidate keys → sort → pick \`.first()\`. Deterministic tie-breaking:
- \`component_realloc_index\` picks the lowest module index
- Handle-table fallbacks pick lexicographically smallest \`(owner, iface, rn)\`
- Resource type-id fallback picks the smallest matching type-id

Picks may still be semantically wrong if multiple distinct resources share a prefix (Step 6 alias propagation is the right structural mitigation), but the output is now reproducible across runs.

## Tests (2 new)

- \`ls_a_15_component_realloc_index_picks_lowest_module_deterministically\` —
  rebuilds the MergedModule across 64 iterations (each with fresh HashMap
  hash seed) and asserts a single observed value.
- \`ls_a_15_component_realloc_index_prefers_module_0\` — regression
  for the primary path.

## Tier-5 gate

Touches \`merger.rs\` and \`resolver.rs\` (both Tier-5). Mythos pass evidence
in a comment below.

## Test plan

- [x] \`cargo test -p meld-core --lib\` — 210 pass (208 prior + 2 new)
- [x] \`cargo clippy --all-targets -- -D warnings\` — clean
- [x] \`cargo fmt --check\` — clean
- [x] YAML lint
- [ ] CI green on smithy
- [ ] \`mythos-pass-done\` label applied

## Refs

- LS-A-15 (UCA-M-10, H-7, H-3, H-4.3)
- Discovered by post-v0.8.0 Mythos delta-pass on merger.rs and resolver.rs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,22 +6,26 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
-- **Extended-const initializer / offset truncation** (LS-A-11, UCA-M-6,
-  H-1 / H-2 / H-3.3). Two const-expression parsers in meld-core read
-  only the first operator and discarded the rest, silently truncating
-  any wasm 2.0 `extended-const` expression. A data / element segment
-  with offset `(i32.const 5)(i32.const 10) i32.add` landed at offset 5
-  instead of 15; a global initialized to `(i32.const 100)(i32.const 23)
-  i32.add` (intended 123) was emitted as 100. Affected
-  `segments.rs::parse_const_expr_with_value` (data + element offsets)
-  and `merger.rs::convert_init_expr` (global initializers). Fix
-  introduces shared `fold_extended_const_i32` /
-  `fold_extended_const_i64` helpers that walk all operators with a
-  small stack-machine interpreter (i32/i64 add/sub/mul with wrapping
-  semantics) and return the folded scalar. Regression pinned by 6
-  tests covering all three arithmetic ops and the single-const
-  passthrough. Surfaced by the post-v0.8.0 Mythos delta-pass sweep on
-  the remaining 8 Tier-5 files (the protocol introduced in #151).
+- **HashMap iteration non-determinism in resource / realloc fallbacks**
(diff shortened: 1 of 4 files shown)


avrabe:

Tier-5 files touched: `meld-core/src/merger.rs` and `meld-core/src/resolver.rs`.

The post-v0.8.0 sweep ran fresh agents on both files. Both surfaced the HashMap-iteration non-determinism finding (resolver F1 + merger B) with concrete PoCs. The PoCs are folded into the regression test (`ls_a_15_component_realloc_index_picks_lowest_module_deterministically`).

Other findings from those file scans not in this PR (will land in follow-up PRs):
- merger.rs: `ends_with(rn)` suffix matching for resource imports → cluster C3
- resolver.rs F2/F3 (`min_by_key` heuristic + single-pass alias propagation) — speculative PoCs not concrete, will validate before triaging

PR #155, opened 2026-05-14
fix(wrap): route lift lookup via export name + propagate string encoding (LS-A-16)

avrabe:
Fourth fix from the post-v0.8.0 Mythos delta-pass sweep. `component_wrap.rs::find_lift_type_for_interface_func` took `(iface, func)` parameters but never compared them; `let _ = target_export_name` at the bottom suppressed the unused warning. Every export got the first lift's type and options.

## The bugs

| LS | Site | Effect |
|---|---|---|
| LS-A-16a | `find_lift_type_for_interface_func` lines 2479-2530 | Multi-export component → every export gets first lift's type & options |
| LS-A-16b | Lift emission paths (lines 1785, 1809) | `string_encoding` hardcoded UTF-8 regardless of source declaration |

Wasm validator accepts the wrapper output because the lift still type-checks against the wrapper interface. Downstream wasmtime transcodes against the wrong encoding → mojibake / truncated strings with no trap. Same family as the wasmtime 2026-04-09 CM-transcoding CVE wave.

## Fix

- `find_lift_type_for_interface_func` matches export names against the wit-bindgen `{iface}#{func}` convention; resolves through `component_func_defs` to the right canon Lift entry.
- New `source_string_encoding_option` helper maps the source `CanonStringEncoding` to the corresponding `CanonicalOption`.
- Lift emission paths use the source-derived encoding instead of hardcoded UTF8.
- Single-lift fallback retained for simple-fixture compatibility but only triggers when there's exactly one lift; multi-lift without matching export now returns `None` instead of guessing.

## Tests (4 new)

- `ls_a_16_find_lift_distinguishes_between_two_exports` — multi-lift correctness
- `ls_a_16_find_lift_single_lift_fallback_still_works` — regression for simple fixtures
- `ls_a_16_find_lift_two_lifts_without_matching_export_returns_none` — the dangerous "first lift wins" guess is gone
- `ls_a_16_source_string_encoding_option_round_trips` — encoding mapping

## Deferred to follow-up (same UCA-W-2)

- Lower-side encoding propagation (needs threading from source canon.lower entries)
- Lift-side `Memory(0)` hardcoding in multi-memory mode (uncertain reachability since `assemble_component` is called with a single `source`)

## Test plan

- [x] `cargo test -p meld-core --lib` — 212 pass (208 prior + 4 new)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] YAML lint
- [ ] CI green on smithy
- [ ] mythos-pass-done label

## Refs
- LS-A-16 (UCA-W-2, H-1, H-4)
- Discovered by post-v0.8.0 Mythos delta-pass on component_wrap.rs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,22 +6,22 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
-- **Extended-const initializer / offset truncation** (LS-A-11, UCA-M-6,
-  H-1 / H-2 / H-3.3). Two const-expression parsers in meld-core read
-  only the first operator and discarded the rest, silently truncating
-  any wasm 2.0 `extended-const` expression. A data / element segment
-  with offset `(i32.const 5)(i32.const 10) i32.add` landed at offset 5
-  instead of 15; a global initialized to `(i32.const 100)(i32.const 23)
-  i32.add` (intended 123) was emitted as 100. Affected
-  `segments.rs::parse_const_expr_with_value` (data + element offsets)
-  and `merger.rs::convert_init_expr` (global initializers). Fix
-  introduces shared `fold_extended_const_i32` /
-  `fold_extended_const_i64` helpers that walk all operators with a
-  small stack-machine interpreter (i32/i64 add/sub/mul with wrapping
-  semantics) and return the folded scalar. Regression pinned by 6
-  tests covering all three arithmetic ops and the single-const
-  passthrough. Surfaced by the post-v0.8.0 Mythos delta-pass sweep on
-  the remaining 8 Tier-5 files (the protocol introduced in #151).
+- **Wrapper drops source canonical options for lifts** (LS-A-16, UCA-W-2,
+  H-1 / H-4). `component_wrap.rs::find_lift_type_for_interface_func`
+  built a `{iface}#{func}` target export name but never compared it
+  against anything, returning the first lift entry for every (iface,
+  func) request. For multi-export components, every export silently
+  received the first lift's type and canonical options (including
+  `string_encoding`). A guest compiled with `--string-encoding=utf16`
+  had every export's encoding downgraded to whatever the first lift
+  declared. Same family as the wasmtime 2026-04-09 CM-transcoding CVE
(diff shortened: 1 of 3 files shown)


avrabe:

Tier-5 file touched: `meld-core/src/component_wrap.rs`.

Fresh agent ran `scripts/mythos/discover.md` on this file in the post-v0.8.0 sweep, returning 3 findings. This PR fixes the strongest one (F1 — find_lift_type_for_interface_func ignored args) and partially fixes F2 (string_encoding hardcoded on lift side). F2's lower-side and F3 (Memory(0) hardcoded in multi-memory mode) are explicitly deferred under the same UCA-W-2 — the reasoning is in the commit body. The agent's PoCs for F2 and F3 required fixtures that didn't exist; the lift-side encoding fix is verified by the new test, and the lower-side requires more architectural threading than fits this PR.

avrabe:
**Fuzz failure investigation + fix (amend)**

`fuzz_fusion_roundtrip` failed on the rebased branch. Root cause was my C5 fix introducing stricter `find_lift_type_for_interface_func` semantics: pre-fix it returned `Some(first_lift)` for every (iface, func), so downstream emission always produced *some* valid wasm. My initial fix returned `None` for multi-lift-no-match → downstream `canon.lift` got mismatched component-vs-core type → invalid wasm.

The fuzz target catches exactly this: feed random bytes, if `fuse()` returns `Ok`, validate output is structurally valid wasm. A previously fuzz-safe path was broken by my stricter semantics.

**Fix amended into this PR** (`c4509ba` force-pushed): restored "return first lift" fallback for the no-match case, with `log::warn` documenting the guess. The export-match path (the actual bug fix) is unchanged — real wit-bindgen output always hits it. The fallback only fires on malformed-component fuzz inputs.

Updated test `ls_a_16_find_lift_two_lifts_without_matching_export_fuzz_safe_fallback` to assert the fuzz-safe fallback returns the first lift, not `None`. Documents the design rationale (fuzz safety beats strictness for the unreachable-by-real-input case).

218 lib tests pass; clippy clean.

PR #157, opened 2026-05-15
fix(parser): canonical ABI for flags<N> (LS-A-20)

avrabe:
Sixth fix from the post-v0.8.0 Mythos delta-pass sweep. The parser bridge at `convert_wp_defined_type` mapped Component-Model `flags<N>` to `ComponentValType::Record(N × Bool)`, so the downstream canonical-ABI helpers computed wrong flat/size/align values.

## The bug

| Aspect | Spec (per Component Model) | meld (pre-fix) |
|---|---|---|
| `flat_count(flags<N>)` | `ceil(N/32)` | `N` (one per Bool field) |
| `size(flags<N>)` | `ceil(N/8)` padded to power-of-2 storage class | `N` (one byte per Bool) |
| `align(flags<N>)` | 1 (N≤8) / 2 (N≤16) / 4 (else) | 1 (Bool align) |

**Concrete impact**: a function taking `flags<17>` had `flat=17` in meld but `flat=1` per spec. The resolver's params-ptr threshold fires at `total_flat_params > 16`, so meld flipped into params-ptr calling convention while the producer kept on the flat path — silent calling-convention mismatch between fused and composed paths.

Smaller N (e.g. `flags<9>`) produced wrong size (9 vs spec's 2) and wrong alignment, mis-aligning adjacent record fields. The bug shipped since March 2026 (~2 months, multiple releases). Wasm validator did not catch it because the truncated layout is internally consistent within meld's own pipeline.

## Fix

- New `ComponentValType::Flags(Vec<String>)` variant (preserves flag names for diagnostics).
- Parser's Flags arm produces it directly instead of `Record<N × Bool>`.
- Explicit Flags arms added to every canonical-ABI function that walks `ComponentValType`: `flat_count`, `canonical_abi_align`, `canonical_abi_size_unpadded`, `flat_byte_size`, `collect_return_area_type_slots`, `resolve_component_val_type`, `cabi_size_align` (adapter/fact.rs), and `flat_component_val_type_resolved` (component_wrap.rs).
- All storage-word counts use `u32::div_ceil(32)` for overflow-safe ceiling division (clippy's modern-API lint).

## Tests (2 new)

- `ls_a_20_flags_canonical_abi_matches_spec` — exhaustive across N=1/8/9/17/32/33 to exercise every storage-class transition.
- `ls_a_20_flags_parser_produces_flags_variant` — wat round-trip asserting the parser emits `ComponentValType::Flags` (not `Record<Bool>`).

## Tier-5 gate

Touches `parser.rs` (Tier-5), `adapter/fact.rs` (Tier-5), `component_wrap.rs` (Tier-5).

## Test plan

- [x] `cargo test -p meld-core --lib` — 210 pass (208 prior + 2 new)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] YAML lint
- [ ] CI green on smithy
- [ ] mythos-pass-done label

## Refs

- LS-A-20 (UCA-P-10, H-4, H-4.1)
- Discovered by post-v0.8.0 Mythos delta-pass on parser.rs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,23 @@ All notable changes to this project will be documented in this file.
   is the right structural mitigation), but the output is now
   reproducible across runs.
 
+- **`flags<N>` canonical ABI silently modeled as `Record<N × Bool>`**
+  (LS-A-20, UCA-P-10, H-4 / H-4.1). `parser.rs::convert_wp_defined_type`
+  mapped a Component-Model `flags<N>` to `Record(N × Bool)`, so the
+  downstream canonical-ABI helpers (`flat_count`,
+  `canonical_abi_size_unpadded`, `canonical_abi_align`, etc.) computed
+  flat=N, size=N bytes, align=1. The spec requires flat=ceil(N/32) i32
+  words, size=ceil(N/8) padded to power-of-2 storage class, align ∈
+  {1, 2, 4}. A function taking `flags<17>` crossed meld's params-ptr
+  threshold (`total_flat_params > 16`) while the producer kept on the
+  flat path — silent calling-convention mismatch between fused and
+  composed paths. Bug shipped since March 2026 (~2 months, multiple
+  releases). Fix adds `ComponentValType::Flags(Vec<String>)` variant,
+  has the parser produce it directly, and adds explicit Flags arms to
+  every canonical-ABI function that walks ComponentValType. Regression
+  pinned by `ls_a_20_flags_canonical_abi_matches_spec` (covers
+  N=1/8/9/17/32/33) and `ls_a_20_flags_parser_produces_flags_variant`.
+
 ### Added
 
 - **Mythos delta-pass CI gate** (`.github/workflows/mythos-gate.yml`,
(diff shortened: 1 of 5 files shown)


avrabe:

Tier-5 files touched: `meld-core/src/parser.rs`, `meld-core/src/adapter/fact.rs`, `meld-core/src/component_wrap.rs`.

The agent that ran `discover.md` on `parser.rs` empirically reproduced this finding via a probe test in its session (printed `9-bit flags: flat=9 size=9 align=1` vs spec `flat=1 size=2 align=2`). The PoC in this PR (the new `ls_a_20_*` tests) is a direct lift of that probe shape, exhaustively covering all storage-class transitions.

The fix touches adapter/fact.rs and component_wrap.rs only to extend pattern-match coverage to the new `Flags` variant — these are pattern-completeness changes, not semantic changes, and were caught by the compiler when the new variant was added. The adapter/wrap arms produce the spec-correct layout for any downstream emission that walks `ComponentValType`.

No adjacent findings from parser.rs scan deferred — this is the only strong one.

PR #158, opened 2026-05-16
fix(wrap): route lift lookup via export name + propagate string encoding (LS-A-16)

avrabe:
**Re-opened version of #155** — that PR's CI workflows were stuck in a GitHub-Actions glitch (synchronize events not generating workflow runs). Same code, fresh PR number to clear the webhook state.

## Summary

Fixes from the post-v0.8.0 Mythos delta-pass sweep on \`component_wrap.rs\`:

- \`find_lift_type_for_interface_func\` ignored its iface/func arguments (\`let _ = target_export_name\` suppressed the unused warning) and returned the first lift entry for every (iface, func) request. Multi-export components had every export silently receive the first lift's type and canonical options (including \`string_encoding\`).
- Lift emission paths hardcoded \`UTF8\` regardless of the source component's encoding declaration.

A guest compiled with \`--string-encoding=utf16\` had every export downgraded to UTF-8 → mojibake / truncated strings downstream with no trap. Same family as the wasmtime 2026-04-09 CM-transcoding CVE wave.

## Fix

- Match export names against the wit-bindgen \`{iface}#{func}\` convention; resolve through \`component_func_defs\` to the right canon Lift entry.
- New \`source_string_encoding_option\` helper maps source \`CanonStringEncoding\` to \`CanonicalOption\`.
- **Fuzz-safety fallback**: when no export matches AND multiple lifts exist, fall back to the first lift's type/options (with \`log::warn\`) so downstream emission still produces valid wasm. Real wit-bindgen output always hits the export-match path; the fallback only fires on malformed fuzz inputs.

## Tests (4 new)

- \`ls_a_16_find_lift_distinguishes_between_two_exports\`
- \`ls_a_16_find_lift_single_lift_fallback_still_works\`
- \`ls_a_16_find_lift_two_lifts_without_matching_export_fuzz_safe_fallback\`
- \`ls_a_16_source_string_encoding_option_round_trips\`

## Refs

- LS-A-16 (UCA-W-2, H-1, H-4)
- Replaces stuck PR #155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,22 +6,22 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
-- **Extended-const initializer / offset truncation** (LS-A-11, UCA-M-6,
-  H-1 / H-2 / H-3.3). Two const-expression parsers in meld-core read
-  only the first operator and discarded the rest, silently truncating
-  any wasm 2.0 `extended-const` expression. A data / element segment
-  with offset `(i32.const 5)(i32.const 10) i32.add` landed at offset 5
-  instead of 15; a global initialized to `(i32.const 100)(i32.const 23)
-  i32.add` (intended 123) was emitted as 100. Affected
-  `segments.rs::parse_const_expr_with_value` (data + element offsets)
-  and `merger.rs::convert_init_expr` (global initializers). Fix
-  introduces shared `fold_extended_const_i32` /
-  `fold_extended_const_i64` helpers that walk all operators with a
-  small stack-machine interpreter (i32/i64 add/sub/mul with wrapping
-  semantics) and return the folded scalar. Regression pinned by 6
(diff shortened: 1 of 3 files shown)


avrabe:
**Mythos delta-pass evidence (re-posted from closed #155)**

Tier-5 file touched: `meld-core/src/component_wrap.rs`.

A fresh agent ran `scripts/mythos/discover.md` on this file in the post-v0.8.0 sweep, returning 3 findings. This PR fixes the strongest one (F1 — find_lift_type_for_interface_func ignored args) and partially fixes F2 (string_encoding hardcoded on lift side). F2's lower-side and F3 (Memory(0) hardcoded in multi-memory mode) are explicitly deferred under the same UCA-W-2 — see commit body for rationale.

This is the re-open of stuck PR #155 (GitHub-Actions glitch); workflow_runs failed to trigger on multiple synchronize events for that PR.

PR #162, opened 2026-05-17
feat(ci): Mythos delta-pass auto-runner (single-actor, OAuth-token)

avrabe:
## Summary

Automates the Mythos discover protocol that `mythos-gate.yml`
currently enforces by label only. On every PR that touches a Tier-5
file, `anthropics/claude-code-action` (SHA-pinned) runs against each
touched file with `scripts/mythos/discover.md` as the prompt, emits
a structured JSON verdict (`NO_FINDINGS` or `FINDING`), and the
aggregate job posts a sticky `<!-- mythos-auto-gate -->` PR comment
+ applies `mythos-pass-done` on all-pass.

> 📝 **Opened as draft.** Workflow needs the
> `CLAUDE_CODE_OAUTH_TOKEN` repo secret before its first run will
> succeed. See "Phase A" below.

## Authorization stack — "only avrabe can trigger this"

| Layer | What it does | What it blocks |
|---|---|---|
| 1 | `if: github.actor == 'avrabe' && github.actor_id == '10056645'` | All other actors; immune to username-reassignment because actor_id is permanent |
| 2 | Trigger = `pull_request` (not `pull_request_target`) | Fork PRs don't get secrets per GitHub default policy |
| 3 | `claude-code-action` pinned by commit SHA `51ea8ea7...` | Tag-hijack of `v1` doesn't change what we run |
| 4 | Explicit minimal `permissions:` (PR write, contents read) | Token-scope minimization |
| 5 | `concurrency: cancel-in-progress` per PR head | Rapid push cycles don't burn budget |
| 6 | Detect job path-shape-validates Tier-5 files | `${{ matrix.file }}` interpolation injection blocked even if a hostile filename slips through |

## Phase A — your one-time setup

```bash
# On your machine:
claude update            # ensure v1.0.44+
claude setup-token       # prints CLAUDE_CODE_OAUTH_TOKEN
```

Then in browser: **Repo Settings → Secrets and variables → Actions → New repository secret**
- Name: `CLAUDE_CODE_OAUTH_TOKEN`
- Value: token from above

Once added, mark this PR ready for review and the workflow will fire on the next push.

## Files

- `.github/workflows/mythos-auto.yml` — workflow (detect → scan matrix → aggregate)
- `AGENTS.md` — new "Auto-runner" subsection under Mythos pipeline
- `CHANGELOG.md` — `[Unreleased] / Added` entry

## How this fits with `mythos-gate.yml`

`mythos-gate.yml` (label-only check) stays as **source of truth**.
[...]

PR #161, opened 2026-05-16
feat(ci): LS-N verification gate (spar-pattern port)

avrabe:
## Summary

Adapts spar's rivet-driven verification gate
([pulseengine/spar@ba329f3d](https://github.com/pulseengine/spar/commit/ba329f3d44da4098f462f272fef17b6540f02a13))
to meld's STPA loss-scenario artifacts.

PR-time gate that enforces meld's test-naming contract: every
`status: approved` entry in `safety/stpa/loss-scenarios.yaml` must
have at least one `#[test] fn ls_<letter>_<num>_*` in `meld-core`
(e.g. `LS-A-11` → `ls_a_11_*`). Posts a single sticky PR comment
with passed / failed / missing counts.

### Bucket semantics

| Bucket | Meaning | Gate behaviour |
|---|---|---|
| **Passed** | ≥1 matching test, all green | ✅ verified |
| **Failed** | ≥1 matching test failed | ❌ **block merge** |
| **Missing** | zero `ls_<>_<n>_*` tests | ⚠️ advisory only |

Missing is advisory (warning, not block) so older approved entries
with ad-hoc test names can be migrated incrementally rather than
blocking every PR.

### Gate state after this PR

19 approved LS entries, **15 passed / 0 failed / 4 missing**.

The 5 newly-passing entries got thin convention aliases (last
commit) so their pre-existing regression tests are discoverable:

| LS  | Original test | Convention alias |
|-----|---------------|------------------|
| LS-P-4  | `test_canonical_abi_size_fixed_size_list_saturates_on_overflow` | `ls_p_4_canonical_abi_size_saturates_on_overflow` |
| LS-P-5  | `test_parser_rejects_truncated_module_section_issue_118`        | `ls_p_5_parser_rejects_truncated_module_section` |
| LS-R-10 | `test_issue112_item5_intra_adapter_preserves_from_import_module`| `ls_r_10_intra_adapter_preserves_from_import_module` |
| LS-CP-3 | `test_issue112_item4_sort_adapter_sites_is_canonical`           | `ls_cp_3_sort_adapter_sites_is_canonical` (adapter-sites half only) |
| LS-A-10 | `cabi_alignment_stackful_retptr_writes_i64_at_offset_8`         | `ls_a_10_cabi_align_retptr_writeback` |

The **4 still-missing** genuinely lack regression tests and will
be addressed in follow-up PRs (one per subsystem):

- **LS-CP-4** — DWARF passthrough emits address-incorrect debug info
[...]

PR #156, opened 2026-05-14
fix(resource-graph,merger): match full (iface, rn) tuple in 4 sites (LS-A-17/18/19)

avrabe:
Fifth fix from the post-v0.8.0 Mythos delta-pass sweep. Four related bugs in `resource_graph.rs` (3) and `merger.rs` (1), all about cleanup / classifier loops dropping the interface dimension of `(component, interface, resource_name)` tuples or using suffix matching where exact match was required.

## The bugs

| LS | Site | Effect |
|---|---|---|
| LS-A-17a | resource_graph.rs definer purge (279-308) | Same-resource-name across different interfaces → wrong definer purged |
| LS-A-17b | resource_graph.rs terminal exporter (242-264) | Unrelated resource import disqualifies the definer |
| LS-A-18 | resource_graph.rs first pass (124-172) | Re-exporter that drops a foreign resource invisible to graph → wrong-table drop → UB |
| LS-A-19 | merger.rs dedup-skip (2127-2153) | `ends_with("float")` matches both `float` and `bigfloat` |

All four lead to silent cross-resource confusion or wrong-handle-table routing with no host trap.

## Fixes

- **definer purge**: filter by `(idx == from_comp && iface == ri && r == rn)` — full tuple match.
- **terminal exporter**: scope `to_also_imports_resource` to the iface under attribution (and `[export]`-prefixed variants).
- **first-pass**: extracted `ResourcePrefixKind` enum; handle `Drop` arm as `ImportsFrom` edge (matches second pass).
- **merger dedup**: exact-match against `format!("[resource-{rep,new}]{rn}")`.

## Tests (3 new)

- `ls_a_17_definer_survives_unrelated_import_with_same_resource_name`
- `ls_a_17_terminal_definer_with_unrelated_resource_import`
- `ls_a_18_drop_on_foreign_resource_registers_node_for_reexporter`

## Tier-5 gate

Touches `resource_graph.rs` and `merger.rs` (both Tier-5).

## Test plan

- [x] `cargo test -p meld-core --lib` — 211 pass (208 prior + 3 new)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] YAML lint
- [ ] CI green on smithy
- [ ] mythos-pass-done label

## Refs

- LS-A-17 (UCA-F-2, H-1, H-3)
- LS-A-18 (UCA-F-2, H-3)
- LS-A-19 (UCA-M-9, H-1, H-10)
[...]

Issue #143, opened 2026-05-12
DWARF Phase 2: address remap (#130)


avrabe:
Sub-issue of **#130** — Phase 2 of the witness-mapping epic.

v0.7 shipped Phase 1.5 (\`DwarfHandling::Strip\` default — stop being wrong). Phase 2 is the real fix: rewrite DWARF addresses across the merged code section so downstream MC/DC tooling (\`pulseengine/witness\`) can produce correct source-line attribution on fused output.

## Scope

Per #130's plan §"Phase 2 — DWARF remap":

- **\`.debug_line\` rewrite.** Single-pass line-program rewrite using the function-body relocation map the merger already builds. Addresses advance sequentially and meld preserves per-body byte offsets modulo index reencoding — straightforward.
- **\`.debug_info\` rewrite.** DIEs that reference code addresses (\`DW_AT_low_pc\`, \`DW_AT_high_pc\`, \`DW_AT_ranges\`) get the same remap. CU concatenation with offset adjustments for multi-input cases.
- **\`.debug_ranges\` / \`.debug_rnglists\`.** Entry-by-entry rewriting.
- **\`.debug_str\` / \`.debug_abbrev\` passthrough.** Address-free; \`.debug_str\` needs string-pool dedup across inputs; \`.debug_abbrev\` merges if byte-equal, else stays per-CU.
- **New \`DwarfHandling::Remap\` variant** on \`FuserConfig\` — opt-in initially, flipped to default once verified against witness on real-world fused modules.
- **End-to-end fixture**: a fused module from \`tests/wit_bindgen/fixtures/\` driven through witness's gimli-based \`(code-offset → file, line)\` build path; assert >X% of \`br_if\` byte offsets get source attribution.

## Dependencies

- Add \`gimli\` (read) + a writer path to \`meld-core\`. Forking the writer-side logic into a thin standalone helper is OK if pulling the full gimli writer pulls too many deps.

## Why not in v0.8 / v0.9

Real DWARF surgery is multi-day work. Witness integration is the verifier — needs the cross-repo loop set up. v0.10 is the soonest realistic milestone.

## Out of scope

- Synthesised DIEs for adapter / inlined code (that's Phase 3, #TBD)
- Variable-level debug info (\`.debug_loc\`, \`.debug_loclists\`) — explicitly out per #130.

## Rivet artifact

- New requirement **SR-35** — "DWARF address remap into merged code section".

## Milestone

v0.10.0

avrabe:
[...]

PR #163, opened 2026-05-17
test(merger): add LS-A-19 regression for exact resource-import lookup

avrabe:
## Summary

Closes the missing-regression gap for **LS-A-19** that the LS-N verification gate (#161) surfaced on its first run. PR #156 fixed the `imp.name.ends_with(rn)` suffix-collision bug in `Merger::add_unresolved_imports` but landed without a dedicated regression test.

Refactors the dedup-skip path's inline `.values().find(...)` lookup into a private helper `find_exact_resource_import_idx` and adds three regression tests pinning the suffix-collision boundaries.

## Why this matters

The bug: when two resources shared a suffix (`float` / `bigfloat`), the dedup-skip path's `imp.name.ends_with("float")` matched both `[resource-rep]float` AND `[resource-rep]bigfloat`. `.find()` returned whichever the HashMap iterated first (also LS-A-15 territory). The wrong import index got stored in `resource_rep_by_component`, routing subsequent `float` calls on the second component to the `bigfloat` import. Validator passes (both share `(i32) -> i32`); runtime sees `bigfloat` rep as `float` — type confusion with no trap (H-10).

## Tests added

| Test | What it pins |
|---|---|
| `ls_a_19_exact_match_picks_float_not_bigfloat` | Both names in tracking; exact match returns the right index regardless of HashMap iteration order |
| `ls_a_19_no_match_returns_none_even_with_suffix_collision` | Only `bigfloat` in tracking, ask for `float` → must return None (buggy `ends_with` would have returned bigfloat's index) |
| `ls_a_19_resource_new_lookup_is_also_exact` | Same case for the `[resource-new]` table |

## Gate impact

LS-N verification gate verdict moves from **15/19 verified (4 missing)** → **16/19 verified (3 missing)**.

Remaining missing-bucket entries:
- LS-CP-4 (DWARF passthrough) — likely subsumed by #130 Phase 2 (#143)
- LS-A-8 (HashMap rep_func selection)
- LS-A-9 (async callback POLL fallthrough)

## End-to-end test for mythos-auto.yml

This is also the **first real PR exercising `mythos-auto.yml`** (added in #162). It touches a Tier-5 file (`merger.rs`), so the auto-runner should fire end-to-end:

1. `Detect Tier-5 changes` job passes the actor gate (avrabe)
2. Matrix runs `claude-code-action` on `merger.rs` with `scripts/mythos/discover.md`
3. Aggregate job posts sticky `<!-- mythos-auto-gate -->` comment
4. If NO_FINDINGS, auto-applies `mythos-pass-done` label

Watching for: budget consumption per single-file scan, sticky-comment markdown rendering, label-application timing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- **LS-A-19 regression coverage** (`meld-core/src/merger.rs`). PR #156
+  fixed the `imp.name.ends_with(rn)` suffix-collision bug
+  (`float` / `bigfloat` cross-resource confusion) but landed without
+  a regression test. The LS-N verification gate (#161) surfaced this
+  gap as the "missing test" advisory bucket. Extracts the
+  exact-match lookup into a private helper
+  `find_exact_resource_import_idx` and adds three regression tests:
+  `ls_a_19_exact_match_picks_float_not_bigfloat`,
+  `ls_a_19_no_match_returns_none_even_with_suffix_collision`, and
+  `ls_a_19_resource_new_lookup_is_also_exact`. Gate result moves
+  from 15/19 verified to 16/19; remaining missing-bucket entries are
+  LS-CP-4 (likely subsumed by #130 Phase 2), LS-A-8, LS-A-9.
+
 ### Added
 
 - **Mythos delta-pass auto-runner** (`.github/workflows/mythos-auto.yml`).
--- a/meld-core/src/merger.rs
+++ b/meld-core/src/merger.rs
@@ -455,6 +455,26 @@ fn effective_module_name(unresolved: &crate::resolver::UnresolvedImport) -> &str
         .unwrap_or(&unresolved.module_name)
 }
 
+/// Resolve the imports-vector index whose name exactly matches
+/// `expected_name` by scanning the values of a per-component
+/// resource-tracking map. Exact match (not `ends_with`) — the prior
+/// `imp.name.ends_with(rn)` form silently conflated two resources
+/// whose names shared a suffix (e.g. `float` matched both
+/// `[resource-rep]float` and `[resource-rep]bigfloat`), letting the
+/// dedup-skip path register the wrong import for the wrong-suffix
+/// component. See LS-A-19 for the regression.
+fn find_exact_resource_import_idx(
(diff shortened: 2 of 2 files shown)


avrabe:
## Admin-merging + plumbing follow-up

Two distinct issues blocked the PR's CI gates; only the first one is a real test failure that would block merge.

### #139 capacity (real)

3 standard CI jobs still queued (Clippy + 2 fuzz smokes) after ~2 hours. After cancelling 5 hung cross-org jobs (loom × 3 + spar × 2, ages 9-22h, matching #139 §4 "hung jobs hold slots forever" pattern), capacity partially recovered but the queue cleared in zigzag.

### mythos-auto.yml plumbing (will fix in #164)

The `Mythos pass (meld-core/src/merger.rs)` job failure is workflow plumbing, NOT a Mythos finding:

```
14:32:12 Error: Unable to locate executable file: unzip. Please verify either the file path exists...
14:32:12 /var/lib/runners/runner5/_work/_temp/.../d9f8a6f8.sh: line 3: bun: command not found
14:32:12 ##[error]Process completed with exit code 127.
```

`oven-sh/setup-bun` failed because the rust-cpu runner lacks `unzip`. The action's fallback to a system `bun` also failed because there is no system bun. claude-code-action's bun-based post-step entrypoints then exit 127.

The placeholder-FINDING fallback we wired in for exactly this scenario fired correctly, but the slug step output was empty (separate plumbing bug), producing `mythos-out/.json` as the artifact path.

### Why admin-merge is OK here

- The LS-A-19 regression is verified locally (`cargo test ls_a_19_` → 3 pass; full suite 236/236)
- The LS-N gate (`verification-gate.yml`) passes — that's the test-coverage gate this PR was written to satisfy
- Format, Test, Coverage, Bench, fuzz_merger_idempotent, fuzz_resolver_terminates, LS-N gate: **all 7 ran and passed**
- The 3 remaining stuck jobs are independent of this PR's contents
- Same #139 admin-merge precedent as #161 / #162

Admin-merge counter for #139:
- PR #163 — admin-merged (this PR)

Follow-up PR #164 (next) will fix mythos-auto plumbing:
1. Pre-install `unzip` (and verify `bun` once `setup-bun` works)
2. Debug the empty-slug step (likely matrix-expression evaluation order)
3. Confirm OAuth-token wiring is the right input for non-interactive use

Issue #139, opened 2026-05-12
Smithy fleet reliability — rolling tracker (meld side)


avrabe:
Rolling tracker for meld's experience of the smithy CI fleet. Companion issue to whatever lands in \`pulseengine/smithy\`; this issue captures meld-specific observations, failed PRs, and the acceptance bar for releases.

ubuntu-latest is **not** a fallback option per project policy. Smithy stability is a release-path dependency.

### 1. Whole-fleet offline events

- **2026-05-03 ~19:00Z**: all 8 \`pulseengine-ci-01-*\` runners reported \`status: offline\` simultaneously. Recovered ~3 h later (status: online, busy: true).
- **2026-05-10 ~17:00Z–17:30Z**: full offline window again, recovered within ~20 min.

Pattern: when the host(s) go down, every queued job sits forever (jobs queue against runner labels, not against fleet health). meld saw a 2 h 11 min queue on PR #135 with zero pickup before the fleet came back.

### 2. Disk-space failures on rust-cpu runners

Runner pool currently 3× rust-cpu (\`pulseengine-ci-01-{5,6,7}\`). Two of three show the same pattern:

| Runner | Symptom |
|---|---|
| \`-5\` | Jobs fail in 30–70 s with \`error: failed to build archive: No space left on device (os error 28)\` |
| \`-6\` | Same as -5; also went offline mid-day twice during the v0.7.0 cycle |
| \`-7\` | Fuzz + bench succeed here |

Concrete failures: PR #134 (Bench compile), PR #135 (Bench, Coverage, fuzz_resolver_terminates), PR #137 (Bench, Coverage, Test, fuzz_parse_component, fuzz_fusion_roundtrip). Each landed via \`gh run rerun --failed\` until \`-7\` was assigned, except PR #137 which had to merge via \`--admin\` because the rerun cycle never converged.

### 3. Per-runner config drift (sanitizer + musl)

PR #135 fuzz failure on runner-5:

\`\`\`
error: sanitizer is incompatible with statically linked libc,
disable it using \`-C target-feature=-crt-static\`
\`\`\`

Same workflow, same toolchain install step (\`dtolnay/rust-toolchain@nightly\` with \`targets: x86_64-unknown-linux-musl\`), succeeds on \`-7\`, fails on \`-5/-6\`. Suggests host-level \`.cargo/config.toml\` or rustup component state differs between runners.

### 4. Cross-org capacity contention

While meld jobs queued, other org repos held rust-cpu slots:

[...]

PR #166, opened 2026-05-19
test(fact): LS-A-9 regression for callback POLL dispatch

avrabe:
## Summary

Closes the missing-regression gap for **LS-A-9** that the LS-N verification gate has been surfacing. Prior fix at `meld-core/src/adapter/fact.rs:3992-3998` added the `WAIT || POLL` OR-pattern to the callback dispatch but landed without a dedicated regression test.

## Why this matters

The async-callback adapter dispatches `[waitable-set-poll]` when the guest callback returns `WAIT (2)` or `POLL (3)`. A previous version matched only `WAIT`, silently letting `POLL` fall through to the YIELD path — which sent `(EVENT_NONE, 0, 0)` to `[callback]` and dropped host-ready events. The fused module diverged from the composed module on P3 async export call paths (H-3, UCA-F-3), and there's no host trap to catch it.

## Test approach

`ls_a_9_callback_adapter_dispatches_both_wait_and_poll`:
1. Builds a minimal `MergedModule` with the prerequisites the callback emitter requires (lift func, `[callback]<export>` companion, `[waitable-set-poll]` import).
2. Drives `generate_async_callback_adapter` end-to-end.
3. Scans the emitted function body for the canonical byte sequence:
   ```
   0x41 0x02 0x46 0x20    i32.const WAIT(2) / i32.eq / local.get
   0x41 0x03 0x46 0x72    i32.const POLL(3) / i32.eq / i32.or
   ```
   in order. A regression that drops the POLL arm or reorders the comparisons fails this test.

Pin-by-substring is robust against unrelated body changes (locals layout, surrounding control flow).

## Gate verdict

| State | Before | After |
|---|---|---|
| Verified | 17/19 | **18/19** |
| Missing | 2 (LS-A-8, LS-A-9) | 1 (LS-A-8) |
| Failed | 0 | 0 |

Remaining: LS-A-8 (HashMap rep_func selection for `list<record<borrow>>`) — surveyed in task #52 at ~2-4h scope.

## End-to-end mythos-auto smoke test

This is the **first PR to touch a Tier-5 source file** (`adapter/fact.rs`) since the auto-runner's plumbing fixes (#164). Watching for:

- Detect job lists `meld-core/src/adapter/fact.rs` ✓
- Scan matrix actually fires (the path that was broken in #163's run)
- claude-code-action emits structured `verdict: NO_FINDINGS | FINDING` JSON
- Sticky comment posted with per-file verdict
- `mythos-pass-done` label auto-applied on NO_FINDINGS

Local tests: `cargo test ls_a_9_` → 1 pass; full `meld-core --lib` → 237/237.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- **LS-A-9 regression coverage** (`meld-core/src/adapter/fact.rs`).
+  PR fixed the callback-mode `if code == WAIT` branch that silently
+  treated `POLL (3)` as a YIELD fall-through (dropping host-ready
+  events on the callback handshake), but landed without a dedicated
+  test. The LS-N verification gate surfaced this gap. Adds
+  `ls_a_9_callback_adapter_dispatches_both_wait_and_poll`, which
+  drives `generate_async_callback_adapter` end-to-end against a
+  minimal merged-module fixture (lift func + `[callback]` companion
+  + `[waitable-set-poll]` import) and asserts the emitted body
+  contains the canonical `i32.const 2 / i32.eq / local.get … /
+  i32.const 3 / i32.eq / i32.or` byte sequence in order. Gate
+  verdict moves from 17/19 verified to **18/19** — only LS-A-8 left
+  in the missing bucket. **First PR to touch a Tier-5 file** since
+  `mythos-auto.yml`'s plumbing fixes (#164), so this also serves as
+  the auto-runner's first true end-to-end matrix-scan smoke test.
+
 - **LS-CP-4 regression coverage + LS-N gate integration-test scan**
   (`tools/run_ls_verification.py`,
   `meld-core/tests/dwarf_strip.rs`). The LS-N verification gate was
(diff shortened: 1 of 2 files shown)


avrabe:
## Admin-merge per #139 (filed 2 smithy issues)

9 of 13 checks green including everything substantive (Test, Clippy, Coverage, Bench, LS-N gate, 2 fuzz smokes, Detect Tier-5 changes). The 4 remaining failures are all #139 infra, not regressions in this PR's content:

- **`Fuzz smoke (fuzz_parse_component)` + `Fuzz smoke (fuzz_resolver_terminates)`** — same "sanitizer is incompatible with statically linked libc" config drift documented in #139 §3. Reruns queued ~25min on the 7/7-busy fleet without picking up.
- **`Mythos pass (meld-core/src/adapter/fact.rs)`** — `unzip` missing on rust-cpu runners. `oven-sh/setup-bun` can't extract the bun zip, claude-code-action's bun-based entrypoints exit 127, placeholder-FINDING fallback fires.
- **`Aggregate findings + label` + `Mythos delta-pass gate`** — cascading from the auto-runner failure (no NO_FINDINGS verdict → no label → label-only gate fails).

Two new tracking issues filed for smithy:
- #167 — `unzip` missing on rust-cpu runners blocks claude-code-action
- #168 — sanitizer/libc config drift recurrence on fuzz smoke jobs

LS-A-9 regression test itself is sound (1 pass, 237/237 in `meld-core` lib). The auto-runner was supposed to validate end-to-end here; with the unzip block, that validation is deferred until #167 ships.

Admin-merge counter for #139 since last reset:
- PR #166 — admin-merged (this PR, +1)

Issue #167, opened 2026-05-19
smithy: unzip missing on rust-cpu runners blocks claude-code-action


avrabe:
## Summary

Self-hosted rust-cpu runners (`pulseengine-ci-01-{5,6,7,9,10,11,12}`) lack `unzip`, which `oven-sh/setup-bun` requires to extract the bun binary. This blocks `anthropics/claude-code-action` from running, which in turn blocks meld's `mythos-auto.yml` (#162) from completing its scan matrix.

## Observed

PR #163 (first end-to-end run): scan job ran with the runner attempting `setup-bun` → fail at `Unable to locate executable file: unzip` → action's bun-based entrypoints exit 127 → placeholder-FINDING fallback fires → workflow fails.

PR #166 (second attempt, after `mythos-auto.yml` plumbing fix in #164): same `unzip` error, same cascade. My in-workflow `apt-get install -y unzip` attempt also fails (probably `apt-get update` issue or sudo perms — exit code 1).

Concrete log snippet:
```
##[error]Error: Unable to locate executable file: unzip. Please verify either the file path exists ...
/var/lib/runners/runner5/_work/_temp/.../d9f8a6f8.sh: line 3: bun: command not found
##[error]Process completed with exit code 127.
```

## Impact

The `mythos-auto.yml` workflow (the AI-driven Mythos discover-protocol auto-runner shipped in #162) cannot complete its matrix scan until `unzip` is available on rust-cpu runners. Every Tier-5 PR by `avrabe` will trigger this failure cascade.

## Asks

1. **Install `unzip` in the rust-cpu runner image** (or via post-provisioning script). It's a 200KB Debian package; cost is trivial.
2. **Alternative**: provide a pre-installed bun binary at a known path so workflows can pass `path_to_bun_executable` to `claude-code-action` and skip `setup-bun` entirely.

## Workaround for now

`mythos-auto.yml` falls back to a FINDING placeholder when structured output is empty; PR-time `mythos-pass-done` label must be applied manually until this is resolved. The label-only `mythos-gate.yml` (#151) is unaffected — that gate just checks for the label.

## References

- #139 §3 (per-runner config drift) — same family of bug, different package
- PR #162 (`mythos-auto.yml` introduction)
- PR #164 (`mythos-auto.yml` plumbing fixes; tried in-workflow `apt-get install unzip` but the install step itself exits 1)
- PR #163 first end-to-end attempt logs: https://github.com/pulseengine/meld/actions/runs/25984380642
[...]

PR #169, opened 2026-05-19
test(fact): LS-A-8 regression for inner-list rep_func selection

avrabe:
## Summary

The **final** missing-bucket entry. Drives the LS-N verification gate to **19/19 verified — full coverage**.

PR f0e981b fixed the `resource_rep_imports.values().next()` HashMap-iteration-order bug in `analyze_call_site` (inner-list borrow rep_func selection) by threading a pre-resolved `rep_import: Option<(String, String)>` through `InnerResource` and looking up per-type. The fix landed without a regression test; the LS-N verification gate has been flagging this since #161.

## Why this matters

For `list<record { x: borrow<A>, y: borrow<B> }>`, the pre-fix code picked the same arbitrary rep_func for both A and B handles. Borrow handles of resource A then routed through resource B's rep_func (silent rep-vs-handle confusion across cross-component handle boundary, H-4 / H-4.2, UCA-A-7). Wasm validator accepts (both `[resource-rep]X` and `[resource-rep]Y` are `(i32) -> i32`); divergence only surfaces at downstream runtime execution.

## Test approach

| Test | What it pins |
|---|---|
| `ls_a_8_inner_list_rep_func_selected_by_type_not_iteration_order` | Adversarial 2-resource fixture, sampled **32×** across rebuilt HashMaps (fresh hash seed per sample). If any residual iteration-order dependence leaked, at least one sample would pair wrongly; assertion pins the per-type lookup as deterministic. |
| `ls_a_8_no_rep_import_skips_fixup_rather_than_picking_arbitrary` | Pre-fix `.values().next()` fall-back-on-miss would fabricate a fixup pointing at an arbitrary rep_func when the resolver couldn't map the type id. The post-fix path correctly skips with `log::warn`. This test pins that skip-not-fabricate behaviour. |

## Gate verdict

| State | Before | After |
|---|---|---|
| Verified | 18/19 | **19/19** |
| Missing | 1 (LS-A-8) | **0** |
| Failed | 0 | 0 |

Advisory missing-bucket is now empty. Every `status: approved` LS entry has a passing regression test.

## Local test run

```
running 2 tests
..
test result: ok. 2 passed; 0 failed; 0 ignored; ...
```

Full lib: 237/237 → **239/239**.

## End-to-end mythos-auto status

This PR touches `meld-core/src/adapter/fact.rs` (Tier-5) so the auto-runner fires. Same `unzip`-missing infra block as PR #166 expected (#167). LS-A-9 test itself was admin-merged via #166; same path likely needed here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- **LS-A-8 regression coverage** (`meld-core/src/adapter/fact.rs`).
+  PR f0e981b fixed the `resource_rep_imports.values().next()`
+  HashMap-iteration-order bug in `analyze_call_site` (inner-list
+  borrow rep_func selection) by threading a pre-resolved
+  `rep_import: Option<(String, String)>` through `InnerResource` and
+  looking up per-type. The fix landed without a dedicated
+  regression test; the LS-N verification gate surfaced this as the
+  **last** missing-bucket entry. Adds two tests:
+  `ls_a_8_inner_list_rep_func_selected_by_type_not_iteration_order`
+  (adversarial 2-resource list, asserts per-byte-offset fixups map
+  to the correct rep_func — sampled 32× to make any residual
+  HashMap-order leak observable) and
+  `ls_a_8_no_rep_import_skips_fixup_rather_than_picking_arbitrary`
+  (pins the `rep_import=None` skip-with-warning path against the
+  pre-fix arbitrary-fallback regression). Gate verdict moves from
+  18/19 verified to **19/19 — full coverage**. The advisory
+  missing-bucket is now empty.
+
 - **LS-A-9 regression coverage** (`meld-core/src/adapter/fact.rs`).
   PR fixed the callback-mode `if code == WAIT` branch that silently
   treated `POLL (3)` as a YIELD fall-through (dropping host-ready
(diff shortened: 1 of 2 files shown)


avrabe:
## Admin-merge per #139 + #170 follow-up

10 of 14 checks green:
- Test ✓, Clippy ✓, Coverage ✓, Bench ✓, LS-N gate ✓, Format ✓
- Detect Tier-5 ✓ (actor gate works), all 4 fuzz smokes green (3/4 done, 1 still queued but will pass)

The 3 failing checks are all the mythos-auto trio:
- `Mythos pass (meld-core/src/adapter/fact.rs)` — claude-code-action aborts with `Could not fetch an OIDC token. Did you remember to add 'id-token: write' to your workflow permissions?`
- `Aggregate findings + label` — cascade
- `Mythos delta-pass gate` — cascade (label not applied)

The OIDC permission fix lives in **PR #170**, which adds `id-token: write` to `mythos-auto.yml`. The 3 mythos failures CANNOT pass on this branch because the fix isn't here. Once #170 merges, every subsequent Tier-5 PR will run mythos-auto cleanly end-to-end.

The LS-A-8 regression test itself is sound (239/239 in `meld-core --lib`, both new tests green locally + on this PR's Test job).

Admin-merge counter for #139:
- PR #169 — admin-merged (this PR, +1)

**Drives the LS-N verification gate to 19/19 verified — full coverage. Advisory missing-bucket is now empty.**

Issue #174, opened 2026-05-20
v0.5 post-ship sweep (2026-05-20)


avrabe:
Automated post-ship verification for v0.5.0, run 2026-05-20 per the release-note callout ("The 2026-05-20 verification routine includes a deeper Mythos sweep that will revisit"). Steps follow the v0.5 post-ship checklist.

---

## Step 1 — v0.5.0 shipped ✅

Tag `v0.5.0` exists and the GitHub release is published (2026-04-29T20:43:49Z). Four binaries uploaded (aarch64-apple-darwin, aarch64-unknown-linux-gnu, x86_64-apple-darwin, x86_64-unknown-linux-gnu). Release closed #103, #118, and partial #94 as advertised.

**Note:** the repo has since advanced three minor releases + one patch (current latest: v0.8.1). All downstream steps are assessed against the current `main`.

---

## Step 2 — Post-release CI on `main` ✅

Latest `main` commit: `17b76df` (PR #171, merged 2026-05-20T04:14:48Z). All 14 check runs green:

| Job | Result |
|---|---|
| Format | ✅ success |
| Clippy | ✅ success |
| Test | ✅ success |
| Coverage | ✅ success |
| Bench compile-only | ✅ success |
| Fuzz smoke × 4 | ✅ success (all four targets) |
| Mythos delta-pass gate | ✅ success (no Tier-5 changes in PR #171) |
| LS-N verification gate | ✅ success |
| Mythos auto-runner | ⏭ skipped (CI-only change; no Tier-5 paths touched) |

No regressions. No action needed.

---

## Step 3 — Stale branches ⚠️ (owner decision needed)

Active branches (skip):
- `feat/141-stream-pair-foundation` → open PR #173
- `bot/config-update-1777161165595` → open PR #110 (bot namespace, do not touch)

Candidates for deletion (no open PR; not in an active worktree):

| Branch | Status |
|---|---|
| `fix/component-wrap-option-drops` | PRs #155 and #158 both **closed without merge** from this branch; fix landed via separate PR #159 from `fix/wrap-lift-string-encoding`. Branch is abandoned. |
| `ci/update-wit-bindgen-fixtures` | No open or merged PR found. Appears to be in-progress work or a forgotten WIP. |
| `feat/explore-I-opaque-rep` | No PR found. Likely exploratory WIP. |
| `feat/issue-92-shim-module` | No PR found. Likely WIP. |
| `feat/per-component-handle-tables` | No PR found. Likely WIP. |

[...]

PR #178, opened 2026-05-21
test(parser): #174 Mythos sweep — section-range invariant guard (NO FINDINGS)

avrabe:
## Summary

Resolves **Step 5** of issue #174 — the v0.5 post-ship Mythos sweep's outstanding unverified hypothesis.

**Verdict: NO FINDINGS.**

## The hypothesis

`parse_core_module` stores `reader.range()` for the element and data sections into `element_section_range` / `data_section_range` (parser.rs:1279 / :1287). `parse_element_segments` / `parse_data_segments` then slice `module.bytes[start..end]` from those ranges with no explicit bounds check (segments.rs:198 / :258). #174 asked: are these LS-P-5 siblings — could `wasmparser` yield a core-module section reader with a range past the buffer, the way `ModuleSection::unchecked_range` could?

## Why it's NO FINDINGS

It cannot. The crux is the difference between `Payload::ModuleSection` and a core-module section:

- `ModuleSection` is yielded **eagerly** with an explicitly *unchecked* range — the nested module isn't parsed yet. That's what made LS-P-5 exploitable.
- A core-module element/data section is only framed once `parse_all` has its full declared content. A truncated section — size LEB claiming more bytes than remain — makes `parse_all` yield an **`Err`**, which `parse_core_module`'s `payload?` propagates. The `*_section_range` field is never set.

So the downstream slice is **defended by construction**: every range that reaches it came from a section `wasmparser` successfully framed, and a framed section's range is in-bounds.

## The oracle

Per the Mythos protocol, a NO FINDINGS verdict still wants an oracle. `truncated_core_section_errors_rather_than_yielding_oob_range` feeds truncated element- and data-section inputs (size LEB = 16, only 2 content bytes) and asserts `wasmparser` rejects each with an `Err` rather than handing back a section reader with an out-of-bounds range.

It's also a **standing regression guard**: a future `wasmparser` bump that changed the framing behaviour would fail this test and reopen the hypothesis — at which point 1279/1287 would need a `checked_section_slice`-style guard before the segments.rs slice.

## Scope

- No production code change — the slice sites are correct as-is given the invariant.
- No LS-N entry (NO FINDINGS — LS-N entries are for confirmed findings).
- Touches `parser.rs` (Tier-5) — the Mythos auto-runner will scan it.

Refs: #174 Step 5, LS-P-5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,24 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- **Regression guard for the `parse_core_module` section-range
+  invariant** (issue #174 Step 5, `meld-core/src/parser.rs`). The
+  v0.5 post-ship Mythos sweep flagged an unverified LS-P-5-sibling
+  hypothesis: `parse_core_module` stores `reader.range()` for the
+  element/data sections and `parse_element_segments` /
+  `parse_data_segments` slice `module.bytes[start..end]` from it
+  without an explicit bounds check. Mythos delta-pass verdict:
+  **NO FINDINGS**. Unlike `ModuleSection::unchecked_range` (the
+  LS-P-5 site), a core-module element/data section is only framed
+  once `wasmparser` has its declared bytes — a truncated section
+  makes `parse_all` yield an `Err`, propagated by
+  `parse_core_module`'s `payload?`, so no out-of-bounds range is
+  ever stored. Adds `truncated_core_section_errors_rather_than_yielding_oob_range`
+  to pin that `wasmparser` framing guarantee: if a future bump
+  changes it, the test fails and the hypothesis reopens.
+
 ### Changed
 
 - **`meld fuse` warns when the `multi` memory default is used**
--- a/meld-core/src/parser.rs
+++ b/meld-core/src/parser.rs
@@ -3388,6 +3388,90 @@ mod tests {
         assert!(matches!(result, Err(Error::NotAComponent)));
     }
 
+    /// #174 Step 5 — Mythos delta-pass regression guard (LS-P-5
+    /// sibling hypothesis: **NO FINDINGS**).
+    ///
+    /// `parse_core_module` stores `reader.range()` for the element and
+    /// data sections into `element_section_range` / `data_section_range`
+    /// (parser.rs:1279 / :1287). `parse_element_segments` /
+    /// `parse_data_segments` (segments.rs:198 / :258) then slice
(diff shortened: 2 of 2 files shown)


avrabe:
## Mythos auto-runner finding — reviewed, dispositioned (not confirmed)

The mythos-auto scan of `parser.rs` returned a FINDING. Validated per `scripts/mythos/validate.md`:

**The finding:** `flat_byte_size` computes the payload width of `result<T,E>` / `variant` as `max(flat_byte_size(arm))` rather than the Component Model's element-wise `flatten_variant` JOIN. For `result<u64, string>` it returns 12; the JOIN-correct value is 16 (joined flat sequence `[i32, i64, i32]`).

**Arithmetic — correct.** `max()` of the arms' byte totals underestimates whenever the arms flatten to different *numbers* of core values: `u64`→`[i64]` (8 B, 1 value), `string`→`[i32,i32]` (8 B, 2 values), `max(8,8)=8`, but the element-wise join `[join(i64,i32), i32]` = `[i64,i32]` = 12 B, +4 discriminant = 16. Real off-by-4.

**Impact claim — rejected.** The finding asserts "adapter code uses this to size a retptr return-area buffer → 4-byte underallocation → OOB write." That is a hallucination: `flat_byte_size` has **zero consumers** in `meld-core/src/` — only its own recursion and one saturation test (`parser.rs:4395`). Retptr return areas are sized by `return_area_byte_size` → `canonical_abi_size_unpadded`, a different function entirely. There is no execution path where the 12-vs-16 discrepancy reaches a buffer allocation.

**Verdict: NOT a confirmed finding.** Per `discover.md`'s oracle rule, a confirmed finding needs a failing PoC — and no PoC can reach dead code. No LS-N entry. The discover step over-reached on impact; this is exactly the discover→validate split working.

**Residual:** the arithmetic *is* a latent correctness defect in a `pub fn`. Tracked for a separate hygiene PR — fix `flat_byte_size`'s `result`/`variant`/`option` arms to the correct element-wise JOIN, regression test, no LS-N (no reachable hazard).

## This PR

PR #178's own change is the #174-Step-5 NO-FINDINGS regression guard (`truncated_core_section_errors_rather_than_yielding_oob_range`) — test-only, unrelated to the `flat_byte_size` finding. 12 substantive checks green. The auto-runner flagged a *pre-existing* line of `parser.rs` because it scans the whole file.

Applying `mythos-pass-done` — this is the gate's designed human-review flow (a maintainer reviewed the finding and recorded the disposition), not a bypass.

Issue #212, opened 2026-05-31
Multi-component fusion gaps surfaced by golden-e2e (exports dropped / cross-link not internalised / bare-world result type)


avrabe:
Building the golden-e2e behavioural-equivalence harness (Tier B) surfaced three real gaps in multi-component fusion. **Tier A is unaffected** — real wit-bindgen "host+guest" compositions fuse and run with byte-identical behaviour (12 checks, incl. `hello_rust/c/cpp` stdout). These gaps are on minimal hand-built compositions and the separate-inputs path.

Repro fixtures + build under `meld-core/../tests/wit_bindgen/fixtures/compose/` (`build.sh`, offline via wasm-tools + wac): a `consumer` whose `runner.compute()` calls a `provider`'s `add(20,22)=42`.

### 1. Separate inputs: cross-component interface link not internalised
```
meld fuse consumer.wasm provider.wasm -o f.wasm --component
```
Fused component **still imports** `golden:math/lib@0.1.0` (consumer's import is not wired to provider's matching export), so `compute()` can't run standalone:
```
component imports instance `golden:math/lib@0.1.0`, but a matching implementation was not found in the linker
```
Expected: meld resolves the import against the other input's export and internalises the call → self-contained component, `compute() == 42`.

### 2. `wac`-composed input: top-level export dropped
```
wac plug consumer.wasm --plug provider.wasm -o composed.wasm   # exports golden:app/runner, runs compute()==42
meld fuse composed.wasm -o f.wasm --component
wasm-tools component wit f.wasm   # => `world root {}` — ALL exports dropped
```
The fused component validates but exposes nothing — functionally empty.

### 3. Bare world-level func export: result type dropped → invalid component
With `export compute: func() -> u32;` (bare world export, not via an interface), fusing the composed component produces an **invalid** component:
```
error: lowered result types `[]` do not match result types `[I32]` of core function 0
```
Moving the export into an interface (`export runner;`) avoids it — so the bug is specific to bare world-level function exports' result types in `component_wrap`.

### Where it bites
These block a true end-to-end "fuse N real components, run the composition" path. (1) is the most impactful — it's meld's headline capability. Likely all in the component-wrap / cross-component import-resolution layer.

### Harness
[...]

PR #214, opened 2026-06-04
fix(component-wrap): preserve multi-component composition exports + types (#212.2)

avrabe:
Fixes **#212.2** — fusing a `wac`-composed multi-component (meld's documented input) dropped or mis-typed the composition's top-level export. Two real bugs, both in `wrap_as_component`, because a composition spreads the **export shell**, the **interface-instance export**, and the **canon-lift** across *different* parsed components — meld's wrap assumed one component held all three (true for single wit-bindgen fixtures, false here).

## The two layers

1. **Export dropped → empty `world root {}`.** Source selection ranked candidates only by `depth_0_sections` count, which ties at 0 for a wac-composed input, so `max_by_key` returned the *last* candidate (a bare-func sub-component) and the real top-level export was lost. Now ranked by `(depth0, top-level instance-export count, prefer-original)`.

2. **Wrong signature → silent ABI mismatch.** With the export surfaced, the lift type was resolved only against that (wrong) component, missing the real lift → fell back to `func() -> result` instead of the true signature. `find_lift_type_for_interface_func` now searches all inputs and resolves a lift via the owning component's **CORE export name → `core_func_index` → matching `CanonicalEntry::Lift`** (the interface-instance shape), returning the *owning* component so its component-relative type index resolves correctly.

## Verification
- **LS-CP-5** (approved). Unit test `ls_cp_5_lift_resolved_cross_component_via_core_export` pins the cross-component resolution (gate-checked).
- **golden-e2e Tier B** `tier_b_fused_composed_matches_host_linked` (now **un-ignored**): fuses a real consumer→provider composition and asserts the fused `compute()` returns the host-linked `add(20,22) == 42` — correct export *and* correct signature — under both memory strategies.
- **Tier A** (single wit-bindgen fixtures, 12 checks) unchanged — **no regression** to the path that assumes one source component.
- Full meld-core suite green; clippy/fmt clean.

## Scope
The *separate*-input linking case (hand meld two unlinked components and have it resolve the cross-component interface link itself) remains open as **#212.1** (`tier_b_separate_inputs_internalise_link`, still `#[ignore]`); the supported workflow there is `wac compose` then `meld fuse`. `component_wrap.rs` is Tier-5 → gets the Mythos delta scan.

### DIFF
--- a/meld-core/src/component_wrap.rs
+++ b/meld-core/src/component_wrap.rs
@@ -52,13 +52,35 @@ pub fn wrap_as_component(
     memory_strategy: MemoryStrategy,
     opaque_resources: &[(String, String)],
 ) -> Result<Vec<u8>> {
-    // Pick the component with the most depth_0_sections (widest interface).
-    // Prefer original (un-flattened) components since flattening may drop
-    // the outer shell that carries the depth_0_sections.
+    // Pick the source component that carries the widest *top-level*
+    // interface. Ranking by `depth_0_sections` alone is ambiguous when
+    // every candidate ties at 0 — e.g. a `wac`-composed input whose
+    // flattened sub-components have no captured depth-0 sections. There
+    // `max_by_key` returns the *last* candidate, often a flattened
+    // sub-component exporting a single bare func, so the real top-level
+    // instance export is lost and the wrapped component comes out with an
+    // empty world (#212). Rank by (depth-0 sections, top-level
+    // instance-export count, prefer-original): instance-export count
+    // surfaces the component holding the composition's real exports, and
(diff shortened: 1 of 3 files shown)


avrabe:
Closing: the Mythos clean-room scan correctly found Strategy B conflates module-local `ModuleExport.index` with component-level `Lift.core_func_index` (different index spaces). The minimal fixture passed only because the consumer has a single lift (fallback returned the right type by luck); the fix is wrong for the general case (core-imported funcs / multiple lifts). Not mergeable. Robust-fix path (resolve `core_func_index` → name via `core_entity_order`, + a divergence-exercising fixture) characterized in #212. Branch `fix/212-instance-export-lift` kept as the basis.

PR #216, opened 2026-06-05
fix(component-wrap): name-based lift resolution for composed exports (#212.2)

avrabe:
## What

Robust fix for **#212.2** — when meld fuses a `wac`-composed
multi-component, `wrap_as_component` was **dropping the composition's
top-level export** and (once surfaced) **mis-typing it** as `() -> result`.

Supersedes the closed #214, whose index-based lift match a clean-room
scan flagged as conflating two distinct index spaces.

## Root cause

For a `wac`-composed input the pieces are spread across *different*
parsed components: the top-level instance export lives in the outer
shell, while the canon-lift carrying the real WIT type lives in a
flattened inner component, reachable only through an interface-*instance*
export. Two compounding failures:

1. **Drop** — source selection ranked by `depth_0_sections` count, which
   ties at 0 for these inputs; `max_by_key` picked the last candidate (a
   bare-func sub-component) → empty `world root {}`.
2. **Mis-type** — the lift was looked up only on that wrong component, so
   the export emitted `() -> result` instead of its true signature.

## Fix (`component_wrap.rs`)

- **Source selection** now ranks by `(depth_0_sections, top-level
  instance-export count, prefer-original)` — the component owning the
  composition's real exports wins even when depth-0 ties.
- **`find_lift_type_for_interface_func`** searches all input components
  and resolves the lift **by name** (owning component's core export name
  → `core_func_index` → matching `CanonicalEntry::Lift`), returning the
  *owning* component so its component-relative type index resolves.
  New `core_func_name` helper walks `core_entity_order` to map a
  component-level core-func index back to its alias name.

Name-based (not index-based) is deliberate: matching `ModuleExport.index`
against `Lift.core_func_index` conflates module-local and flattened
component core-func index spaces — correct only when they coincide.

## Verification

- **`ls_cp_5_lift_resolved_cross_component_via_core_export`** — adversarial
  unit test (lift at `core_func_index=1`, unnamed builtin at 0) pins the
  name resolution.
- **golden-e2e Tier B** fixture rebuilt to be **adversarial**: inner core
  module *imports* `add` (so module-local and component core-func index
  spaces diverge) and exports two *differently-typed* funcs
  (`compute()->u32=42`, `greet()->u8=7`) checked by typed wasmtime calls.
  An index-matching fix swaps/mis-types them and fails here.
  `tier_b_fused_composed_matches_host_linked` asserts `(42, 7)` under both
  memory strategies. Tier A (single-component wit-bindgen) unchanged.
- Full `meld-core` suite green; LS gate green (LS-CP-5 mapped).

## Safety traceability

`LS-CP-5` updated to record the name-based resolution and the divergent-
type acceptance fixture.


avrabe:
## Mythos Tier-5 delta-pass — `meld-core/src/component_wrap.rs`

Per `scripts/mythos/HOWTO-gate.md`, a clean-room discover pass
(`scripts/mythos/discover.md`) was run against the Tier-5 diff in this PR
by a fresh agent session with **no prior context** about the change's intent.

### Result: **NO FINDINGS**

The reviewer treated the diff adversarially and chased every hypothesis in
`discover.md` against the real source, using the parser (`parser.rs`) and the
resolver (`resolver.rs`) as ground truth for the component-level core-function
index space.

**Central off-by-one / ABI-swap hypothesis — refuted.** The new
`core_func_name` helper (`component_wrap.rs`) counts the core-func index space
as `CanonicalFunction(_)` → always +1, `CoreAlias(idx)` → +1 only when
`component_aliases[idx]` is `CoreInstanceExport{ kind: Func }`. This is
**byte-for-byte the same counting** the authoritative resolver performs in its
two independent `core_entity_order` walks (`build_entity_provenance` and
`build_resource_type_to_import`), and both consume the *same*
`core_entity_order` vector the parser populated and against which
`CanonicalEntry::Lift::core_func_index` was assigned. An interleaved non-Func
`CoreAlias` (Memory/Table/Global) before the target — the exact divergence the
reviewer tried to construct — is skipped identically by both, so no silent ABI
swap occurs.

**Other hypotheses sound:** all index accesses use `.get()` (no panic on
attacker-shaped `alias_idx`); the `u32` func counter cannot overflow within a
memory-bounded parsed vector; the source-ranking change is self-healing because
`find_lift_type_for_interface_func` now searches `source` *plus all* other
components, so a ranking that picks a source lacking the lift still resolves it
elsewhere; the first-lift fallback was already a non-guaranteeing fuzz-safety
path. The new adversarial unit test
`ls_cp_5_lift_resolved_cross_component_via_core_export` exercises exactly the
interleaved-unnamed-canon-before-the-lift shape and passes; all 17
`component_wrap` unit tests are green.

No concrete input was found that makes the new code emit a wrong lift
type/options where the old code was correct. Per `discover.md`'s
both-oracles-or-silence rule, **no finding is reported** rather than a
speculative one.

> NB the automated `mythos-auto.yml` jobs (`Mythos pass …`, `Aggregate
> findings + label`) failed on their own infra (empty slug output / installation-
> token step), unrelated to this change. The authoritative gate is the
> human-verified `mythos-pass-done` label per `HOWTO-gate.md`.

PR #220, opened 2026-06-09
feat(memory): MemoryStrategy::Auto default — sound single-memory fusion (#172)

### LINKED ISSUE #172: meld fuse defaults to --memory multi; fused module is rejected by wasm-opt and unusable for synth without extra flags
## Summary

`meld fuse` defaults to `--memory multi`. The fused core module then carries one linear memory per input component. Every standard downstream tool then rejects it unless given extra flags — and for an MCU AOT target (single flat address space) the multi-memory form has no valid lowering at all.

## Repro

```
# two trivial components, fuse with defaults
meld fuse a.component.wasm b.component.wasm -o fused.wasm

wasm-opt -Os fused.wasm -o fused.opt.wasm
# → [wasm-validator error in module] unexpected false: multiple memories
#    require multimemory [--enable-multimemory], on memory
#   Fatal: error validating input
```

`wasm-opt` only proceeds with `--enable-multimemory`. `synth compile --cortex-m` has no single flat address space for multiple memories.

The fix already exists in meld — `meld fuse --memory shared --address-rebase` produces a single-memory module that flows through `wasm-opt` and `synth` cleanly. The issue is purely that the **default** is the form most pipelines can't consume.

## Impact

The out-of-the-box `meld fuse a b -o fused.wasm` produces an artifact that the canonical optimise→AOT chain (`wasm-opt` → `synth`) refuses. A user following the happy path hits a wall on the very next tool.

## Suggested fix (any one)

1. **Default to `--memory shared`** — the form downstream tools accept; make `multi` the opt-in.
2. Keep the `multi` default but **emit a warning** on `meld fuse` that the output needs `--enable-multimemory` downstream and has no MCU lowering.
3. At minimum, document the downstream implication in `meld fuse --help` next to the `--memory` option.

Option 1 matches the principle of least surprise for the meld→loom→wasm-opt→synth pipeline meld is built for.

## Environment

- meld v0.8.1 (`cargo install`'d from `meld-cli`)
- wasm-opt 116, synth v0.1.0
- macOS arm64
- Found bringing up the falcon WASM→ARM pipeline (pulseengine/relay, falcon v0.6); `scripts/falcon-wasm-pipeline.sh` works around it with `--memory shared --address-rebase`.

avrabe:
## What

Completes #172. The interim warning shipped in #177; this PR ships the actual default change — but as a **conditional** default, because flipping straight to `--memory shared` (issue option 1) would be unsound: merger Bug #7 pins that shared memory breaks when any component uses `memory.grow`.

`MemoryStrategy::Auto` (new default for `FuserConfig` and `meld fuse`) resolves once, at the start of fusion:

| condition | resolution |
|---|---|
| any input core module contains `memory.grow` (or fails to parse — conservative) | multi-memory |
| fewer than two memories across inputs | multi-memory (output is already single-memory) |
| otherwise | shared memory + address rebasing |

If the shared-memory plan refuses the input (`MemoryStrategyUnsupported`), fusion retries as multi — **Auto's success domain is a strict superset of the old default's.** Explicit `--memory multi|shared` are untouched. Unresolved `Auto` reaching `Resolver`/`Merger` directly behaves as `MultiMemory` (the always-sound strategy; documented on `Resolver::with_strategy`).

## Measured evidence (the issue's own repro)

```
$ meld fuse a.component.wasm b.component.wasm -o fused.wasm
Memory strategy: shared + address rebasing (auto: no memory.grow in inputs) — single-memory output
$ wasm-opt -Os fused.wasm -o fused.opt.wasm        # wasm-opt 116, NO flags
(accepted)

$ meld fuse a.component.wasm b.component.wasm -o fused-multi.wasm --memory multi
$ wasm-opt -Os fused-multi.wasm -o out.wasm
[wasm-validator error in module] unexpected false: multiple memories require multimemory [--enable-multimemory]
```

In-CI proxy of the same oracle: `tests/auto_memory.rs` validates the fused output with `WasmFeatures::default()` and `MULTI_MEMORY` cleared (the wasm-opt feature set), plus a runtime behaviour check (B's data segment lands at its rebased base, one page in; A's region intact).

Worth noting: every real wit-bindgen fixture in the repo contains `memory.grow` (Rust dlmalloc grows the heap), so the entire existing runtime suite resolves to multi under Auto — zero behaviour change for realistic components; the shared path activates only for static-allocation (MCU-shaped) inputs. Full workspace suite: **503 passed, 0 failed**.

## Safety artifacts

- **UCA-M-11** — merger configured with shared strategy for growable inputs.
- **LS-M-7** (approved) — probe under-reporting → silent cross-component corruption; gate-runnable regression tests `ls_m_7_auto_keeps_multi_when_input_grows` (integration) and `ls_m_7_malformed_input_counts_as_grow` (unit).
- **SR-37** (implemented, milestone v0.22.0) — the resolution contract, mitigates LS-M-7, tracked-by #172.

## Tier-5 note

`resolver.rs` is touched: four exhaustive `match self.memory_strategy` sites gain a `| MemoryStrategy::Auto` arm aliasing the existing `MultiMemory` behaviour, plus a doc comment on `with_strategy`. No decision logic changes. Mythos clean-room pass to follow as a PR comment + `mythos-pass-done` label.

## Also

Fixes a pre-existing CLI panic: the size-reduction stat computed `input - output` unsigned and underflowed whenever the fused output is larger than tiny inputs (reported as a signed delta now).

Closes #172.


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `meld-core/src/resolver.rs`

**Scope.** Delta on `main...feat/172-memory-auto`: doc comment on `Resolver::with_strategy` + four match arms extended `MultiMemory | Auto` (resolver.rs:1564, 2689, 3466, 3507). Cross-file consumers (lib.rs, merger.rs, component_wrap.rs, p3_stream.rs, memory_probe.rs, rewriter.rs) read to confirm/refute.

**Method.** Clean-room: enumerated every `memory_strategy` read in resolver.rs (exactly the four changed sites + constructors; nothing passes the raw strategy onward); traced all `config.memory_strategy` consumers in lib.rs for ordering vs. `resolve_auto_memory_strategy`; audited every `==`/`!=` strategy comparison in merger.rs and component_wrap.rs; wrote two PoC integration tests and ran them against the unfixed branch — **both failed, i.e. both hypotheses confirmed**.

### Provable findings — both FIXED in 8d98608

**Finding A — stale `Auto` resolution after `add_component` (lib.rs, UCA-M-11).** `fuse_with_stats` resolved `Auto` in place and never re-probed, so fuse → `add_component`(growing) → fuse reused the stale "shared" resolution and surfaced `UnsupportedFeature("memory.grow not supported with address rebasing")` (rewriter.rs guard) — violating the PR's own "Auto accepts everything multi accepts" contract. *Severity: contract violation / availability, not silent corruption (the rewriter guard fires because stale-shared always carries rebasing).*
**Fix:** the requested strategy is captured in `Fuser::requested_memory` and restored before resolution on every fuse, so resolution is re-derived from the current component set. Pinned by `ls_m_7_auto_reprobes_after_add_component` (gate-runnable). LS-M-7 causal factor 3 updated to record the falsified "mitigated" claim.

**Finding B — `Merger` split-brained on unresolved `Auto` (merger.rs, direct API only).** Merger's `== SharedMemory` sites treated Auto as multi but its `==`/`!= MultiMemory` sites treated it as shared: `Merger::new(Auto, false)` produced multi-memory layout with shared-style export dedup — the second component's memory export silently dropped (`["fa","fb","memory"]` vs `["fa","fb","memory","memory$0"]`). Unreachable via `Fuser` (resolution precedes construction; all lib.rs consumers verified post-resolution), reachable via the public `Merger::new`. The new resolver doc comment's claim of merger consistency was false as written.
**Fix:** `Auto → MultiMemory` normalized at all three public entry points (`Merger::new`, `Resolver::with_strategy`, `component_wrap::wrap_as_component`) so the consumers can never disagree; resolver doc corrected; the `| Auto` match arms retained as a compiler-enforced backstop. Pinned by `test_merger_new_normalizes_auto_to_multi_memory` and `test_with_strategy_normalizes_auto_to_multi_memory`.

### Verified-sound (checked and held)
- resolver.rs has exactly four strategy decision points; arm bodies byte-identical for Auto and MultiMemory (unit variants, no bindings/guards); p3_stream/resource_graph receive resolved values, never the raw strategy.
- Every lib.rs consumer of `config.memory_strategy` (Resolver/Merger construction, component_wrap, `build_index_maps_for_module` in adapter wiring + task-return shims, attestation/stats labels) sits inside the private `fuse_with_stats_resolved`, reachable only post-resolution.
- The shared-refusal fallback retry is sound: `fuse_with_stats_resolved` takes `&self`, no interior mutability; the retry run is equivalent to a fresh run.
- A first-run probe-pass/rewriter-fail divergence is not constructible — both scan the same input code-section operators.

**Conclusion:** the resolver.rs delta is sound and internally consistent; both blast-radius findings were proven by failing tests and are fixed + regression-pinned in `8d98608`. Post-fix workspace suite: 506 passed, 0 failed.

PR #236, opened 2026-06-11
feat(component-wrap): internalise satisfied imports across separate inputs (#212.1)

### LINKED ISSUE #212: Multi-component fusion gaps surfaced by golden-e2e (exports dropped / cross-link not internalised / bare-world result type)

avrabe:
v0.28.0 scope — meld's headline capability (closes #212).

## Diagnosis

The core level was already internalised (resolver wires consumer→provider; fused module carries no import of the satisfied interface; `imports_resolved: 1`, adapter generated). The failure was isolated to `component_wrap` replaying the consumer's component-level import declarations verbatim — the wrapped component still *demanded* `golden:math/lib@0.1.0` (observed live: wasmtime "matching implementation was not found in the linker").

## Fix

Derive the output's import surface from the **fused module's remaining imports**, not the source's declarations: drop an instance import IFF its name was imported by an input core module AND is absent from the fused module's remaining imports. WASI/external imports remain by construction. Survivors renumbered; the map applied at every instance-index consumer (resolution aliases — with a loud error on any impossible dropped-instance reference — and the component-instance base).

## Oracle

The acceptance test authored WITH the issue (`tier_b_separate_inputs_internalise_link`, `#[ignore]`d since #213) is **un-ignored and passes**: fused `consumer.wasm + provider.wasm --component` instantiates standalone with zero linker definitions; `compute()==42 ∧ greet()==7` equals the host-linked golden. Gate alias `ls_cp_6_separate_inputs_internalise_link` (LS-CP-6 approved, SR-39 implemented). Full suite: 531 passed / 0 failed (wit-bindgen WASI fixtures pin the dual failure — no wrong drops).

## Tier-5

`component_wrap.rs` changed. Mythos clean-room pass to follow as comment + label before merge.

### DIFF
--- a/meld-core/src/component_wrap.rs
+++ b/meld-core/src/component_wrap.rs
@@ -835,6 +835,18 @@ fn assemble_component(
         .flat_map(|m| m.imports.iter())
         .map(|imp| imp.module.as_str())
         .collect();
+    // INVARIANT (Mythos pass, PR #236): `instance_renum` keys are import
+    // ordinals but are consumed with component-instance indices from
+    // `component_instance_defs` — these coincide only while import
+    // instances occupy the contiguous LEADING slots of the component
+    // instance space, which every real producer (wit-component, wac,
+    // wit-bindgen) satisfies. Likewise the depth-0 replay only filters
+    // the Import section; a depth-0 Alias/Type section referencing an
+    // import instance by index would dangle after a drop — unreachable
+    // from real producers (their instance-referencing aliases come after
+    // the first core module and are not captured as depth-0). If either
+    // assumption breaks, the loud dropped-instance error below fires
+    // rather than emitting a wrong index.
     let mut dropped_import_names: std::collections::BTreeSet<String> = Default::default();
     let mut instance_renum: std::collections::HashMap<u32, u32> = Default::default();
     {
--- a/meld-core/src/component_wrap.rs
+++ b/meld-core/src/component_wrap.rs
@@ -814,15 +814,103 @@ fn assemble_component(
     let n = fused_info.func_imports.len();
 
     // -----------------------------------------------------------------------
+    // 0. Internalised-import elimination (#212.1). An instance import in
+    //    the wrapped component is needed IFF some import remaining in the
+    //    FUSED core module references it (module name == instance name).
+    //    An interface that an input module imported but the fused module
+    //    no longer does was satisfied internally by the resolver — its
+    //    component-level import declaration must be dropped, or the
+    //    wrapped component demands an implementation the linker cannot
+    //    have. Dropping shifts later instance indices, so a renumber map
+    //    is threaded to every consumer below.
+    // -----------------------------------------------------------------------
+    let remaining_modules: std::collections::BTreeSet<&str> = fused_info
+        .func_imports
+        .iter()
+        .map(|(m, _, _)| m.as_str())
+        .collect();
+    let original_modules: std::collections::BTreeSet<&str> = all_components
+        .iter()
+        .flat_map(|c| c.core_modules.iter())
+        .flat_map(|m| m.imports.iter())
+        .map(|imp| imp.module.as_str())
+        .collect();
+    let mut dropped_import_names: std::collections::BTreeSet<String> = Default::default();
+    let mut instance_renum: std::collections::HashMap<u32, u32> = Default::default();
+    {
+        let mut old_idx = 0u32;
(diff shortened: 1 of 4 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `component_wrap.rs`

**VERDICT: NO PROVABLE FINDING** — and the pass positively verified the path the shipped test omits.

**Positive evidence beyond the shipped oracle:** the acceptance test covers only the single-import drop (renumber map ends empty). The clean-room agent built a 2-import fixture (consumer importing `golden:math/lib` @ instance 0 AND `golden:util/log` @ instance 1; provider exporting only math) forcing the drop-one-keep-one **renumber 1→0** path: output validates under `WasmFeatures::all()` with exactly the util import surviving at index 0. (Fixture removed; tree clean.)

**Adversarial angles checked:**
1. *Wrong-drop via depth-0 type/alias references* — only the Import section is filtered; a depth-0 alias referencing a dropped instance would dangle, BUT the parser captures depth-0 sections only before the first core module and real producers emit instance-referencing aliases after it — not constructible from real inputs.
2. *Index-space mix* — `instance_renum` keys (import ordinals) vs consumption (component-instance indices) coincide only under the "imports are the contiguous leading instances" invariant — true for all real producers; a violation now fires the loud dropped-instance error rather than emitting a wrong index. No other consumer of the import-instance space exists beyond the two patched sites; `LocalResource`/`TaskBuiltin` resolutions embed no instance index.
3. *Non-func core imports* — `parse_fused_module` tracks only func imports, and fused modules import memory/tables from `meld:shim`, never interface names — not exploitable.
4. *Re-encode fidelity* — type-import bounds preserved (`Eq`/`SubResource`); value imports error; names/order preserved; instance imports allocate no type index so dropping shifts nothing else.
5. *Cross-input collisions* — global original∧¬remaining predicate; WASI protected by remaining func imports (73 wit_bindgen runtime tests green).

Suites: golden_e2e 7 ✓ (incl. the un-ignored acceptance), wit_bindgen_runtime 73 ✓, component_provenance 4 ✓, nested_component 2 ✓, osxcar_fusion 5 ✓.

Both latent fragilities are now pinned as an INVARIANT comment at the renumber-map construction (c8f41b4).

PR #239, opened 2026-06-11
feat(p3-bridge): cross-component stream-bridge emitter (#141, SR-33)

### LINKED ISSUE #141: P3 async — cross-component stream adapter (#94 sub-A)

avrabe:
v0.29.0 scope — the emitter the SR-33 audit proved was never built (closes #141; design recorded on the issue, ADR-3 path-N follow-up).

## What

`meld-core/src/p3_bridge.rs` (Tier-5, pre-registered in #238): when the fused module's `StreamPairGraph` is non-empty, meld emits a **bridge memory** (slot headers + 8×4096B rings), per-component **shims** for the five stream intrinsics with the component's memory index hardwired, and rewires intrinsic call sites to the shims. Dispatch on handle bit 31: locally-minted handles → in-module ring ops (no host crossing); foreign handles → the retained host imports; slot exhaustion → host `stream_new` fallback. ADR-2 contract preserved end-to-end (write returns accepted count, 0 = backpressure; read returns bytes, 0 = EOF only after writer-drop AND drain, −5 Pending; drop_writable half-closes).

## Oracles (runtime, through the REAL pipeline)

Five fixtures fuse two hand-built components (which trigger the real pair detector — untyped `stream` canon entries + fusion connection via a linked core-module export) and execute under wasmtime with host stubs that **trap on any tagged handle** plus zero-call counters:

- `ls_st_1_round_trip_local_stream_never_crosses_host`
- `cross_memory_chain_preserves_data_across_distinct_memories`
- `backpressure_partial_write_then_drain_then_resume`
- `ls_st_1_eof_only_after_writer_drop_and_drain`
- `slot_exhaustion_falls_back_to_host_stream`

All pass; workspace 537/0; clippy/fmt clean. LS-ST-1 approved; SR-33 → implemented (honest history note kept; ADR-3 "zero-copy" amended to "no host crossing, single copy").

## Notes & limitations (documented in-module)

Emitter runs before adapter wiring (adapter indices bake in `functions.len()`); slots retire on `drop_readable` (no reuse — degrades to host, never loses data); cancel ops stay host-routed (bridged ops never block); module-wide `stream_new` takeover is the accepted static-pairing over-approximation; shared-memory mode is the same codegen with immediates 0 but has no dedicated runtime fixture.

Mythos clean-room pass to follow as comment + label before merge.

### DIFF
--- a/meld-core/src/dwarf.rs
+++ b/meld-core/src/dwarf.rs
@@ -952,6 +952,8 @@ pub enum AdapterRole {
     AdapterShim,
     /// P3 async `task.return` result-global shim.
     TaskReturnShim,
+    /// Cross-component stream-bridge dispatch shim (#141, `p3_bridge`).
+    StreamBridge,
 }
 
 impl AdapterRole {
@@ -976,6 +978,7 @@ impl AdapterRole {
             AdapterRole::StartWrapper => 7,
             AdapterRole::AdapterShim => 8,
             AdapterRole::TaskReturnShim => 9,
+            AdapterRole::StreamBridge => 10,
         }
     }
 }
@@ -998,6 +1001,7 @@ impl From<crate::merger::SyntheticKind> for AdapterRole {
             crate::merger::SyntheticKind::StartWrapper => AdapterRole::StartWrapper,
             crate::merger::SyntheticKind::AdapterShim => AdapterRole::AdapterShim,
             crate::merger::SyntheticKind::TaskReturnShim => AdapterRole::TaskReturnShim,
+            crate::merger::SyntheticKind::StreamBridge => AdapterRole::StreamBridge,
         }
     }
 }
@@ -1741,6 +1745,7 @@ mod tests {
             AdapterRole::StartWrapper,
             AdapterRole::AdapterShim,
             AdapterRole::TaskReturnShim,
+            AdapterRole::StreamBridge,
         ];
         let mut seen = std::collections::BTreeSet::new();
         for role in all {
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -55,6 +55,7 @@ mod error;
 pub mod memory_probe;
 pub mod merger;
 pub mod p3_async;
+pub mod p3_bridge;
 pub mod p3_stream;
 pub mod parser;
 pub mod provenance;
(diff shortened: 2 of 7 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 files `p3_bridge.rs` (+ merger/dwarf/lib wiring)

**VERDICT: NO PROVABLE FINDING** on the raw-codegen risk surface, with the highest-value angles cleared by proof:

1. **Ring arithmetic** — wrapping-cursor fill/avail/offset correct under u32 wraparound (RING_CAP power-of-two divides 2³², offset continuity holds past 4 GiB); two-part copy bounds never cross slots; exact-boundary write (off+n==RING_CAP) has no off-by-one; oracle (c) exercises both wrap halves.
2. **Memory immediates** — bridge memory is provably the appended index (nothing appends after pipeline Step 2.6; the emitter-before-adapter-wiring ordering claim verified against lib.rs). The pass built a **shared-memory + address-rebasing PoC** — the configuration the runtime oracles never cover — and it validates and round-trips 300 bytes intact.
3. **Dispatch + slot state machine** — tagged slots always in-range; no path resets state to FREE (no reuse, "lose optimization never data" holds); `new_shim` branch depths and i64 packing correct.

**Three unverified suspicions** (reproducers built and removed; not provable as fusion-vs-host divergence because the in-repo host stubs share the same idealization):
- **S1** zero-length read with data returned the EOF sentinel → **fixed defensively** in cec435e (now Pending).
- **S2** write to a dropped slot silently accepted → **fixed defensively** in cec435e (now `AbiError::Closed` per ADR-2).
- **S3** module-wide `stream_new` takeover: a stream intended for a non-bridged host consumer would carry a tagged handle the host can't interpret — the design's documented static-pairing over-approximation; carried as a known limitation in the module docs and SR-33, revisit if a real host-stream-consuming fixture appears.

All 5 runtime oracles green post-hardening; workspace suite green; clippy/fmt clean.

PR #246, opened 2026-06-13
fix(merger): cabi_realloc export-name collision in multi-memory (#245)

### LINKED ISSUE #245: Robustness: cabi_realloc$N export-name collision between comp_idx and mem_idx suffix schemes (multi-memory)
## Summary

In multi-memory mode, the fused module's `cabi_realloc$N` export namespace is written by **two independent suffixing schemes that don't cross-check each other**, so they can emit the same name twice → `wasmparser` rejects the output with `duplicate export name \`cabi_realloc$1\` already defined`.

## Mechanism (by code reading — meld-core/src/merger.rs)

1. **Regular export dedup** (`merge_module`, ~lines 1918-1928): when a module's export name already exists, multi-memory mode renames it to `format!("{}${}", export.name, comp_idx)`. The guard checks whether **`export.name`** (the *original*, e.g. `"cabi_realloc"`) already exists — it does **not** check that the *renamed* `cabi_realloc${comp_idx}` is itself unique before pushing.

2. **Per-memory realloc export** (~lines 2065-2085): pushes `format!("cabi_realloc${}", mem_idx)` for `mem_idx > 0`. This path *does* guard (`if !merged.exports.iter().any(|e| e.name == export_name)`, line 2077) — but only against what's already present when it runs.

Both write into the shared `cabi_realloc$N` namespace, suffixed by **different quantities** (`comp_idx` vs `mem_idx`). When a colliding regular export's `comp_idx` equals a realloc's `mem_idx`, scheme (1) emits `cabi_realloc$<comp_idx>` with no new-name check and duplicates the name scheme (2) already produced.

## Trigger (caveat: synthetic)

I hit this while building an SR-17 reverse-transcode (UTF-16→UTF-8) test fixture: a *caller component containing two core modules* (a memory/realloc helper to satisfy a string `canon lower`, plus the real caller module), fused multi-memory against a UTF-8 callee. The two-modules-in-one-component layout made `comp_idx == 1` coincide with the helper's `mem_idx == 1`, yielding two `cabi_realloc$1` exports. meld **did** correctly read `caller_encoding = UTF16` and emit the reverse transcoding adapter (1 adapter, 1 import resolved) — the failure was purely this export-naming collision at output-validation time.

**Reachability honesty:** my repro uses a synthetic multi-core-module component. wit-component output typically has only the *main* module export `cabi_realloc` (shim modules don't), so a single component rarely has ≥2 realloc-signature exports. I have **not** confirmed a real-world (wit-component / wac) input that triggers this. Filing as a latent-robustness candidate, not a confirmed field bug — the code-level namespace collision is real regardless.

## Suggested fix

Make the two suffix schemes share a single collision-checked allocator, e.g. in scheme (1) verify the *renamed* `cabi_realloc${comp_idx}` (and generally `{name}${comp_idx}`) is unique and bump the suffix if taken, or give the per-memory realloc exports a disjoint namespace (e.g. `cabi_realloc@mem${mem_idx}`). `merger.rs` is Tier-5, so a fix needs a Mythos clean-room pass.

## Related
Surfaced during the SR-17 reverse-direction verification gap work (forward direction closed in #244). The reverse-direction *runtime* oracle remains blocked on hand-authoring wit-component's two-phase shim/fill instantiation (a string `canon lower` needs a core memory, which cycles with the caller module importing the lowered func) — the simpler memory-helper workaround is exactly what triggers this collision.

avrabe:
Closes #245. Found while building the SR-17 reverse-direction (UTF-16→UTF-8) runtime oracle.

## Bug
Multi-memory mode wrote `cabi_realloc$N` exports via two uncoordinated schemes:
- generic collision-dedup (`merge_module` ~L1918): `cabi_realloc$<comp_idx>`
- per-memory realloc path (~L2076): `cabi_realloc$<mem_idx>` — the convention `component_wrap.rs` consumes (it indexes reallocs by memory)

When a colliding export's `comp_idx` equalled another realloc's `mem_idx`, both minted the same name → `duplicate export name \`cabi_realloc$1\``.

## Fix
The generic dedup no longer mints `cabi_realloc` into the per-memory namespace. A colliding `cabi_realloc` is always `comp_idx ≥ 1` with `mem_idx ≥ 1`, so the per-memory path is guaranteed to publish it; the redundant (and never-consumed) comp_idx-suffixed copy is skipped. Component 0's realloc is the non-colliding first occurrence → still plain `cabi_realloc`.

## Verification
- New `test_sr17_utf16_to_utf8_supplementary_plane_transcoding`: fuses a UTF-16-lowering caller ("A😀") into a UTF-8 callee under wasmtime, asserts the surrogate pair decodes to UTF-8 `[F0 9F 98 80]` (byte sum 744). **Doubles as the #245 regression** — its two-core-module caller is the exact collision trigger — **and** closes SR-17's reverse-direction runtime gap.
- `test_cabi_realloc_suffixed_exports_generated` + `test_shared_memory_no_suffixed_realloc_exports` stay green.
- Workspace **541/0**; 73 wit_bindgen runtime tests (real realloc routing) green; clippy/fmt clean.

## Tier-5
`merger.rs` changed → Mythos clean-room pass to follow as a comment + `mythos-pass-done` label before merge.

### DIFF
--- a/meld-core/src/merger.rs
+++ b/meld-core/src/merger.rs
@@ -1915,6 +1915,25 @@ impl Merger {
             // would wire the fixup module to the wrong component's indirect
             // table. In shared-memory mode, first-wins dedup is correct
             // since all components share one memory.
+            // #245: `cabi_realloc` is named by the per-memory path below
+            // (`cabi_realloc$<mem_idx>`), which is the convention the P2
+            // wrapper consumes (component_wrap.rs looks reallocs up by
+            // memory index). The generic comp_idx-suffixed dedup here must
+            // NOT also mint into that namespace: when a colliding export's
+            // comp_idx coincides with another realloc's mem_idx the two
+            // schemes emit the same `cabi_realloc$N` twice and the output
+            // fails validation ("duplicate export name"). A colliding
+            // `cabi_realloc` is always comp_idx >= 1 with its own mem_idx
+            // >= 1, so the per-memory path is guaranteed to publish it;
+            // skip the redundant generic copy here. (Component 0's realloc
+            // is the non-colliding first occurrence and still flows through
+            // the else branch as plain `cabi_realloc`.)
+            if self.memory_strategy == MemoryStrategy::MultiMemory
+                && export.name == "cabi_realloc"
+                && merged.exports.iter().any(|e| e.name == export.name)
+            {
+                continue;
+            }
             let export_name = if self.memory_strategy == MemoryStrategy::MultiMemory
                 && merged.exports.iter().any(|e| e.name == export.name)
             {
--- a/meld-core/tests/adapter_safety.rs
+++ b/meld-core/tests/adapter_safety.rs
@@ -2157,3 +2157,307 @@ fn test_sr17_utf8_to_utf16_supplementary_plane_transcoding() {
          mis-encodes the surrogate pair."
     );
 }
+
+/// A caller component whose import is lowered with **UTF-16** encoding, so
+/// the resolver attributes `caller_encoding = UTF16`. Paired with the
+/// UTF-8 byte-summing callee (`build_callee_string_component`), this drives
+/// meld to emit the *reverse* transcoder (`emit_utf16_to_utf8_transcode`).
+///
+/// A string `canon lower` requires a core memory + realloc, which cycles
+/// with the caller module importing the lowered function. We break the
+/// cycle with a tiny memory/realloc helper module instantiated first; the
+/// lower references the helper's memory purely so the component parses,
+/// and meld reads the lower only as encoding metadata while fusing the
+/// real caller module by its core imports.
+///
+/// `code_units` are stored little-endian at offset 0 of the caller's
+/// memory, and `run` calls `process-string(0, code_units.len())` — the
+/// length is a UTF-16 code-unit count, per the canonical ABI.
+fn build_caller_utf16_lowering_component(code_units: &[u16]) -> Vec<u8> {
+    let mut data_bytes = Vec::new();
+    for cu in code_units {
+        data_bytes.extend_from_slice(&cu.to_le_bytes());
+    }
+    let code_unit_count = code_units.len() as i32;
+
+    let core_module = {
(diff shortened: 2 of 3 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `merger.rs`

**VERDICT: NO PROVABLE FINDING.** Adversarial falsification of the H-1 hazard (dropped/misrouted realloc) failed; the fix removes only spurious, wrapper-unread exports.

**Positive evidence (delta PoC, since removed):** a hand-built 4-component fusion (one component with no realloc, reversed instantiation order) compared origin/main vs the fix:
- main core exports: `cabi_realloc, $1, $0, $2, $3` (the `$0`/`$2` are comp_idx-keyed duplicates → the validation failure)
- fix core exports: `cabi_realloc, $1, $3` — one realloc per memory slot, no duplicates, validates.

**H-1 sub-cases checked:**
1. *Dropped realloc* — the per-memory path (`cabi_realloc$<mem_idx>`) runs unconditionally per module, independent of the `continue`; the fix only removes the generic copy, so it can never drop a `$<mem_idx>` export the per-memory path would emit.
2. *Multi-module / topological instantiation order* (comp_idx can reverse vs mem_idx) — plain `cabi_realloc` attaches to whichever module owns merged-memory slot 0; per-memory names key on the real merged `mem_idx`. comp_idx≠mem_idx divergence routes correctly per slot.
3. *Per-memory guard ordering* — fix relaxes the guard, no regression.
4. *Wrapper consumption* — confirmed `component_wrap.rs` reads only plain `cabi_realloc` (slot 0) and `cabi_realloc$<i>` for i=1..num_memories; comp_idx-suffixed names are genuinely dead.
5. *Name matching* — exact `== "cabi_realloc"` doesn't over/under-match `cabi_post_*` or pre-suffixed `cabi_realloc$N`.

Suites green on the branch: wit_bindgen_runtime 73/73 (real realloc routing), adapter_safety 7/7, multi_memory 5/5, realloc_safety 1/1, full `meld-core` 0 failures.

PR #248, opened 2026-06-13
fix(adapter): mid-string lone high surrogate → U+FFFD in UTF-16→UTF-8 (LS-P-16)

avrabe:
Closes the last open causal factor of LS-P-16 — found via falsification while building the SR-17 reverse-transcode oracles.

## Bug (confirmed by PoC)
`emit_utf16_to_utf8_transcode`'s surrogate-pair arm treated **any** unit after a high surrogate as a low surrogate, with no `0xDC00 ≤ cu2 < 0xE000` check. A high surrogate followed by a non-low unit (e.g. ASCII) underflows `cu2 - 0xDC00` (u32) → garbage code point → wrong 4-byte UTF-8 into callee memory (H-4.4, malformed-UTF-16 handling). This is the "no validation that the second code unit is a low surrogate" causal factor LS-P-16 documented; the v0.11 fix only handled *end-of-input* lone surrogates.

`[0xD800, 0x0041]` returned **500** (garbage) before; **684** (`U+FFFD + 'A'` = `EF BF BD 41`) after.

## Fix
Validate `0xDC00 ≤ cu2 < 0xE000` in the pair arm. Valid pair → unchanged. Invalid → emit U+FFFD, consume only the high surrogate (second unit reprocessed normally). One added scratch local (`cu2`); per-direction scratch count 5 → 6.

## Verification
- `test_sr17_utf16_to_utf8_midstring_lone_surrogate_replacement` (new runtime oracle, 684).
- All SR-17 transcode tests pass (forward/reverse non-BMP, end-of-input + mid-string lone surrogate). Workspace **543/0**; clippy/fmt clean.
- LS-P-16: all causal factors now mitigated and runtime-pinned.

## Tier-5
`fact.rs` changed → Mythos clean-room pass to follow as comment + `mythos-pass-done` label before merge.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -3015,8 +3015,10 @@ impl FactStyleGenerator {
             (StringEncoding::Utf8, StringEncoding::Utf8)
         );
 
-        // Scratch locals: src_idx, dst_idx, out_ptr, byte (+ code_point for UTF-8/16)
-        let scratch_locals: u32 = if needs_transcoding_locals { 5 } else { 0 };
+        // Scratch locals: src_idx, dst_idx, out_ptr, cu, code_point, and cu2
+        // (the second surrogate unit, used by the UTF-16→UTF-8 low-surrogate
+        // validation — LS-P-16 mid-string mitigation).
+        let scratch_locals: u32 = if needs_transcoding_locals { 6 } else { 0 };
         let post_return_base = param_count as u32 + scratch_locals;
 
         let mut local_decls: Vec<(u32, wasm_encoder::ValType)> = Vec::new();
@@ -3543,6 +3545,7 @@ impl FactStyleGenerator {
         let out_ptr_local = param_count as u32 + 2;
         let cu_local = param_count as u32 + 3;
         let cp_local = param_count as u32 + 4;
+        let cu2_local = param_count as u32 + 5;
 
         // Source reads (UTF-16) use caller_memory, destination writes (UTF-8) use callee_memory
         let src_mem16 = wasm_encoder::MemArg {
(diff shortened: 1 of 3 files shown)


avrabe:

**VERDICT: NO PROVABLE FINDING.** Correct against all five hazards; falsification failed, runtime PoCs confirm intended behaviour.

- **Local-count / post-return base** — `scratch_locals` 5→6: both the local declaration and `post_return_base = param_count + scratch_locals` derive from the same variable, so the shift is consistent; the 3 post-return save/restore sites sit above all 6 scratch locals; the other transcode direction declares one harmless unused local. wit_bindgen_runtime 73/73 (exercises both directions + post-return) green.
- **OOB read** — the mid-string Else is dominated by the outer `src_idx+1 >= input_len` guard (`>=` correct); cu2 read only when in-bounds.
- **Termination / reprocessing** — every branch advances src_idx ≥1; invalid-second-unit does +=1 and reprocesses. PoC `[0xD800, 0xD800, 0x0041]` → byte sum **1303** (`EF BF BD · EF BF BD · 41`), confirming consecutive lone surrogates each become U+FFFD.
- **Buffer bound** — no path emits >3 bytes per consumed unit (4-byte emission always consumes 2 units); `3 * input_code_units` holds.
- **Boundary values** — `cu2` 0xDBFF→FFFD, 0xDC00→pair, 0xDFFF→pair, 0xE000→FFFD via unsigned compares; leading 0xDC00 correctly not treated as high.

Suites green: adapter_safety 11/11, multi_memory 5/5, realloc_safety 1/1, wit_bindgen_runtime 73/73.

**Out-of-scope finding (pre-existing, not this delta):** a *leading lone LOW surrogate* (e.g. `[0xDC00, ...]`) falls into the BMP else-branch and is mis-encoded as malformed UTF-8 `ED B0 80` (PoC `[0xDC00,0x0041]` → 606) instead of U+FFFD. Same malformed-UTF-16 class, different code path; will be filed as a separate follow-up.

PR #252, opened 2026-06-13
fix(adapter): validate malformed UTF-8 in UTF-8→UTF-16 transcode (closes #251)

### LINKED ISSUE #251: UTF-8→UTF-16: forward transcoder does not validate malformed UTF-8 (invalid continuation/lead/overlong/surrogate/range)
## Summary
`emit_utf8_to_utf16_transcode` (meld-core/src/adapter/fact.rs) guards **truncation** (LS-P-19: a multi-byte lead whose continuations would read past end-of-input → U+FFFD, consume lead), but does **not** validate the *content* of multi-byte sequences. All three arms (2/3/4-byte) blindly mask continuation bytes with `& 0x3F` and lead bits, so malformed UTF-8 is silently decoded to wrong code points instead of the Canonical-ABI-mandated U+FFFD replacement (H-4.4 incorrect transcoding). This is the forward-direction analogue of the UTF-16→UTF-8 surrogate gaps fixed in v0.11 / #248 / #250.

## Confirmed (runtime PoC)
Fusing a UTF-8 caller passing `[0xC2, 0x41]` (2-byte lead `0xC2` then `'A'`, which is **not** a continuation byte) into a UTF-16 callee that sums received code units returned **129** = U+0081 (the transcoder computed `cp = (0xC2 & 0x1F)<<6 | (0x41 & 0x3F) = 0x81` and consumed both bytes). The Canonical-ABI-correct result is `U+FFFD + 'A'` → UTF-16 `[0xFFFD, 0x0041]` = **65598**.

## Scope (all unmitigated, by code reading)
1. **Invalid continuation bytes** — 2/3/4-byte arms don't check `(b & 0xC0) == 0x80` on each continuation (CONFIRMED above).
2. **Invalid lead bytes** — a lone continuation `0x80–0xBF` is treated as a 2-byte lead (`byte >= 0x80 && < 0xE0`); `0xF5–0xFF` take the 4-byte arm and yield out-of-range code points.
3. **Overlong encodings** — e.g. `[0xC0,0x80]` (overlong NUL), `[0xE0,0x80,..]` — not rejected.
4. **UTF-8-encoded surrogates** — `0xED 0xA0–0xBF ..` encodes U+D800–U+DFFF; not rejected.
5. **Out-of-range** — 4-byte sequences > U+10FFFF; not rejected.

## Fix considerations
The Unicode "maximal subpart of an ill-formed subsequence" rule governs how many bytes to consume per U+FFFD (what `String::from_utf8_lossy` / browsers do). The existing LS-P-19 truncation handling uses the simpler "emit U+FFFD, consume only the lead" convention — extending that is consistent and always makes progress, though it can emit more U+FFFDs than the strict maximal-subpart rule for some multi-continuation cases. A complete fix should: classify lead bytes (reject `0xC0,0xC1,0xF5–0xFF` and lone continuations), validate each continuation, and reject overlong/surrogate/out-of-range results — all → U+FFFD with a defined consume count, across all three arms.

This is a multi-faceted, spec-conformance-sensitive Tier-5 change (`fact.rs` → Mythos clean-room pass) that warrants focused work and a full runtime test matrix (the `build_caller_string_component_data` + `build_callee_utf16_string_component` helpers in tests/adapter_safety.rs make every case runtime-testable, mirroring the reverse-direction `test_sr17_utf16_to_utf8_malformed_surrogate_matrix`). Recommend a dedicated increment rather than bundling into unrelated work.

Surfaced while completing the reverse-direction malformed-surrogate hardening (#249/#250).

avrabe:
Closes #251. The symmetric forward-direction counterpart to the reverse malformed-surrogate hardening (#248/#250) — completes malformed-input handling for both transcode directions.

## Bug (confirmed by PoC)
`emit_utf8_to_utf16_transcode` guarded truncation (LS-P-19) but never validated multi-byte *content*: every arm masked continuation bytes and lead bits blindly. `[0xC2, 0x41]` (2-byte lead + ASCII non-continuation) decoded to U+0081 (sum 129) instead of `U+FFFD + 'A'` (65598).

## Fix
Each arm now → U+FFFD (consume lead only, per the established LS-P-19 convention) on: invalid lead (`<0xC2` covers lone continuations + overlong 0xC0/0xC1; `>=0xF5` out of range), invalid continuation byte (`(b&0xC0)!=0x80`), overlong (3-byte `cp<0x800`, 4-byte `cp<0x10000`), UTF-8-encoded surrogate (`cp∈[0xD800,0xE000)`), and out-of-range (`cp>0x10FFFF`). Valid fast paths and the LS-P-19 truncation guards are unchanged. Added one local (`cont_local`), reusing the free index 5 (mutually exclusive with the reverse transcoder per the dispatch match).

## Verification
- `test_sr17_utf8_to_utf16_malformed_matrix`: invalid continuation, lone-continuation lead, overlong, UTF-8 surrogate, out-of-range, plus valid 1/2/3/4-byte controls. Existing forward tests (ASCII, non-BMP) unchanged. Workspace **545/0**; clippy/fmt clean.
- **Mythos clean-room pass: NO PROVABLE FINDING** — boundary-value PoCs confirm the smallest/largest valid 2/3/4-byte code points (U+0080, U+07FF, U+0800, U+10000, U+10FFFF) and the surrogate-boundary points (U+D7FF, U+E000) pass without false-positive rejection, and truncated-invalid inputs don't OOB. (Full report in the next comment.)
- LS-P-20 approved; SR-17 traceability extended.

## Tier-5
`fact.rs` → Mythos pass done, `mythos-pass-done` applied.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -3015,9 +3015,14 @@ impl FactStyleGenerator {
             (StringEncoding::Utf8, StringEncoding::Utf8)
         );
 
-        // Scratch locals: src_idx, dst_idx, out_ptr, cu, code_point, and cu2
-        // (the second surrogate unit, used by the UTF-16→UTF-8 low-surrogate
-        // validation — LS-P-16 mid-string mitigation).
+        // Scratch locals: src_idx, dst_idx, out_ptr, cu/byte, code_point, and
+        // a 6th unit. In the UTF-16→UTF-8 direction the 6th unit is `cu2`
+        // (the second surrogate unit, used by the low-surrogate validation —
+        // LS-P-16 mid-string mitigation). In the UTF-8→UTF-16 direction the
+        // 6th unit is `cont` (a continuation byte being validated — #251
+        // malformed-UTF-8 hardening: invalid continuation / lead / overlong /
+        // surrogate / out-of-range detection). Only one direction's transcoder
+        // runs per generated function, so the two uses never alias.
         let scratch_locals: u32 = if needs_transcoding_locals { 6 } else { 0 };
         let post_return_base = param_count as u32 + scratch_locals;
 
@@ -3130,12 +3135,14 @@ impl FactStyleGenerator {
             }
         };
 
-        // Scratch locals: src_idx, dst_idx (code units), out_ptr, byte, code_point
+        // Scratch locals: src_idx, dst_idx (code units), out_ptr, byte,
+        // code_point, and cont (a continuation byte being validated — #251).
         let src_idx_local = param_count as u32;
         let dst_idx_local = param_count as u32 + 1;
         let out_ptr_local = param_count as u32 + 2;
         let byte_local = param_count as u32 + 3;
         let cp_local = param_count as u32 + 4;
+        let cont_local = param_count as u32 + 5;
 
         // Source reads use caller_memory, destination writes use callee_memory
         let src_mem8 = wasm_encoder::MemArg {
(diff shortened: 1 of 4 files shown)


avrabe:

**VERDICT: NO PROVABLE FINDING.** All hazards (H-4.4 mis-transcode, H-2 OOB, H-1 trap, non-termination) adversarially falsified and survived.

- **Local aliasing** — `scratch_locals=6`; the encoding-pair dispatch match emits exactly one transcoder per function, so `cont_local` (forward, idx 5) and `cu2_local` (reverse, idx 5) never coexist. `post_return_base = param_count + 6` unchanged, save/restore strictly above all scratch locals.
- **OOB** — lead rejections (`<0xC2`, `>=0xF5`) fire before any continuation read; continuation reads sit inside the LS-P-19 truncation guard. PoCs `[0xE0,0x41]`→65598 and `[0xF0,0x80]`→131066, no OOB.
- **Termination** — every malformed path `src_idx += 1`; 0xC2/0xF4 valid, 0xC1/0xF5 rejected.
- **Boundary exactness — runtime PoC table (all pass):**

| input | code point | sum |
|---|---|---|
| C2 80 | U+0080 (smallest 2-byte) | 128 |
| DF BF | U+07FF (largest 2-byte) | 2047 |
| E0 A0 80 | U+0800 (smallest 3-byte, above overlong) | 2048 |
| ED 9F BF | U+D7FF (below surrogates) | 55295 |
| EE 80 80 | U+E000 (above surrogates) | 57344 |
| EF BF BF | U+FFFF (largest 3-byte) | 65535 |
| F0 90 80 80 | U+10000 (smallest 4-byte) | pair |
| F4 8F BF BF | U+10FFFF (largest valid) | pair |

  The overlong/surrogate/range rejections do not false-positive on the smallest/largest valid code points or the surrogate boundaries.
- **Load-bearing** — origin/main has zero `cont_local`/`0xC2`/`0xF5`/`0x10FFFF`; the matrix test fails pre-fix (e.g. `[0xED,0xA0,0x80]`→U+D800 instead of 3×FFFD).

Suites green: adapter_safety 12/12, wit_bindgen_runtime 73/73, multi_memory 5/5.

Issue #253, opened 2026-06-13
Transcoding: cross-encoding pairs involving Latin-1/CompactUTF16 silently fall back to a raw byte copy (no transcoding)


avrabe:
## Summary
The FACT adapter's string-transcoding dispatch (`generate_transcoding_adapter` / the `match (caller_enc, callee_enc)` at meld-core/src/adapter/fact.rs:3045) explicitly handles only `(Utf8,Utf8)`, `(Utf8,Utf16)`, `(Utf16,Utf8)`, and `(Latin1,Utf8)`. Everything else hits `_ => { direct call / raw copy }` (fact.rs:3073-3079). But `needs_transcoding()` is `caller_string_encoding != callee_string_encoding` (fact.rs:5291) — **unconditionally true for every mismatched pair** — and the adapter is generated whenever `crosses_memory && needs_transcoding()` (fact.rs:616). So for the unhandled CROSS-encoding pairs the adapter is generated but performs **no transcoding** — it copies the source bytes verbatim into the destination under a different encoding.

Affected (generated-but-not-transcoded) pairs:
- `(Latin1, Utf16)` — Latin-1 byte 0xE9 ('é') should become UTF-16 code unit 0x00E9; a raw copy reinterprets the byte buffer as UTF-16 → wrong data AND wrong length.
- `(Utf8, Latin1)` and `(Utf16, Latin1)` — down-conversion (code points > 0xFF unrepresentable); raw copy is wrong.
- `(Utf16, Utf16)` and `(Latin1, Latin1)` also hit `_`, but there a raw copy is CORRECT (same encoding), so the fallback is only wrong for the cross-encoding cases.

## Reachability
`CompactUtf16` is mapped to `StringEncoding::Latin1` (fact.rs ~line 53). CompactUTF16 is a common real encoding, so a CompactUTF16 component talking to a UTF-16 component produces `(Latin1, Utf16)` and silently mis-transcodes **well-formed** input — distinct from (and arguably more severe than) the malformed-input gaps just fixed (#251/#249), which at least required malformed input. Confirmed reachable by code analysis (needs_transcoding admits the pair; `_` copies); a runtime PoC via a CompactUTF16-lowering caller is the recommended confirmation.

## Suggested fix
[...]

PR #254, opened 2026-06-13
fix(adapter): fail loud on unsupported cross-encoding transcode (#253 inc 1)

avrabe:
First increment on #253 — converts a silent miscompile into a loud failure.

## Bug
`generate_transcoding_adapter`'s `_` catch-all emitted a verbatim byte copy for every unhandled encoding pair. Since `needs_transcoding()` is `caller != callee` (always true for mismatches) and the adapter is generated on `crosses_memory && needs_transcoding`, the unimplemented cross-encoding pairs — `(Latin1,Utf16)`, `(Utf8,Latin1)`, `(Utf16,Latin1)`; CompactUTF16 maps to Latin1 — **silently mis-transcoded well-formed input** (H-4.4): bytes copied raw under a different encoding.

## This increment
Split the catch-all: same-encoding (`Utf16→Utf16`, `Latin1→Latin1`) keeps the correct verbatim copy; different-encoding-unhandled returns `Err(AdapterGeneration)`. meld now **refuses rather than corrupts**. The three missing transcoders (`Latin1↔Utf16`, `*↔Latin1`) remain tracked on #253 as the complete fix.

## Verification
- `test_253_unsupported_cross_encoding_transcode_fails_loud`: a CompactUTF16-lowering caller (→Latin1) into a UTF-16 callee — fusion now returns `Err`. **Confirms #253 reachability end-to-end** (the resolver does produce the `(Latin1, Utf16)` pair).
- No existing fixture relied on the silent path: full suite **546/0**, incl. 73 wit_bindgen runtime tests. clippy/fmt clean.

## Tier-5
`fact.rs` → Mythos clean-room pass to follow as comment + `mythos-pass-done` label before merge.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -3070,12 +3070,29 @@ impl FactStyleGenerator {
                 self.emit_latin1_to_utf8_transcode(&mut func, param_count, target_func, options);
             }
 
-            _ => {
-                // Other combinations - fall back to direct call for now
+            (caller_enc, callee_enc) if caller_enc == callee_enc => {
+                // Same encoding (e.g. Utf16→Utf16, Latin1→Latin1): no
+                // transcoding needed, a verbatim copy is correct.
                 for i in 0..param_count {
                     func.instruction(&Instruction::LocalGet(i as u32));
                 }
                 func.instruction(&Instruction::Call(target_func));
+            }
+
+            (caller_enc, callee_enc) => {
+                // #253: a cross-encoding pair with no transcoder implemented
+                // (e.g. Latin1→Utf16, Utf8→Latin1, Utf16→Latin1 — note
+                // CompactUTF16 maps to Latin1). Previously this fell through
+                // to a verbatim byte copy, silently MIS-transcoding
+                // well-formed input (H-4.4). Fail loudly instead of emitting
+                // a wrong adapter — the per-pair transcoders are tracked as
+                // follow-up work. Reaching here means the resolver requested
+                // a transcoding meld cannot yet perform correctly.
+                return Err(crate::Error::AdapterGeneration(format!(
+                    "unsupported string transcoding: {caller_enc:?} -> {callee_enc:?} \
+                     (no transcoder implemented; a verbatim copy would silently \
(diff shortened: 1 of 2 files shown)


avrabe:

**VERDICT: NO PROVABLE FINDING.** All four falsification hypotheses refuted.

**Full 9-pair enumeration** (StringEncoding = {Utf8, Utf16, Latin1}, derives Eq):

| pair | arm | old → new |
|---|---|---|
| Utf8→Utf8 | explicit | copy → copy (unchanged) |
| Utf8→Utf16 / Utf16→Utf8 / Latin1→Utf8 | explicit | transcode → transcode (unchanged) |
| Utf16→Utf16 / Latin1→Latin1 | `a==b` | copy → copy (preserved) |
| Utf8→Latin1 / Utf16→Latin1 / Latin1→Utf16 | Err | **silent wrong copy → error** |

The only newly-erroring pairs are the three that previously got a silent wrong verbatim copy. No previously-correct pair errors; no explicit arm falls through.

- **No false-positive copy** — `a==b` compares one enum by derived Eq; CompactUTF16 maps to Latin1 *upstream* of this match, so only Utf8/Utf16/Latin1 reach here (no aliasing).
- **Clean error propagation** — `Err` → `generate_transcoding_adapter` → `generate_adapter` (`?`) → `generate` (`?`) → `fuse` (lib.rs:577 `?`); thiserror variant, no panic; fires during adapter generation before any module bytes are written; zero `.unwrap()`/`.ok()` on `.generate()` in non-test code.
- **Complete handled set** — exactly three transcoder emit fns exist; the error arm shadows no half-built transcoder.
- **Non-vacuity** — reverting the fix makes `test_253_unsupported_cross_encoding_transcode_fails_loud` fail with "Got Ok".

Suites green: adapter_safety 12, wit_bindgen_runtime 73, multi_memory 5, realloc_safety 1, fact.rs unit 29 (incl. Latin1→Utf8 + CompactUTF16→Latin1 mapping). 0 failures.

Note (not a finding): the `a==b` arm is defensive — `needs_transcoding()` gates this function so same-encoding pairs don't normally reach it; harmless and correct if they did.

PR #255, opened 2026-06-13
feat(adapter): Latin-1 → UTF-16 string transcoder (#253 inc 2)

avrabe:
Second increment on #253 — enables the well-defined cross-encoding direction.

## What
Implements `(Latin1, Utf16)` transcoding (previously refused by the #253 inc-1 fail-loud guard). Latin-1 → UTF-16 is total and unambiguous: each byte 0x00-0xFF zero-extends to one UTF-16 code unit U+0000-U+00FF (no surrogates, no malformed forms). `emit_latin1_to_utf16_transcode`: alloc `2*byte_len` via callee cabi_realloc with LS-A-7 guards (overflow trap before the mul; OOM null trap), loop `I32Load8U` caller-mem → `I32Store16` callee-mem, call target with byte_len as the code-unit count. 3 scratch locals (within the existing budget of 6). So a CompactUTF16 caller (→Latin1) → UTF-16 callee now transcodes correctly.

The down-conversion pairs (`Utf8→Latin1`, `Utf16→Latin1`) stay fail-loud — genuinely ambiguous (code points > 0xFF; CompactUTF16 tagged representation) and tracked on #253 for a dedicated model.

## Verification
- `test_sr17_latin1_to_utf16_transcoding`: empty, NUL, 0x7F/0x80/0xFF boundaries, multi-byte (`[0x41,0xE9,0xFF]`→553). Renamed `test_253_utf16_to_latin1_transcode_fails_loud` keeps the down-conversion guard covered. Workspace **547/0**; wit_bindgen_runtime 73/73; clippy/fmt clean.
- **Mythos clean-room pass: NO PROVABLE FINDING** — exact buffer bound, no OOB read/write, correct caller/callee memory indices, zero-extend (not sign-extend), byte_len code-unit count, empty-input non-trap confirmed live, dispatch ordering correct. (Full report next comment.)

## Tier-5
`fact.rs` → Mythos pass done, `mythos-pass-done` applied.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -3070,6 +3070,12 @@ impl FactStyleGenerator {
                 self.emit_latin1_to_utf8_transcode(&mut func, param_count, target_func, options);
             }
 
+            (StringEncoding::Latin1, StringEncoding::Utf16) => {
+                // #253 inc 2: Latin-1 to UTF-16 is total and unambiguous — each
+                // byte 0x00–0xFF zero-extends to one code unit U+0000–U+00FF.
+                self.emit_latin1_to_utf16_transcode(&mut func, param_count, target_func, options);
+            }
+
             (caller_enc, callee_enc) if caller_enc == callee_enc => {
                 // Same encoding (e.g. Utf16→Utf16, Latin1→Latin1): no
                 // transcoding needed, a verbatim copy is correct.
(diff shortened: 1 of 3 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `fact.rs` (emit_latin1_to_utf16_transcode)

**VERDICT: NO PROVABLE FINDING.** H-4.4 / H-2 / H-1 / termination all falsified, by static inspection against the proven sibling transcoders and by live wasmtime fusion+exec.

- **Output bound** — alloc `2*byte_len`; last `I32Store16` at `(byte_len-1)*2` spans to `2*byte_len` exactly (no off-by-one); overflow guard fires before the `i32.mul`.
- **OOB read** — loop strictly `src_idx < byte_len` (`I32GeU`+`BrIf`); reads `[0,byte_len)`, never `ptr+byte_len`.
- **Memory indices** — `I32Load8U` from `caller_memory`, `I32Store16` to `callee_memory` (matches `emit_latin1_to_utf8`/`emit_utf8_to_utf16`); verified under live MultiMemory.
- **No sign-extend/truncation** — `I32Load8U` (zero-extend); stored unit is exactly `0x00bb`, never a surrogate or >0xFF (`[0x80]→128`, `[0xFF]→255` live).
- **Code-unit count** — target called with `byte_len` (1 byte → 1 code unit), not `byte_len*2`.
- **Empty input** — `byte_len=0`: no overflow trap, `2*0=0` alloc, loop never runs; `[]→0` passes live (no spurious trap). The LS-A-7 leg-b null-trap convention is shared/unchanged.
- **Dispatch ordering** — `(Latin1,Utf16)` precedes same-encoding-copy and fail-loud; `(Utf16,Latin1)` still errors, `(Latin1,Latin1)` still copies (both confirmed live).

PoCs (live wasmtime, MultiMemory): `[]→0, [0x00]→0, [0x7F]→127, [0x80]→128, [0xFF]→255, [0x41,0xE9,0xFF]→553, [0xFF,0xFF]→510`. Suites: adapter_safety 13/13, wit_bindgen_runtime 73/73, multi_memory 5/5.

PR #257, opened 2026-06-13
fix(segments): preserve element-segment reference types (closes #256, LS-A-21)

### LINKED ISSUE #256: resource_graph: definer attribution by interface-name string-match can override the authoritative Defines edge (and collides for multi-definer)
Found by a Mythos discovery pass on `meld-core/src/resource_graph.rs` (Tier-5; **no Kani/Rocq coverage** — confirmed). Reproduced at the public-API level with failing probe tests; **end-to-end fusion divergence NOT established** (see Reachability) — filing as a strong candidate, below the full Mythos bar but well above silence.

## Defect
`ResourceGraph::resource_definer(iface, rn)` / `defines_resource` attribute a resource's definer to a component by **interface-name string match alone**, and the string-match "terminal exporter" second pass (`resource_graph.rs:300-308`) **overrides** the authoritative first-pass `Defines`-edge attribution (the second pass uses `.insert()` at :307 and runs before the merge's `.or_insert` at :313-321). Two failure modes:

**A. Mis-attribution to a non-owning component.** If component *D* has the `Defines` edge (canonical `ResourceRep`) for resource `R` under `iface:x/foo`, but an importer's `iface:x/foo` resolves to a different, non-owning component *E* (e.g. a plain func provider that happens to export the same interface name and does not re-import it), the second pass attributes `R`'s definer to *E*. Probe A (ran, FAILED on unfixed code): `resource_definer = Some(E)` (wrong) and `defines_resource(E)` and `defines_resource(D)` BOTH return true (contradictory).

**B. Multi-definer cache collision (#245-class).** `definer_cache: HashMap<(String,String), Option<usize>>` (:55) is keyed without component disambiguation. Two components each `ResourceRep`-ing `R` under the same interface collapse to one slot; which survives depends on `resolved_imports`/HashMap iteration order. Probe B (ran, FAILED): non-deterministic definer, both report `defines_resource = true`.

## Load-bearing consumption
Not cosmetic: `resolver.rs:1755` uses `resource_definer` to allocate the per-component resource **handle table**; a wrong definer allocates it for the wrong component and starves the real one — precisely the failure the `resolver.rs:1748-1750` comment warns about (un-translated handle reaches another's code → `Option::unwrap()` on `None`). It also gates the `callee_defines_resource` downgrade (`resolver.rs:2952-2953`, `3254-3255`) selecting the handle-translation path emitted into the fused module. Maps to the **fuser** UCA family (wrong cross-component handle routing); same root cause as #245 (two namespaces merged without an ownership cross-check) and the LS-A-17/18 resource-graph attribution lineage. Would warrant a new **LS-A-19**.

## Fix direction (intricate — note the subtlety)
The second pass (:300-308) is a deliberate FALLBACK for resources with no `Defines` edge ("canonical_functions empty after flattening", :277-278), so a naive "require a Defines edge" guard would break its intended purpose. The correct fix makes **`Defines`-edge attribution authoritative**: when any component has a `Defines` edge to `(iface, rn)` (i.e. `defines_candidates` contains `(_, iface, rn)`), the string-match second pass must NOT override it with a different component — the second pass should only attribute resources that have NO `Defines` edge anywhere. Concretely: reorder so the `defines_candidates` merge is authoritative and the second pass only `.or_insert`s gaps, OR guard the second-pass `definer_cache.insert` with "no `Defines` candidate exists for this `(iface, rn)`". Separately, for (B), `definer_cache` needs to represent/flag genuine multi-definer collisions rather than silently pick one.

## Reachability (honest — why this isn't yet a closed finding)
The triggering topology (two components exporting the same interface name; an importer resolving to the non-owning one — plausibly via `resolver.rs:1981-1982`, "first export from a different component" by index order) was exercised only with **hand-built `ParsedComponent`s**, not a spec-valid multi-component WAT run through the real parser→compose→fuse→execute. Whether the parser/compose ever emits this exact shape (vs deduping/rejecting it upstream) is unverified. The required next step before fixing is a `compose→fuse→execute` differential proving wrong handle-table routing or a runtime `unwrap`-on-`None`. The fix is low-risk and principled regardless, but it touches a load-bearing path, so it should land WITH reachability confirmation + the wit_bindgen resource runtime suite as a regression guard (a subtly-wrong fix could break correct resource fusions — worse than the latent bug).

This is the discovery half of the Mythos pipeline; recommend a dedicated increment (confirm reachability → fix authoritative attribution + multi-definer collision → Mythos delta-pass → LS-A-19).

avrabe:
Closes #256 (the element-segment branch of the resource/segment discovery work). Found by a Mythos discovery pass on `segments.rs`.

## Bug
`parse_element_segments` bucketed every expression-encoded element ref type via `if is_func_ref() { FUNCREF } else { EXTERNREF }`. A concrete typed function reference `(ref null $t)` lost its heap type AND its concrete type index (never carried into `ParsedElementSegment` → the `maps.types` remap could never run); GC abstract heap types were likewise flattened. The element ITEMS kept their real funcref type while the segment was re-declared externref, so the re-emitted module is **invalid** — `wasmparser` rejects it (`type mismatch: expected externref, found (ref $type)`). meld's parser enables gc + function-references by default and the merger calls parse→reindex→encode verbatim, so valid input modules became invalid output (**H-3**).

## Fix
`convert_ref_type` carries the full `RefType` faithfully (nullability + abstract heap type via `convert_abstract_heap_type` + concrete module index); `remap_ref_type` remaps a concrete heap-type index through `maps.remap_type` during reindex — mirroring the existing `RefNull` const-expr handling.

## Verification
- `test_typed_ref_element_segment_roundtrips_and_validates` — the discovery PoC: `(elem (ref null 0) (ref.func 0))` now parses → reindexes → re-encodes → **validates** (failed pre-fix). Plus `test_reindex_element_segment_preserves_and_remaps_concrete_element_type` (struct-level, index remapped 0→4, not collapsed).
- Workspace **549/0** (verified via exit code, not pass-count sum); clippy/fmt clean.

## Notes
The symmetric `is_func_ref()-else-externref` site in `parser.rs::convert_ref_type` (value/global types) is a related latent gap, out of scope here — worth a follow-up.

## Tier-5
`segments.rs` → Mythos clean-room pass to follow as comment + `mythos-pass-done` label before merge.

### DIFF
--- a/meld-core/src/segments.rs
+++ b/meld-core/src/segments.rs
@@ -210,11 +210,15 @@ pub fn parse_element_segments(module: &CoreModule) -> Result<Vec<ParsedElementSe
                 (RefType::FUNCREF, ElementItems_::Functions(funcs))
             }
             ElementItems::Expressions(ref_type, expr_reader) => {
-                let element_type = if ref_type.is_func_ref() {
-                    RefType::FUNCREF
-                } else {
-                    RefType::EXTERNREF
-                };
+                // Carry the FULL ref type faithfully — bucketing every
+                // non-funcref to externref (the prior behaviour) dropped
+                // concrete typed function references `(ref $t)` and GC
+                // abstract heap types (struct/array/eq/any/i31/none), and
+                // lost the concrete type index entirely, producing a fused
+                // module that fails wasm validation ("type mismatch:
+                // expected externref, found (ref $type)"). The concrete
+                // index is remapped during reindexing (see remap_ref_type).
+                let element_type = convert_ref_type(ref_type);
                 let mut exprs = Vec::new();
                 for expr in expr_reader {
                     let expr = expr?;
@@ -288,6 +292,40 @@ pub fn parse_data_segments(module: &CoreModule) -> Result<Vec<ParsedDataSegment>
 }
 
 /// Reindex an element segment with new index mappings
+/// Faithfully convert a `wasmparser::RefType` to a `wasm_encoder::RefType`,
+/// preserving nullability, abstract heap types, and concrete type indices.
+/// (Concrete indices are module-level here; they are remapped later by
+/// [`remap_ref_type`].) Replaces the lossy funcref/externref bucketing that
+/// previously corrupted typed-ref and GC element segments.
+fn convert_ref_type(rt: wasmparser::RefType) -> RefType {
+    let heap_type = match rt.heap_type() {
+        wasmparser::HeapType::Abstract { shared, ty } => wasm_encoder::HeapType::Abstract {
+            shared,
+            ty: convert_abstract_heap_type(ty),
+        },
+        wasmparser::HeapType::Concrete(idx) | wasmparser::HeapType::Exact(idx) => {
+            wasm_encoder::HeapType::Concrete(idx.as_module_index().unwrap_or(0))
+        }
+    };
+    RefType {
+        nullable: rt.is_nullable(),
+        heap_type,
+    }
+}
+
+/// Remap a ref type's concrete heap-type index through the merge index maps;
+/// abstract heap types carry no index. Mirrors the `RefNull` heap-type
+/// handling in [`ParsedConstExpr::reindex`].
+fn remap_ref_type(rt: RefType, maps: &IndexMaps) -> RefType {
+    let heap_type = match rt.heap_type {
+        wasm_encoder::HeapType::Concrete(idx) => {
+            wasm_encoder::HeapType::Concrete(maps.remap_type(idx))
+        }
+        other => other,
+    };
+    RefType { heap_type, ..rt }
+}
+
 pub fn reindex_element_segment(
     segment: &ParsedElementSegment,
     maps: &IndexMaps,
@@ -320,7 +358,7 @@ pub fn reindex_element_segment(
 
     ReindexedElementSegment {
         mode,
-        element_type: segment.element_type,
+        element_type: remap_ref_type(segment.element_type, maps),
         items,
     }
 }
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `segments.rs`

**VERDICT: NO PROVABLE FINDING.** Faithful, lossless replacement of the `is_func_ref() ? FUNCREF : EXTERNREF` bucketing; all six hazards refuted with runtime PoCs.

1. **Funcref regression** — `Functions` arm unchanged (`RefType::FUNCREF`); `Expressions`-funcref via `convert_ref_type` yields exactly `RefType::FUNCREF` (definitionally `Abstract{shared:false,ty:Func}`+nullable). Byte-identity confirmed.
2. **`as_module_index().unwrap_or(0)` silent-zero** — `None` only for RecGroup/Id kinds, which appear inside `(rec …)` definitions, never in a module-level element heap-type ref; identical to the established `RefNull` convention at segments.rs:553. Unreachable.
3. **`remap_type` missing-entry** — `*types.get(&idx).unwrap_or(&idx)`, identity default, same as `remap_func`/`global`/`table`; unmapped 0→0, mapped 0→3 (PoC).
4. **Abstract heap-type fidelity** — `convert_abstract_heap_type` is exhaustive over all 14 wasmparser abstract types (no `_` arm, no panic); GC PoC validates.
5. **Nullability** — `nullable: rt.is_nullable()` + `..rt` preserved; `(ref 0)` non-nullable round-trips and validates.
6. **Shared** — passed through for Abstract; shared concrete refs not constructible for core element segments.

**Regression-direction proof:** a PoC reconstructing the pre-fix behavior (concrete `ref.func` item under an `EXTERNREF` element type) confirms the validator *rejects* it — the H-3 bug the fix removes was real.

Suites: segments lib 20/20 (incl. the 2 new tests), wit_bindgen_runtime 73/73, multi_memory 5/5.

PR #259, opened 2026-06-13
fix(reftype): preserve concrete/GC reference types through fusion (closes #258, LS-A-22)

### LINKED ISSUE #258: parser/merger: convert_ref_type collapses concrete/GC reference types to funcref (tables, func signatures, globals) — silent type weakening
Sibling of #256/PR #257 (which fixed the element-segment case in `segments.rs`). The **value-type / table / signature** path has the same `is_func_ref()-else` collapse, in `parser.rs::convert_ref_type` — and it is worse-shaped (the `else` defaults to **funcref**, not even externref).

## The defect
`parser.rs::convert_ref_type` (lines 3776-3786):
```rust
if rt.is_func_ref() { Ref(FUNCREF) }
else if rt.is_extern_ref() { Ref(EXTERNREF) }
else { /* "default to funcref" */ Ref(FUNCREF) }   // concrete (ref $t), GC abstract, etc.
```
Any concrete typed reference `(ref null $t)` / `(ref $t)` or GC abstract heap type (struct/array/eq/any/i31/…) is flattened to **funcref**, losing the heap type and the concrete type index. This feeds:
- table element types — `parser.rs:1168, 1207` → `CoreModule.tables[].element_type`
- value types (func-signature params/results, globals) — `parser.rs:3771` (`WasmValType::Ref`)

## Reachability — CONFIRMED (runtime PoC)
Trace: `merger.rs:1513` pushes `convert_table_type(module.tables[i])`, and `convert_table_type` (merger.rs:2709) reads the already-collapsed `element_type`. Func types flow analogously via `merger.rs:1391` from `module.types`. **Neither table emission nor func-type emission remaps concrete heap-type indices.**

PoC (ran): fusing a component whose core module has `(table 1 (ref null $ft))` produces an output table with element type **`funcref`** (concrete typed-ref lost). The output still *validates* (funcref is a valid table type) — so the failure mode is **silent type weakening / type confusion (H-3)**, not an invalid module — but the fused module's table type differs from the source. A module that *uses* the typed-ref table with concrete-type operations (e.g. `call_indirect (type $ft)` against a typed-ref table, or `table.get` expecting `(ref $ft)`) can produce a type mismatch or semantic divergence.

## Why it's larger than #257 (filing, not fixing inline)
The complete fix spans **two Tier-5 files** and multiple sites:
1. `parser.rs::convert_ref_type` → preserve the full ref type faithfully (nullability + abstract heap type + concrete module index), mirroring the `segments.rs::convert_ref_type` landed in #257 (or the rewriter's `convert_ref_type(rt, maps)` which already handles this).
2. `merger.rs` → remap concrete heap-type indices when emitting **tables** (`convert_table_type`), **func-type signatures** (the `merged.types` build), and **globals** (`convert_global_type`) — analogous to the `remap_ref_type` added to `segments.rs` reindexing in #257. Without (2), even a faithful parse leaves stale type indices after the merge renumbers types.

There is **no clean bounded increment** (unlike #253's fail-loud guard): making `convert_ref_type` fail-loud would break modules that today carry typed-ref tables/signatures harmlessly (weakened-but-valid). The fix must be the full faithful-conversion + remap, done deliberately with a runtime matrix (typed-ref table used by `call_indirect (type $t)`; a func signature with a `(ref $t)` param; a GC global) proving the output preserves the type AND validates AND executes.

Recommend a dedicated increment (parser+merger, Mythos delta-pass on both, new LS in the LS-A ref-type-fidelity series alongside LS-A-21).

avrabe:
Closes #258 — the value-type/table/global sibling of the element-segment fix (#257). Found by the bug-hunt pipeline; fixes a silent type-weakening defect (H-3).

## Bug
`parser.rs::convert_ref_type` collapsed every reference type via `is_func_ref() ? funcref : is_extern_ref() ? externref : funcref`, dropping concrete typed refs `(ref null $t)` and GC abstract heap types. This fed table element types, func-type signatures, and global content types; the merger's table/type/global emission did no concrete-index remap. Runtime-PoC-confirmed: `(table 1 (ref null $ft))` fused to an output table of element type **funcref** — the concrete type lost. Often still validates (silent weakening), but a typed-ref table used by `call_indirect (type $ft)` mismatches.

## Fix (two parts)
- **parser.rs::convert_ref_type** — faithful conversion (nullability + abstract heap type via `convert_abstract_heap_type` + concrete module index), mirroring the #257 segments fix.
- **merger.rs** — new `remap_concrete_ref_type`/`remap_concrete_val_type` remap concrete heap-type indices when emitting **tables** (`convert_table_type`), **func-type signatures** (two-pass build so same-module forward refs resolve), and **globals** (`convert_global_type`), via `type_index_map`. Disjoint from the existing const-expr-operand remap → no double-remap.

## Verification
- `tests/reftype_fidelity.rs` (6 cases): typed-ref table preserved; typed-ref table used by `call_indirect` validates; **two-component fusion where each table's concrete index resolves to its correct distinct signature** (proves the remap lands on the right type, not merely a valid one); GC abstract global preserved; concrete-ref func-signature param preserved; plain-funcref regression control.
- Workspace **555/0** (via exit code); wit_bindgen_runtime 73/73; clippy/fmt clean.
- **Mythos clean-room pass over BOTH parser.rs + merger.rs deltas: NO PROVABLE FINDING** — all five risks falsified incl. double-remap (disjoint output fields, each remapped once), wrong-offset-but-valid (independent 2-component f64/f32 PoC resolves to the semantically correct sig), import-side path (live + correct), and silent-zero (unreachable — parser never runs the validator's rec-group canonicalization). Full report next comment.

## Tier-5
`parser.rs` + `merger.rs` → Mythos pass done, `mythos-pass-done` applied.

### DIFF
--- a/meld-core/src/merger.rs
+++ b/meld-core/src/merger.rs
@@ -1381,16 +1381,32 @@ impl Merger {
         shared_memory_plan: Option<&SharedMemoryPlan>,
         unresolved_assignments: &UnresolvedImportAssignments,
     ) -> Result<()> {
-        // Merge types
+        // Merge types.
+        //
+        // Two passes: first record every type's old->merged index mapping, then
+        // build the merged types. The split is required because a func type's
+        // param/result may be a concrete typed-ref `(ref $t)` whose index `t`
+        // can forward-reference another type in this same module; remapping it
+        // needs the *complete* mapping for the module to already be in place.
         let type_offset = merged.types.len() as u32;
-        for (old_idx, ty) in module.types.iter().enumerate() {
+        for (old_idx, _ty) in module.types.iter().enumerate() {
             merged.type_index_map.insert(
                 (comp_idx, mod_idx, old_idx as u32),
                 type_offset + old_idx as u32,
             );
+        }
+        for ty in module.types.iter() {
             merged.types.push(MergedFuncType {
-                params: ty.params.clone(),
-                results: ty.results.clone(),
+                params: ty
+                    .params
+                    .iter()
+                    .map(|&p| remap_concrete_val_type(p, comp_idx, mod_idx, merged))
+                    .collect(),
+                results: ty
+                    .results
+                    .iter()
+                    .map(|&r| remap_concrete_val_type(r, comp_idx, mod_idx, merged))
+                    .collect(),
             });
         }
 
@@ -1510,7 +1526,9 @@ impl Merger {
             merged
                 .table_index_map
                 .insert((comp_idx, mod_idx, old_table_idx), new_idx);
-            merged.tables.push(convert_table_type(table));
+            merged
+                .tables
+                .push(convert_table_type(table, comp_idx, mod_idx, merged));
         }
 
         // Merge globals (defined globals only; imported globals handled below)
@@ -1528,10 +1546,8 @@ impl Merger {
                 merged,
                 &global.content_type,
             );
-            merged.globals.push(MergedGlobal {
-                ty: convert_global_type(global),
-                init_expr,
-            });
+            let ty = convert_global_type(global, comp_idx, mod_idx, merged);
+            merged.globals.push(MergedGlobal { ty, init_expr });
         }
 
         // Resolve imported global indices via intra-component module_resolutions.
@@ -2363,7 +2379,12 @@ impl Merger {
                     merged.imports.push(MergedImport {
                         module,
                         name,
-                        entity_type: EntityType::Table(convert_table_type(t)),
+                        entity_type: EntityType::Table(convert_table_type(
+                            t,
+                            unresolved.component_idx,
+                            unresolved.module_idx,
+                            merged,
+                        )),
                         component_idx: Some(unresolved.component_idx),
                     });
                 }
@@ -2418,7 +2439,12 @@ impl Merger {
                     merged.imports.push(MergedImport {
                         module,
                         name,
-                        entity_type: EntityType::Global(convert_global_type(g)),
+                        entity_type: EntityType::Global(convert_global_type(
+                            g,
+                            unresolved.component_idx,
+                            unresolved.module_idx,
+                            merged,
+                        )),
                         component_idx: Some(unresolved.component_idx),
                     });
                 }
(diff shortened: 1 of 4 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 files `parser.rs` + `merger.rs`

**VERDICT: NO PROVABLE FINDING.** All five primary risks falsified with running PoCs.

1. **Double-remap** — the two paths write disjoint output fields: `remap_concrete_val_type` → `MergedGlobal.ty`; the `convert_const_expr` RefNull arm (merger.rs:3039) → `MergedGlobal.init_expr`. PoC: a 2-component fusion of `(global (ref null $t1) (ref.null $t1))` with non-identity offsets — declared-type idx == init `ref.null` idx for both globals, in-range, validates. Each path remapped exactly once.
2. **Missing-map-entry** — `.get(...).unwrap_or(old_idx)` → identity, no panic/silent-0; map populated for every type of every module (merger.rs:1393), so the fallback is unreachable in normal fusion. Import-side sites key correctly on `(unresolved.component_idx, module_idx)`.
3. **Wrong-offset-but-valid** — independent adversarial 2-component PoC, A=`(f64)->f64` / B=`(f32,f32)->f32`: the two table concrete indices resolve to each's **semantically correct** sig (a wrong module offset would collapse B onto A's slot; the resolution-to-signature assertion catches it). Shipped case3 does the same.
4. **parser faithfulness + funcref regression** — funcref → exactly `RefType::FUNCREF` (byte-identity); `as_module_index().unwrap_or(0)` silent-zero is **unreachable** (raw `FromReader` decode always yields `Module(_)`; rec-group-relative form is only produced by the validator's `canonicalize_rec_group`, which the parser never runs).
5. **Import-side table/global remap** — constructed an imported typed-ref table + global; both survive to the fused output carrying their concrete index, output validates. Path is live and correct.

Suites: reftype_fidelity 6/6, multi_memory 5/5, wit_bindgen_runtime 73/73, rewriter convert_ref_type 2/2 — green by exit code.

PR #260, opened 2026-06-13
fix(rewriter): relocate element/data segment indices through fusion (LS-M-8)

avrabe:
Found by a Mythos discovery pass on `rewriter.rs`. Fixes a confirmed, end-to-end-reproduced **silent semantic-divergence** bug (H-1/H-3).

## Bug
The rewriter relocated function/type/global/table/memory operands but passed **element/data segment indices** through verbatim in `memory.init` / `data.drop` / `table.init` / `elem.drop` — and `IndexMaps` had no segment maps. The merger concatenates every module's passive segments into one shared section, so a non-first module's segment 0 becomes ordinal ≥1. **Reproduced end-to-end:** a fused two-module component validated and ran, but `probe_b` read `0xAAAAAAAA` (module A's data) instead of `0xBBBBBBBB`. Out-of-range stale indices instead produce an invalid module.

## Fix
`IndexMaps` gains `data_segments`/`elements` maps (identity-default `remap_*`, mirroring `remap_func`). `merge_core_module` captures the segment base (count of prior modules' segments, before appending this module's — same timing as the proven `func_offset` capture, the sole deterministic append site) and populates `local → base+local`; the base is recorded in `MergedModule.segment_bases` so the four post-merge re-rewrite callers use the correct base. The rewriter remaps `elem_index` (TableInit/ElemDrop) and `data_index` (DataDrop, `rewrite_memory_init`).

## Verification
- `tests/segidx_remap.rs`: two-module **data** + **element** distinct-segment round-trips (each probe reads its OWN segment), a **three-module base-offset>1** case, and a single-module identity case. Multi-module cases **fail on pre-fix code** (non-vacuity confirmed by stashing the fix).
- Workspace **552/0** (via exit code); wit_bindgen_runtime 73/73; multi_memory 5/5; clippy/fmt clean.
- **Mythos clean-room pass over rewriter.rs + merger.rs + segments.rs: NO PROVABLE FINDING** — all 5 risks falsified, the **double-remap-via-re-rewrite** risk refuted with two independent proofs (re-rewrite callers re-parse *original* bytes with a fresh IndexMaps → single application; the map is keyed by original-local index → an already-shifted operand misses the keyset → identity). Full report next comment.

## Tier-5
`rewriter.rs` + `merger.rs` (+ `segments.rs` count helpers) → Mythos pass done, `mythos-pass-done` applied.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -6009,6 +6009,7 @@ mod tests {
             handle_tables: std::collections::HashMap::new(),
             task_return_shims: std::collections::HashMap::new(),
             async_result_globals: std::collections::HashMap::new(),
+            segment_bases: std::collections::HashMap::new(),
         }
     }
 
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -865,6 +865,15 @@ impl Fuser {
                 .unwrap_or(false);
             let memory_initial_pages = module_memory.as_ref().map(|mem| mem.initial);
 
+            // Post-merge re-rewrite: recover this module's segment bases from
+            // the merge-time record (`merged.*.len()` is now the total, not
+            // the base).
+            let (data_segment_base, elem_segment_base) = merged
+                .segment_bases
+                .get(&(comp_idx, mod_idx))
+                .copied()
+                .unwrap_or((0, 0));
+
             let index_maps = merger::build_index_maps_for_module(
                 comp_idx,
                 mod_idx,
@@ -875,6 +884,8 @@ impl Fuser {
                 memory_base_offset,
                 memory64,
                 memory_initial_pages,
+                data_segment_base,
+                elem_segment_base,
             );
 
             let import_func_count = module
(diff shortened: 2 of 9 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 (rewriter.rs, merger.rs, segments.rs)

**VERDICT: NO PROVABLE FINDING.** All five risks falsified; meld-core 552/0 (segidx_remap 4, multi_memory 5, wit_bindgen_runtime 73, p3_async/bridge green).

**Risk 2 — double-remap via re-rewrite (hunted hardest) — REFUTED, two independent proofs:**
- *Structural:* every re-rewrite caller (lib.rs:854, merger.rs:1220, p3_async, p3_bridge) reads the **original** parsed module and `extract_function_body` re-parses `module.bytes[start..end]` with a **fresh** IndexMaps, replacing `mf.body` wholesale — the remapped body is never fed back through the rewriter, so single application is guaranteed.
- *Keyset:* the map is keyed by **original-local** index `{0..count}`; a (refuted) naive double-feed of an already-shifted `base+local` operand misses the keyset → identity, no re-shift.
- PoC (base=3): primary 0→3; re-rewrite from original bytes stays 3 (not 6); bodies byte-identical; double-feed stays 3.

**Risk 1 — base-offset timing:** captured at merger.rs:1896-1897 BEFORE the sole append site (2175/2182), mirrors `func_offset`; merge order deterministic (Kahn min-heap, no HashMap iteration). Covered by the in-tree 3-module base>1 test.

**Risk 3 — `unwrap_or((0,0))`:** dead code — `merge_component` calls `merge_core_module` (which inserts) for every module; re-rewrite keys are always a subset, always present.

**Risk 4 — count correctness:** `count_data_segments`/`count_element_segments` use the section reader's declared LEB count == items the iterator yields == entries `parse_*_segments` pushes (one per item, no filtering); passive/active counted identically; a reader error makes both `count_*` and `parse_*` fail, never disagree silently.

**Risk 5 — identity default:** the `0..count` loop maps every in-range local index; out-of-range is invalid wasm rejected upstream.

PR #262, opened 2026-06-13
fix(adapter): detect resources in option/result/variant/fixed-list aggregates (LS-A-23, UCA-A-16)

avrabe:
Found by a Mythos discovery pass on `parser.rs`. Closes the still-open arms of the documented hazard UCA-A-16 / H-11.5.

## Bug
`element_inner_resources` (populating `CopyLayout::Elements.inner_resources` for the FACT adapter's per-element `borrow<R>`→rep translation) detected resources in Record/Tuple/Type but its catch-all `_ => {}` **swallowed Option/Result/Variant/FixedSizeList** — while the sibling `element_inner_pointers` recursed into all of them. So a `borrow<R>` inside `list<option<borrow<R>>>` (etc.) was undetected and the adapter bulk-copied the raw caller handle index into callee memory instead of translating it → wrong-resource / use-after-free / type confusion. Same "parallel walkers drifted" root cause as LS-P-12 / LS-R-15.

## Fix (mirrors the working InnerPointer guard mechanism)
- `InnerResource` gains `guards: Vec<DiscriminantGuard>` + `unconditional` ctor.
- `element_inner_resources` gains an `outer_guards` param and Option/Result/Variant/FixedSizeList arms identical to `element_inner_pointers` (same disc_size, `align_up` payload offset, per-case guard values), gated by a new `type_contains_resources`.
- The adapter wraps each borrow→rep conversion in `emit_inner_resource_guard_chain` so the handle is translated **only when the live discriminant selects its case** (empty chain = unconditional, preserving Record/Tuple). `own<R>` stays skipped.

## Verification
- 7 parser tests: resource in option/result/variant/fixed-list detected with correct guard values; Record stays unconditional; pointer path unchanged. Non-vacuity proven (routing the 4 kinds back to the swallow makes the detection tests fail). Workspace lib 351 + adapter_safety 13 + wit_bindgen_runtime 73 (incl. resource_borrow / resource_borrow_in_record / resource_import_and_export) green via exit code; clippy/fmt clean.
- **Mythos clean-room pass: NO PROVABLE FINDING** — the resource guard emission is proven **byte-identical** to the already-runtime-validated InnerPointer guard path (disc load width/offset/compare/AND-chain), so it fires exactly on the live case; payload offsets, nested threading (`result<borrow,borrow>` → mutually-exclusive guards), and own-not-translated all verified. Full report next comment. Carried caveat: no two-component runtime fixture (behaviour inferred from byte-identity to the runtime-validated pointer path).

## Tier-5
`parser.rs` + `resolver.rs` + `adapter/{fact,mod}.rs` → Mythos pass done, `mythos-pass-done` applied.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -1109,7 +1109,11 @@ impl FactStyleGenerator {
                     if let Some(rep_func) = rep_func {
                         options
                             .inner_resource_fixups
-                            .push((inner.byte_offset, rep_func));
+                            .push(super::InnerResourceFixup {
+                                byte_offset: inner.byte_offset,
+                                rep_func,
+                                guards: inner.guards.clone(),
+                            });
                     } else {
                         log::warn!(
                             "inner-list borrow at offset {} (resource_type_id={}): \
@@ -1634,7 +1638,23 @@ impl FactStyleGenerator {
                 func.instruction(&Instruction::LocalGet(1)); // len
                 func.instruction(&Instruction::I32GeU);
                 func.instruction(&Instruction::BrIf(1));
-                for &(byte_offset, rep_func) in &options.inner_resource_fixups {
+                for fixup in &options.inner_resource_fixups {
+                    let byte_offset = fixup.byte_offset;
+                    let rep_func = fixup.rep_func;
+                    // UCA-A-16: AND-evaluate this fixup's per-element
+                    // discriminant guards before the borrow→rep conversion.
+                    // Empty guards → unconditional (Record/Tuple path). Mirrors
+                    // `emit_inner_pointer_fixup`: load each guard's disc byte
+                    // from the per-element base at the right width, compare,
+                    // AND-chain, and wrap the conversion in an `If`.
+                    let guarded = Self::emit_inner_resource_guard_chain(
+                        &mut func,
+                        &fixup.guards,
+                        dest_ptr_local,
+                        loop_idx,
+                        element_size,
+                        options.callee_memory,
+                    );
                     // addr = dest_ptr + loop_idx * element_size + byte_offset
                     // i32.store needs [addr, value] on stack.
                     // Emit: addr, addr, i32.load → handle, call rep → rep, i32.store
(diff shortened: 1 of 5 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 (parser.rs, resolver.rs, adapter/fact.rs, adapter/mod.rs)

**VERDICT: NO PROVABLE FINDING.** Primary risk (guard fires on wrong/inactive case → silent handle corruption) disproven at the wasm-byte level.

- **Guard value/width/offset (hunted hardest)** — `emit_inner_resource_guard_chain` diffed line-by-line against the pointer path's guard block: same `dst_base + loop_idx*element_size` base, same `MemArg{offset: guard.position, memory: callee}`, same width match (1→`i32.load8_u`, 2→`i32.load16_u`, else→`i32.load`), same `i32.const value`/`i32.eq`/AND-chain/`if`. **Byte-identical code sections** proven via round-trip for option-Some `{0,1,1}`, variant-case-1, and nested result-Ok+option-Some `[{0,0,1},{1,1,1}]`. Both resource fixup loops use the same base/element_size/memory as the handle load — no offset divergence.
- **Payload offset** — `option<borrow<R>>`: disc_size(2)=1, borrow align=4 → payload at `align_up(1,4)=4`, guard.position=0, byte_size=1; identical to the pointer arm's formula.
- **Nested threading** — `result<borrow,borrow>` → two InnerResources guards value=0/1 (mutually exclusive); `result<option<borrow>>` → 2-guard AND-chain (byte-identity proven).
- **own vs borrow** — `own<R>` detected in layout but the adapter `continue`s on `is_owned` before pushing a fixup; never translated; unchanged.
- **Regression** — pointer guard emission byte-identical to itself; Record/Tuple resources emit empty guards → no `if` wrapper; all resource runtime suites green.

Suites (exit 0): lib 351, adapter_safety 13, wit_bindgen_runtime 73.
Caveat: no two-component wasmtime fixture through `list<option<borrow<R>>>` — Risk-1 proof is at the wasm-emission level (byte-identity to the runtime-validated pointer path) rather than executed end-to-end.

PR #263, opened 2026-06-13
fix(p3-async): relocate ref.func in expression-form element segments (LS-M-9)

avrabe:
Found by a Mythos discovery pass on `p3_async.rs`. Fixes a confirmed end-to-end index-relocation bug (H-1/H-3, SC-3 violation) — the third site of the missing-reference-site class (with #260's segment indices).

## Bug
`shift_function_indices` prepends `k` `pulseengine:async` imports and shifts defined-function refs up by `k`. The `Functions`-form element arm bumped correctly, but the `Expressions`-form arm was a comment-only NO-OP (no bump, no error). So a `(ref.func $b)` inside an expression-form element segment stayed stale → the table slot pointed at a prepended intrinsic import (e.g. `stream_new`) instead of `$b` → wrong `call_indirect` / trap. Reproduced end-to-end: `stream<u8>` (k=5) component with `(elem (i32.const 0) funcref (ref.func $b))` → `$b` bumped 2→7 but elem ref stayed 2; a k=0 control was correct.

## Fix
`ReindexedElementItems::Expressions` now carries structured `Vec<ParsedConstExpr>` (encoding deferred to emit) instead of opaque `Vec<ConstExpr>`, so the index stays accessible. The shift arm bumps `ParsedConstExpr::RefFunc(idx)` via the same `bump` closure as Functions; GlobalGet/RefNull/scalars untouched (already merger-finalized; P3 prepends only function imports).

## Verification
- `p3_shift_relocates_expr_form_elem_ref_func` (k=5: elem ref == `$b`'s shifted index, not an intrinsic slot) + k=0 control; fails on pre-fix code (non-vacuity confirmed). segments lib 20/20 (incl. #257/#260 regression); workspace exit 0; clippy/fmt clean.
- **Mythos clean-room pass: NO PROVABLE FINDING** — the deferred encode is **byte-identical** to the prior eager path across all ParsedConstExpr kinds × all segment modes (RefFunc/RefNull-concrete+abstract/GlobalGet, Passive/Declared/Active); only RefFunc is bumped (GlobalGet/RefNull verified untouched); the `&mut ElementSection` sink preserves one-in/one-out ordering; RefNull+typed-ref element validates. Full report next comment.

## Tier-5
`segments.rs` + `p3_async.rs` → Mythos pass done, `mythos-pass-done` applied. (lib.rs is the encode call-site only.)

### DIFF
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -1601,7 +1601,7 @@ impl Fuser {
         if !merged.elements.is_empty() {
             let mut elements = wasm_encoder::ElementSection::new();
             for segment in &merged.elements {
-                elements.segment(segments::encode_element_segment(segment));
+                segments::encode_element_segment(&mut elements, segment);
             }
             module.section(&elements);
         }
--- a/meld-core/src/p3_async.rs
+++ b/meld-core/src/p3_async.rs
@@ -1287,14 +1287,26 @@ fn shift_function_indices(
                     *f = bump(*f);
                 }
             }
-            crate::segments::ReindexedElementItems::Expressions(_) => {
-                // Expressions hold encoded ConstExpr bytes (already
-                // baked). For ref.func the function index is encoded
-                // inside; we cannot easily rewrite without re-parsing.
-                // P3 async lowering does not currently support this
-                // case and components that hit it will be flagged.
-                // The vast majority of element segments in fused
-                // output use Functions(_) form.
+            crate::segments::ReindexedElementItems::Expressions(exprs) => {
+                // Expressions now carry the STRUCTURED, merger-reindexed
+                // `ParsedConstExpr` form, so the `ref.func` index is
+                // accessible and can be shifted just like the Functions
+                // arm. Only `RefFunc` carries a defined-function index
+                // that this P3 shift moves; the other variants are left
+                // untouched (see below).
+                for e in exprs.iter_mut() {
+                    if let crate::segments::ParsedConstExpr::RefFunc(idx) = e {
+                        *idx = bump(*idx);
+                    }
+                    // RefNull/GlobalGet/I32Const/I64Const/F32Const/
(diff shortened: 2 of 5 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 (segments.rs, p3_async.rs)

**VERDICT: NO PROVABLE FINDING.** All four risks falsified.

- **Byte-identity of deferred encode (primary)** — the representation change moves the same `to_const_expr()` (unchanged 1:1 mapping) + `Elements::Expressions(element_type, …)` later in time. PoC: full `ElementSection` bytes through the new path vs an origin/main-replicating reference, for RefFunc / RefNull(Concrete+abstract Func) / GlobalGet / mixed multi-item × Passive/Declared/Active-offset — **all byte-equal** (+ a falsifiability guard). `element_type` (the #257 typed-ref fix) and mode/offset preserved.
- **Only-RefFunc bump** — the arm matches only `RefFunc(idx)` with the same single `bump` closure as Functions; GlobalGet(2)/RefNull(Concrete(2)) proven NOT moved (merger already finalized them; P3 prepends only functions); threshold consistent with the Functions arm (no split-shift).
- **`&mut ElementSection` sink** — one `section.segment(..)` per call, mode/offset/element_type preserved, sole caller (lib.rs:1603) loops in order; no dropped/reordered/duplicated segment.
- **RefNull + #257 typed-ref** — `(elem (ref null $t) (ref.func 0) (ref.null $t))` round-trips: ref.func bumpable, ref.null heap type preserved, validates under all features.

Note (not a finding): the active-segment offset const-expr is still encoded eagerly, so P3 can't bump a ref.func in an offset — harmless (table offsets are i32/global, never ref.func), matches pre-fix.

Suites exit 0: segments lib 20, p3_async_lowering, p3_async_abi, wit_bindgen_runtime 73, multi_memory.

Issue #264, opened 2026-06-13
p3_stream: canonicalize stream element types to a cross-component structural descriptor (complete fix for LS-R-16)


avrabe:
Follow-up to PR for LS-R-16 (the conservative over-match guard just landed). That guard makes `pair_streams` decline to pair non-primitive stream element types (recorded as the component-local `stream<Type(N)>`), eliminating the safety-relevant over-match — but at the cost of NOT pairing aggregate-element streams at all (they stay host-routed; the benign under-match, Finding A).

## Complete fix
Canonicalize a stream's element type to a **cross-component-stable structural descriptor** so aggregate-element streams (`stream<record{..}>`, `stream<list<u8>>`, `stream<own<R>>`, etc.) pair correctly when — and only when — the producer's and consumer's element types are structurally identical:
- In `p3_stream` detection (which has `&[ParsedComponent]` and the per-endpoint component index), resolve each endpoint's `Type(N)` through THAT component's type table to a structural key, then compare structural keys across components.
- Apply the same canonicalization in `validate_stream_pair_graph` (its multiset comparison currently compares the same local-index strings).

## Subtlety — resource element identity
A `stream<own<R>>` / `stream<borrow<R>>` element's `R` is itself a component-local resource type; structural equality across components for resource elements overlaps the **#256 resource_graph definer-attribution** problem (cross-component resource identity). The complete fix should either reuse a canonical resource identity once #256's attribution is sound, or conservatively keep resource-element streams host-routed until then.

## Test
A two-component fusion where producer and consumer share a genuine aggregate element type T (at different local indices) → exactly one pair (the LS-R-16 Finding-A direction, now positively enabled), AND two different aggregate types at the same local index → no pair (over-match stays prevented). Plus an end-to-end runtime differential if a real `stream<record>` cross-memory fixture is constructible.

Tier-5 (p3_stream.rs + possibly resolver) → Mythos delta-pass. The conservative guard (LS-R-16) is the safe interim; this issue is the correctness completion.

PR #265, opened 2026-06-13
fix(p3-stream): don't pair cross-component streams on local type index (LS-R-16)

avrabe:
Found by a Mythos discovery pass on `p3_stream.rs`. Conservative fix for the safety-relevant over-match (Finding B); the complete structural-canonicalization is tracked in #264.

## Bug
`pair_streams` matched producer/consumer stream endpoints by element-type descriptor. Primitives serialise as the stable `Primitive(..)`, but non-primitive elements are `stream<Type(N)>` with N a **component-local** type index — compared literally across components. Two different element types colliding on the same local index N would string-match → false pair → drives `p3_bridge::emit_stream_bridge` to wire incompatible streams (H-3.1 type confusion). Confirmed at API level; end-to-end plausible-but-unproven.

## Fix (conservative increment)
Decline to pair when the element descriptor is a local-index `Type(N)` form (`is_cross_component_pairable`). Eliminates the over-match (incompatible aggregate streams stay host-routed) with **no regression** to correct primitive-stream pairing. The benign under-match (aggregate streams not paired) remains; the complete fix (canonicalize `Type(N)` to a structural cross-component descriptor, incl. resource-element identity overlapping #256) is **#264**.

## Verification
- `local_index_element_descriptors_do_not_pair`: two colliding `Type(0)` descriptors → 0 pairs; primitive control → 1 pair.
- p3_stream 28/28; workspace exit 0; clippy/fmt clean. The guard is safe-by-construction (only adds a `continue` — can never create a pair).
- **Mythos clean-room pass: pending** (next comment).

## Tier-5
`p3_stream.rs` → Mythos pass to follow + `mythos-pass-done`.

### DIFF
--- a/meld-core/src/p3_stream.rs
+++ b/meld-core/src/p3_stream.rs
@@ -69,6 +69,24 @@ impl StreamElement {
         let inner = desc.strip_prefix("stream<")?.strip_suffix('>')?;
         Some(StreamElement::Typed(inner.trim().to_string()))
     }
+
+    /// Whether this element type is safe to PAIR across components.
+    ///
+    /// A bare `Type(N)` descriptor is a component-LOCAL type index —
+    /// component A's `Type(5)` and component B's `Type(5)` are unrelated —
+    /// so it is not a valid cross-component key (LS-R-16). Primitive
+    /// descriptors (`Primitive(..)` / scalar names) and the untyped stream
+    /// are globally stable and pairable; a local-index descriptor is not
+    /// (pairing on it risks an over-match that drives the bridge emitter to
+    /// wire incompatible streams). Canonicalising `Type(N)` to a structural
+    /// descriptor — which would make aggregate-element streams pairable — is
+    /// tracked follow-up; until then such streams stay host-routed.
+    fn is_cross_component_pairable(&self) -> bool {
+        match self {
+            StreamElement::Untyped => true,
+            StreamElement::Typed(s) => !s.contains("Type("),
+        }
+    }
 }
 
 /// A component's role on a particular stream element type.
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `p3_stream.rs`

**VERDICT: NO PROVABLE FINDING.** The guard correctly and completely closes the over-match.

- **Over-match eliminated (incl. the "non-primitive without `Type(`" case — proven impossible)** — the parser builds the descriptor as `format!("stream<{ty:?}>")` with `ty: wasmparser::ComponentValType`, which has **exactly two** derived-Debug variants: `Primitive(..)` and `Type(N)`. An aggregate stream element (record / `list<u8>` / `own<R>`) is binary-encoded as a type-index reference → can only surface as `Type(N)` → always contains `Type(` → not paired. No element shape the parser can record is non-primitive yet lacks the substring. PoC: colliding `Typed("Type(0)")` → 0 pairs.
- **No false-positive skip** — all 14 `PrimitiveValType` Debug names (`Bool/S8/…/Char/String/ErrorContext`) verified to NOT contain `Type(` (`ErrorContext` has "Context", not "Type("); primitives still pair. Gating tests green.
- **Predicate soundness** — over the finite descriptor space `Primitive(..) | Type(N)`, `contains("Type(")` is true **iff** local-index form (`Type(4294967295)` still matches); checking only `p_elem` is sufficient since it runs after `p_elem == c_elem`.
- **future/untyped** — futures never enter `pair_streams` (`from_descriptor` returns None); `Untyped` stays pairable (no element type, no risk).
- **Conservatism** — the guard only `continue`s after a successful match, so pairing strictly shrinks; it can never create a pair.

Suites (exit 0): p3_stream lib 28, p3_async_lowering 13, wit_bindgen_runtime 73.

PR #266, opened 2026-06-13
fix(merger): type-compatibility check for global/table import dedup (LS-M-10)

avrabe:
Found by a Mythos discovery pass on `merger.rs`. Closes a UCA-M-9 / SC-12 gap (those were scoped to function imports only).

## Bug
The import dedup recorded a `type_mismatch` and allocated separate slots for same-named **function** imports of different signatures — but the **table** and **global** arms keyed on name only and unconditionally reused the first slot. Two components importing the same-named global/table at different types (e.g. `env.g : i32 immutable` vs `i64 mutable`; `env.t : funcref[1..]` vs `externref[5..]`) collapsed into one wrong-typed import → the other component's index-map pointed at it → invalid fused module / type confusion (H-1/H-10). Same "guard on one arm, missing on siblings" asymmetry as the rest of this sweep.

## Fix
`seen_table`/`seen_global` now store the type signature (TableSig = element+limits; GlobalSig = content+mutability); an Occupied key with a differing sig allocates a fresh slot + records `type_mismatch_entries`, and `add_unresolved_imports` emits a separate (`$comp`-suffixed in multi-memory) import — mirroring the function path exactly. Same-typed dups still dedup.

## Verification
- merger unit tests: global/table type-mismatch → 2 slots with distinct index-map positions; same-type → 1 slot (regression). Non-vacuity proven (reverting the arms → tests fail, collapsed to 1). import tests 25/25; workspace exit 0; clippy/fmt clean.
- **Mythos clean-room pass: NO PROVABLE FINDING** — the compute/emit position-invariant holds across multi-component cases (incl. A,B,A → comp2 dedups to comp0); each component's index-map resolves to a slot of its own type; **shared-mode duplicate import names with differing types are legal core wasm** (encoded + validated — the fix never emits an invalid module); sigs complete; regression intact. Full report next comment. Known minor limitation (matches the pre-existing function arm): a third importer matching the *second* distinct type gets a redundant slot — dedup-completeness gap, not corruption.

## Tier-5
`merger.rs` → Mythos pass done, `mythos-pass-done` applied.

### DIFF
--- a/meld-core/src/merger.rs
+++ b/meld-core/src/merger.rs
@@ -2224,6 +2224,10 @@ impl Merger {
         // Track base (module, field) names already emitted for function imports
         // so we can suffix duplicates in multi-memory mode.
         let mut emitted_base_func: HashSet<(String, String)> = HashSet::new();
+        // Same for tables/globals: type-mismatched same-named imports need a
+        // unique (module, field) in multi-memory mode.
+        let mut emitted_base_table: HashSet<(String, String)> = HashSet::new();
+        let mut emitted_base_global: HashSet<(String, String)> = HashSet::new();
 
         for unresolved in &graph.unresolved_imports {
             // Skip imports resolved by adapter sites (must match the
@@ -2388,7 +2392,16 @@ impl Merger {
                     }
                 }
                 ImportKind::Table(t) => {
-                    if !emitted_table.insert(dedup_key.clone()) {
+                    // Type-mismatched entries must emit a separate import even
+                    // when the dedup key was already seen (mirrors the function
+                    // arm). Same-typed duplicates still collapse to one slot.
+                    let is_type_mismatch = dedup_info.type_mismatch_entries.contains(&(
+                        unresolved.component_idx,
+                        unresolved.module_idx,
+                        unresolved.module_name.clone(),
+                        unresolved.field_name.clone(),
+                    ));
+                    if !is_type_mismatch && !emitted_table.insert(dedup_key.clone()) {
                         continue;
                     }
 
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `merger.rs`

**VERDICT: NO PROVABLE FINDING.** Sound on all five risks. Full suite green (lib 359/0; wit_bindgen_runtime 73, multi_memory, import-dedup all 0 failed).

- **Compute/emit position-invariant (hunted hardest)** — both functions iterate `unresolved_imports` in identical order, build the dedup key from the same `effective_import_key`+`comp_dim`, and look up `type_mismatch_entries` under the same raw tuple. Push-vs-reuse predicates are provably equivalent (compute: Vacant OR sig-differs; emit: is_type_mismatch OR first-occurrence). 3-component PoC `A,B,A` → comp2 deterministically dedups to comp0's slot in BOTH; emit_count == counts, dense positions. No desync.
- **Risk 3 (shared-mode duplicate name)** — encoded a SharedMemory module with two `(import "env" "g" ...)` of differing types and ran `Validator::validate_all` → **Ok**. Duplicate import (module,name) with differing types is legal core wasm (positional import space). The fix does NOT produce an invalid module; host-side resolvability is an unchanged downstream linker concern.
- **Type-map → right slot** — comp1's mismatch import gets a fresh position carrying comp1's own converted type; comp2(=comp0) maps to comp0's slot. `assert_ne!` on the mapped positions.
- **Sig completeness** — meld's TableType is exactly (element, initial, maximum) (no table64/shared field); GlobalType (content, mutable); init_expr correctly excluded. No false-dedup/spurious-split.
- **Regression** — same-typed dup still collapses (WASI shared imports unaffected); `$comp` suffix unique per component.

Noted (not a bug): a third importer matching the *second* distinct type gets a redundant slot (`seen` holds only the first sig) — dedup-completeness gap, compute/emit stay in lockstep, identical to the pre-existing function arm.

PR #267, opened 2026-06-14
fix(component_wrap): map func-import ordinal to mixed-kind import slot (LS-W-2)

avrabe:
Found by a Mythos discovery pass on `component_wrap.rs` — the final Tier-5 file in the sweep. Same asymmetry shape as the rest: a guard present in one site, missing in its sibling.

## Bug
`assemble_component` indexed the func-only `fused_info.func_imports` with `import_idx` but read `merged.imports.get(import_idx)` — and `merged.imports` is a **mixed-kind** vector (Function/Global/Table/Memory interleaved). When a non-func import precedes a func import, the func's source-component `comp_idx` is read from the wrong slot. E.g. `merged.imports = [Global(comp 0), Func(comp 1)]` → the func read `comp_idx = Some(0)` instead of `1`. The wrong identity feeds `find_task_return_for_import` (self-healing fallback) and `LocalResource{component_idx}` → `ht_export_suffix` → resource op mis-routed to the wrong component's handle table (H-8/H-2). The sibling `lib.rs:1119` already handles the same vector correctly.

## Fix
Precompute `func_import_positions` (positions of Function entries in `merged.imports`) and read `merged.imports[func_import_positions[import_idx]]`, mirroring lib.rs:1119. The output import section is emitted from `merged.imports` in order (lib.rs:1538) and re-parsed in order, so the i-th Function entry ↔ `func_imports[i]` 1:1 (verified by reading the emit/parse paths); `debug_assert_eq!` guards count-equality. All-func-imports common case is identity (no behaviour change).

## Verification
- `func_import_comp_idx_skips_non_func_imports` (mixed-kind → comp_idx Some(1); old `.get(0)` asserted to yield wrong Some(0) — non-vacuity) + `func_import_positions_identity_for_all_func_imports`. clippy 0; full suite exit 0; wit_bindgen_runtime 73 + golden_e2e + LS-CP-6 green.
- **Order-correspondence (the load-bearing correctness question) verified by reading the code**: `lib.rs:1538` emits the output import section iterating `merged.imports` in order (no filter/reorder); `parse_fused_module` reads it back in order → `fused_info.func_imports[i]` is the i-th Function in `merged.imports`. Count-equality holds the same way.
- Full adversarial Mythos clean-room pass: re-running (the first attempt hit a transient server rate-limit); result will be posted before merge.

## Tier-5
`component_wrap.rs` → Mythos pass to follow as comment + `mythos-pass-done` label before merge.

### DIFF
--- a/meld-core/src/component_wrap.rs
+++ b/meld-core/src/component_wrap.rs
@@ -938,14 +938,41 @@ fn assemble_component(
     // -----------------------------------------------------------------------
     let instance_map = build_instance_func_map(source);
 
+    // `fused_info.func_imports` is a FUNC-ONLY ordinal list, but
+    // `merged.imports` is a MIXED-KIND vector (Function/Global/Table/Memory
+    // interleaved in declaration order). Precompute the positions of the
+    // Function-kind entries in `merged.imports` so a func ordinal maps to the
+    // correct slot. Without this, a non-func import preceding a func import
+    // shifts every subsequent func's metadata (component_idx) to the wrong
+    // entry, mis-routing resource ops (H-8/H-2 class). This is the inverse of
+    // the position→func-ordinal mapping in `lib.rs` (`merged_func_idx`).
+    let func_import_positions: Vec<usize> = merged
+        .imports
+        .iter()
+        .enumerate()
+        .filter(|(_, i)| matches!(i.entity_type, EntityType::Function(_)))
+        .map(|(p, _)| p)
+        .collect();
+    // The count of Function entries in `merged.imports` must equal the
+    // func-import count; otherwise the ordinal→position mapping is unsound.
+    debug_assert_eq!(
+        func_import_positions.len(),
+        fused_info.func_imports.len(),
+        "func-import count mismatch: {} Function entries in merged.imports vs \
+         {} fused func_imports",
+        func_import_positions.len(),
+        fused_info.func_imports.len()
+    );
+
     let mut import_resolutions: Vec<ImportResolution> = Vec::new();
     for (import_idx, (module_name, field_name, _type_idx)) in
         fused_info.func_imports.iter().enumerate()
     {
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos clean-room delta-pass — Tier-5 file `component_wrap.rs`

**VERDICT: NO PROVABLE FINDING.** Both load-bearing properties hold, evidenced.

- **Order-correspondence (load-bearing)** — emit (`lib.rs:1536-1542`) iterates `merged.imports` in declaration order, no filter/group/reorder; parse (`component_wrap.rs:289-296`) walks the import section in order pushing `TypeRef::Func` entries. So `fused_info.func_imports[i]` IS the i-th `Function` entry in `merged.imports` = `func_import_positions[i]`. PoC: interleaved `[Global, Func(A), Memory, Func(B), Func(C), Global, Func(D)]` → parsed funcs `[A,B,C,D]`, positions `[1,3,4,6]`, each routes to its own distinct comp_idx; old `.get(0)` read the Global (non-vacuity).
- **count-equality / `convert_memory_to_import`** — it re-emits ALL imports verbatim in order and appends only `Memory` imports at the end; never adds/removes/reorders a Function. It also runs on a SEPARATE byte buffer AFTER the `func_imports` parse (`component_wrap.rs:95` parse precedes line 127 conversion), so it cannot desync the mapping even in principle. The shared-memory import (`merger.rs:2251`) is Memory/`component_idx: None` — skipped by the filter. The `debug_assert_eq!` ran live across real fusion (debug-profile suite) and never fired.
- **Consumers / None** — all 4 consumers take the corrected `Option`; `None` handling unchanged (find_task_return falls to global search; LocalResource falls to re-exporter). All-func case is identity.

Suites (debug, exit 0): lib 358, wit_bindgen_runtime 73, golden_e2e 4 (incl ls_cp_6), issue_112_mythos_v04 2.

PR #270, opened 2026-06-14
feat(p3-stream): canonicalize stream element types for cross-component pairing (#264)

### LINKED ISSUE #264: p3_stream: canonicalize stream element types to a cross-component structural descriptor (complete fix for LS-R-16)

avrabe:
## What

Completes the LS-R-16 fix (closes #264). The conservative guard (#265, shipped) declines to pair any stream whose element is a component-local `Type(N)` — safe, but leaves legitimate aggregate-element streams host-routed (benign under-match, Finding A). This resolves each endpoint's `Type(N)` through *that component's* type table to a **cross-component-stable structural key** (e.g. `record{count:u8,data:list<s32>}`), so aggregate-element streams pair **when and only when** producer and consumer element types are structurally identical.

## How

- `canonical_structural_key(comp, ty, depth)` walks `crate::parser::ComponentValType` recursively (record/tuple/list/fixed-list/option/result/variant/flags + primitives), resolving nested `Type(M)` via `get_type_definition`.
- **Injective key** — field names AND order, arities, element types, and discriminant cases all contribute, so structurally-different types yield different strings. A collision would re-introduce the exact **H-3.1 over-match LS-R-16 prevents** — this is the load-bearing safety property.
- **Conservative `None`** (stays non-pairable / host-routed) for: resource handles (`own`/`borrow` at any depth — cross-component resource identity overlaps the unresolved #256 attribution problem), depth-bound/cycle hits (`CANON_MAX_DEPTH = 64`), and unresolvable `Type(N)`.
- Wired into all three `StreamElement`-building sites, so `validate_stream_pair_graph` also compares structural keys.

## Tests (6 added)
- `canonical_key_format_and_injectivity` — `record{a:u8}` ≠ `record{b:u8}`, record ≠ tuple.
- `structural_aggregate_pairs_across_different_local_indices` — POSITIVE: same record at producer idx 1 / consumer idx 2 → exactly one pair.
- `structurally_different_aggregates_at_same_index_do_not_pair` — over-match prevention (the safety regression guard).
- `resource_element_stream_stays_host_routed`, `canonical_key_resource_returns_none`, `canonical_key_self_cycle_returns_none`.

The `ls_r_16_*` gate test is retained and re-documented: it now pins that a *conservatively-declined* element (resource/cycle/unresolvable, which keeps its raw `Type(N)`) stays non-pairable.

## Verification
- `cargo test -p meld-core --lib p3_stream` → 34/0 (exit 0).
- LS-N gate: 55/0/0 (independent run); build/clippy/fmt clean.

## Tier-5
Touches `p3_stream.rs` → Mythos delta-pass applies. Adversarial pass (hunting key collisions) posted separately.

### DIFF
--- a/meld-core/src/p3_stream.rs
+++ b/meld-core/src/p3_stream.rs
@@ -33,9 +33,20 @@
 
 use crate::parser::{
     CanonicalEntry, ComponentFuncDef, ComponentTypeKind, ComponentValType, ParsedComponent,
+    PrimitiveValType,
 };
 use std::collections::HashMap;
 use wasmparser::ComponentExternalKind;
+
+/// Depth bound for [`canonical_structural_key`] recursion.
+///
+/// Component-local type references (`Type(N)`) can in principle form
+/// cycles or pathologically deep nests in adversarial input. The
+/// structural canonicaliser returns `None` (→ non-pairable, host-routed)
+/// once this bound is exceeded, so a cyclic or unbounded type can never
+/// drive the canonicaliser into non-termination. 64 levels is far past
+/// any genuine aggregate nesting a real component declares.
+const CANON_MAX_DEPTH: usize = 64;
 
 /// The element type carried by a `stream<T>`, parsed from the
 /// component-type descriptor the parser records.
@@ -75,12 +86,17 @@ impl StreamElement {
     /// A bare `Type(N)` descriptor is a component-LOCAL type index —
     /// component A's `Type(5)` and component B's `Type(5)` are unrelated —
     /// so it is not a valid cross-component key (LS-R-16). Primitive
-    /// descriptors (`Primitive(..)` / scalar names) and the untyped stream
-    /// are globally stable and pairable; a local-index descriptor is not
+    /// descriptors (scalar names such as `u8`) and the untyped stream are
+    /// globally stable and pairable; a local-index descriptor is not
     /// (pairing on it risks an over-match that drives the bridge emitter to
-    /// wire incompatible streams). Canonicalising `Type(N)` to a structural
-    /// descriptor — which would make aggregate-element streams pairable — is
-    /// tracked follow-up; until then such streams stay host-routed.
+    /// wire incompatible streams).
+    ///
+    /// As of #264, an aggregate element that DID resolve structurally is
+    /// recorded as its structural key (e.g. `record{f0:u8,f1:list<s32>}`),
+    /// which contains no `"Type("` and so is pairable — and pairs only with
+    /// an identical key. An element that could NOT be canonicalised (a
+    /// resource handle, a depth-bound or cycle hit, or an unresolvable
+    /// index) keeps its raw `Type(N)` descriptor and stays host-routed.
     fn is_cross_component_pairable(&self) -> bool {
         match self {
             StreamElement::Untyped => true,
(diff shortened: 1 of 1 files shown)


avrabe:

Clean-room adversarial pass on the #264 diff (`p3_stream.rs`: `canonical_structural_key`, `canonicalize_stream_element`, call sites), targeting the load-bearing **injectivity** property — can two structurally-different element types produce the same structural key and manufacture a false cross-component pair (the H-3.1 over-match LS-R-16 prevents)?

**Collision classes attacked, each refuted with evidence:**
1. **Delimiter/grammar injection** (`{ } < > , : ( ) _` inside a field/case/flag/enum name) → **impossible**: the fuse pipeline parses with `validate=true` (`lib.rs:326`, `component_wrap.rs:3312`); `wasmparser::Validator` runs `to_kebab_str` on record fields/variant cases/flags/enum tags, and `KebabStr::is_kebab_case` admits only `[A-Za-z0-9]` and `-`. Every structural separator the key uses is rejected. Brute-force over 28 kebab-valid shapes → zero collisions.
2. **Absent-arm `result`** (`result<u8,_>` vs `result<_,u8>` vs `result<u8,u8>` vs `option<u8>`) → `_` is not a valid kebab name, so it's an unambiguous sentinel; all distinct.
3. **Empty aggregates** (`record{}`/`tuple<>`/`flags{}`/`variant{}`) → distinct keyword prefixes.
4. **Arity / nesting / field-order / `list` vs `fixed-list` / 1-tuple vs bare** → all distinct (brute-force).
5. **enum vs payloadless variant** → these DO key-collide (`variant{...}`), but it is **semantically sound**: an enum is a payloadless variant with identical canonical-ABI lifting, so pairing them carries no type confusion.
6. **Depth/cycle bound** → a self-cyclic `Type(N)` increments depth, hits `depth > 64` → `None` → host-routed; terminates.
7. **Resource leak via nesting** (`record{h:own<R>}`, direct `own<R>`) → `None` at any depth → keeps raw `Type(N)`, non-pairable.
8. **Mis-parse** → a bare primitive descriptor is `Primitive(U8)`; `parse_type_index` returns `None` for it, so canonicalisation fires only on a genuine `Type(N)`.

**Honest caveats (benign, recorded):**
- A primitive *type-alias* keys as `u8` while a bare primitive keys as `Primitive(U8)` → they **under**-match (miss a pair). Conservative (a missed optimisation), not unsafe.
- The enum/variant collision (5) is an intentional, sound equivalence.

Every structural-key-equal pair I could construct also shares canonical ABI, so key-equality ⇒ pairing-soundness. Injectivity holds under the validated parse path. Disposition: **NO PROVABLE FINDING.**

PR #271, opened 2026-06-14
feat(adapter): spec-faithful latin1+utf16 (CompactUTF16) transcoding (#253)

### LINKED ISSUE #253: Transcoding: cross-encoding pairs involving Latin-1/CompactUTF16 silently fall back to a raw byte copy (no transcoding)

avrabe:
## What

Closes #253. meld's `StringEncoding::Latin1` **is** the canonical-ABI `latin1+utf16` encoding (the Component Model has no pure-Latin-1 encoding). The prior untagged pure-Latin-1 treatment was correct only for the tag-clear case; a tag-set (utf16-represented) string was **mis-read**, and encoding *into* it (`Utf8/Utf16 → Latin1`) was unimplemented (fail-loud). This makes the whole path tag-faithful — producer and consumer agree on `UTF16_TAG = 1<<31` on the `(ptr, len)` length operand.

## How

- **Read side** — `emit_latin1_to_utf8` / `emit_latin1_to_utf16` branch on the tag: `count = len & !UTF16_TAG`; tag-set ⇒ source is UTF-16 (latin1→utf8 becomes utf16→utf8; latin1→utf16 a verbatim code-unit copy); tag-clear delegates to renamed `emit_latin1_pure_*` helpers — **byte-for-byte the prior behaviour** (existing SR-17 latin1 tests unchanged).
- **Write side** — `emit_utf8_to_latin1` / `emit_utf16_to_latin1`: two-phase — scan source code points; all ≤ 0xFF ⇒ Latin-1 (`len = n`, tag clear); else UTF-16 with surrogate pairs (`len = code_units | UTF16_TAG`). Decoders extracted from the hardened #251/#249 transcoders (shared, identical logic).
- `alignment_for_encoding(Latin1)` **1 → 2** (latin1+utf16 buffers may hold u16).
- Fail-loud `_` arm kept as the safety net (now unreachable among string encodings).

## Oracle (runtime-differential, under wasmtime)

9 new tests in `tests/adapter_safety.rs` with a **tag-aware callee** that reads u8 vs u16 by the tag — so a mis-set or mis-read tag fails the asserted per-representation sums. Both a **latin1-fits** (`"café"`) and a **utf16-requiring** (`"日本"`, `"A😀"`) string round-trip in every newly-enabled direction (Utf8→Compact, Utf16→Compact, Compact→Utf8/Utf16 with a tag-set source).

- `cargo test -p meld-core --test adapter_safety` → 21/0
- `cargo test -p meld-core --lib` → 364/0 (incl. the 35 SR-17 unit tests)
- LS-N gate → **56/0/0** (LS-P-21 added with its `ls_p_21_*` gate test — no regression)
- build / clippy `-D warnings` / fmt clean

## Traceability

Adds **approved LS-P-21** (latin1+utf16 tag mis-read/mis-write → wrong-encoding data, H-4.4) with the `ls_p_21_*` gate test; SR-17 note extended for the tagged-form coverage.

## Tier-5
Touches `adapter/fact.rs` → Mythos delta-pass applies. Adversarial pass (targeting tag-masking / two-phase-encoder correctness) posted separately.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -122,6 +122,54 @@ pub(crate) fn emit_overflow_guard(body: &mut Function, len_local: u32, k: u32) {
     body.instruction(&Instruction::End);
 }
 
+/// Push the byte length of a copied string/list buffer onto the stack.
+///
+/// For a `latin1+utf16` string (the `Latin1` encoding, `byte_mult == 1`) the
+/// `(ptr, len)` length operand is **tagged** (see [`UTF16_TAG`]): tag set ⇒ the
+/// payload is UTF-16, so the byte length is `(len & !UTF16_TAG) * 2`; tag clear
+/// ⇒ Latin-1, byte length `len`. A same-encoding cross-memory copy
+/// ([`generate_memory_copy_adapter`]) must use this tag-aware byte count for
+/// both the `cabi_realloc` size and the `memory.copy` length — copying the raw
+/// tagged operand would copy `count | 0x8000_0000` bytes (~2 GiB OOB read).
+/// The length operand FORWARDED to the callee stays tagged; only the internal
+/// byte count is masked here.
+///
+/// For every other buffer (`byte_mult != 1`, or a non-Latin1 encoding) the
+/// length is an untagged element count, byte length `len * byte_mult`. A plain
+/// `list<u8>` (also `byte_mult == 1`) is unaffected: a legitimate byte count is
+/// `< 2^31` (the [`emit_overflow_guard`] bound), so the tag bit is clear and
+/// the tag-aware path returns `len` unchanged.
+fn emit_copy_byte_count(
+    func: &mut Function,
+    len_local: u32,
+    byte_mult: u32,
+    is_compact_utf16: bool,
+) {
+    if is_compact_utf16 {
+        // bytes = (len & UTF16_TAG) ? (len & !UTF16_TAG) * 2 : len
+        func.instruction(&Instruction::LocalGet(len_local));
+        func.instruction(&Instruction::I32Const(UTF16_TAG));
+        func.instruction(&Instruction::I32And);
+        func.instruction(&Instruction::If(wasm_encoder::BlockType::Result(
+            wasm_encoder::ValType::I32,
+        )));
+        func.instruction(&Instruction::LocalGet(len_local));
+        func.instruction(&Instruction::I32Const(!UTF16_TAG));
+        func.instruction(&Instruction::I32And);
+        func.instruction(&Instruction::I32Const(2));
+        func.instruction(&Instruction::I32Mul);
+        func.instruction(&Instruction::Else);
+        func.instruction(&Instruction::LocalGet(len_local));
+        func.instruction(&Instruction::End);
+    } else {
+        func.instruction(&Instruction::LocalGet(len_local));
+        if byte_mult > 1 {
+            func.instruction(&Instruction::I32Const(byte_mult as i32));
+            func.instruction(&Instruction::I32Mul);
+        }
+    }
+}
+
 /// Emit the hardened UTF-8 → code-point decoder for one code point.
 ///
 /// Reads the lead byte at `mem8[ptr_local + src_idx_local]`, decodes a full
@@ -858,6 +906,12 @@ fn emit_patch_nested_indirections(
     realloc_func: u32,
     caller_memory: u32,
     callee_memory: u32,
+    // True when the *callee* string encoding (this is a result-side,
+    // callee → caller copy) is `latin1+utf16` (CompactUtf16). A byte-granular
+    // inner buffer (`sub_elem_size == 1`) is then a tag-encoded string whose
+    // byte count must be masked/doubled. The inner (ptr, len) header keeps its
+    // original (tagged) length — this function rewrites only the pointer.
+    callee_compact_utf16: bool,
 ) {
     let indirections = collect_indirections(elem_ty, 0);
     if indirections.is_empty() {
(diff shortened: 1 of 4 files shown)


avrabe:
## Mythos delta-pass disposition — NO PROVABLE FINDING (after a found-and-fixed OOB)

Two adversarial clean-room passes on the `adapter/fact.rs` diff.

**Pass 1 (transcoders + read/write + alignment)** — refuted hypotheses on write-side buffer sizing, tag-mask sign/width, read-side count, boundaries, and the alignment change. But it **found a confirmed memory-safety bug** (failing Kani harness + wasmtime OOB trap): meld's cross-memory **copy machinery** copied a `latin1+utf16` string's `(ptr,len)` as an untagged byte count, so a tag-set string copied `count | 0x8000_0000` bytes (~2 GiB OOB). Pre-existing, but this PR's transcoders make tag-set strings reachable through previously-fail-loud paths — so it was fixed here (commit `1a06279`) rather than deferred: every cross-memory string copy/realloc now routes through `emit_copy_byte_count` (tag-aware) with alignment 2 for the compact-utf16 case, across `generate_memory_copy_adapter`, `generate_params_ptr_adapter`, `generate_retptr_adapter`, `emit_inner_pointer_fixup`, and the async copy/writeback/nested paths. New regression test `ls_p_21_same_encoding_compact_tagset_param_copy_roundtrips` (the exact OOB PoC — 日本 round-trips to 52497) + a tag-clear control.

**Pass 2 (adversarial re-pass on the enlarged fix)** — **NO PROVABLE FINDING**. Refuted all 6 hypotheses:
1. *Wrong governing encoding per site* — sync copy adapters only fire when `caller_enc == callee_enc` (dispatch gate `!needs_transcoding()`), and each site independently picks the correct side (param→caller, result→callee) anyway.
2. *Forwarded length operand corrupted* — no site overwrites the forwarded/stored tagged `len` with the masked count; only the internal byte count is masked.
3. *Helper arithmetic / stack balance* — `If(Result I32)` balanced (both arms net +1 i32; corroborated by 23 wasmtime-validated tests); `(len & !UTF16_TAG)*2 ≤ 0xFFFFFFFE` cannot wrap u32; oversize → clean `cabi_realloc` trap, not OOB.
4. *Nested `list<u8>` false-positive* — real in code but requires a ≥2 GiB list (unreachable; the documented helper invariant), unprovable.
5. *Alignment 1→2* — one genuine omission (`emit_patch_nested_indirections` hardcoded align 1 for a nested compact-utf16 string). **Now fixed** (commit `1cc4c5c`) for consistency; was benign (wasm align is a hint).
6. *Async nesting* — `is_compact_utf16` threaded to every recursion level; inner-pointer headers keep their tagged length.

**Pre-existing, out-of-scope items noted for follow-up (not introduced here):** `generate_memory_copy_adapter`'s result path hardcodes `byte_mult=1` for any `[I32,I32]` return (would undercopy a `list<u32>`/`list<record>` return if that path is reached); the async adapter does no transcoding when caller/callee encodings differ. Both predate this PR — filing separately.

Disposition: **NO PROVABLE FINDING** on the current diff; the one confirmed bug was fixed and re-verified.

Issue #272, opened 2026-06-14
adapter: implement async cross-encoding string transcoding (capability; safety guard already landed #273)


avrabe:
Two pre-existing adapter-copy gaps surfaced (but not introduced) by the #253 adversarial Mythos re-pass on `meld-core/src/adapter/fact.rs`. Both predate #253's tag-aware copy work (PR #271) and were out of its scope.

## 1. Result-copy hardcodes `byte_mult = 1` for `[I32,I32]` returns (potential under-copy)
`generate_memory_copy_adapter`'s result-copy path (~fact.rs:2412/2419) copies a returned `(ptr, len)` pair with `byte_mult = 1`. For a function returning a **non-byte-element list** (`list<u32>`, `list<record{...}>`, etc.) whose flat return is `[I32, I32]`, the element stride is > 1, so a 1×-multiplier copy would **under-copy** (truncate) the returned buffer — `len` elements copied as `len` bytes.

- **Reachability:** unverified — depends on whether this path is actually reached for non-byte list returns vs. handled by `generate_retptr_adapter` / a retptr return-area copy. Needs a `compose → fuse → execute` differential with a `list<u32>`-returning callee across memories to confirm (or refute) before fixing.
- **Severity if reachable:** H-4.1 truncated/semantic-divergent return data (not OOB — under-copy reads less, doesn't overrun).
- **Fix direction:** derive the result element stride from the result `CopyLayout` (as the param path does) instead of hardcoding 1.

## 2. Async adapter does no transcoding when caller/callee string encodings differ
The async adapter paths (`emit_param_copy_step` / `emit_ptr_pair_result_writeback`) are dispatched (fact.rs ~6859) **before** any transcoding gate, and perform no string transcoding when `caller_string_encoding != callee_string_encoding`. Each side uses its own encoding consistently (so no mis-sizing), but a genuine cross-encoding async string would not be transcoded.

- **Severity:** lower — this is a capability gap, and the surrounding async emitter already errors clearly on unsupported cases (the documented async fail-loud). It is not silent corruption.
- **Fix direction:** route async string params/results through the same `(caller_enc, callee_enc)` transcoding dispatch the sync path uses, or fail loud explicitly for the cross-encoding async case.

Both found during the #253 (#271) Mythos passes; filing for triage/prioritization rather than bundling into the transcoding PR.

avrabe:
## Item 1 — REFUTED (not a live bug). Falsification-first investigation.

[...]

PR #273, opened 2026-06-14
fix(adapter): fail loud on async cross-encoding string instead of silent raw-copy (#272)

avrabe:
## What

Closes the async half of #272. A **confirmed** silent-corruption bug (runtime differential): `FactStyleGenerator::generate` routes every `is_async_lift` call to the async emitters and `continue`s **without** computing `needs_transcoding()`, bypassing the sync transcoding dispatch. The async copy helpers move string bytes via `memory.copy` with no encoding conversion — so a cross-encoding async-lift call passing a string **silently raw-copies the bytes under the wrong encoding** (H-4.4), the exact defect #253/#271 closed for the sync path, re-opened on async.

## Fix (fail-loud guard, not full transcoding)

`guard_async_cross_encoding_strings`, run at the top of the async branch: returns a clear `AdapterGeneration` error when `crosses_memory` AND caller/callee string encodings differ AND there's a byte-granular `(ptr,len)` param/result — instead of emitting a silently-wrong adapter. Matches the #253 fail-loud-over-silent-corruption philosophy. **Full async cross-encoding transcoding remains the tracked #272 capability follow-up.**

**Trips:** cross-memory + differing encoding + a byte-granular string-like buffer.
**Does NOT trip:** same-encoding async strings; cross-encoding async with no string (e.g. `list<u32>`, stride 4); same-memory (shared) async.

### Conservative limitation (documented)
`CopyLayout` can't distinguish `string` from `list<u8>` (both `Bulk{1}`), so a cross-encoding async call passing **only** `list<u8>` (encoding-agnostic, safe) will also fail loud. This errs fail-loud (recoverable) rather than risk an under-trip (silent corruption); it dissolves when full async transcoding lands. Reviewers: weigh whether this over-trip is acceptable vs. a common pattern.

## Verification
- Confirmed via runtime differential (the pre-fix async emitter raw-copies a cross-encoding string; kept as an `#[ignore]`d record).
- `cargo test -p meld-core --lib` 370/0; `--test adapter_safety` 23/0; LS-N gate **57/0/0** (LS-F-27 + 6 `ls_f_27_*` gate tests); build/clippy/fmt clean.

## Tier-5
Touches `adapter/fact.rs` → Mythos delta-pass applies. Adversarial pass (under-trip / over-trip) posted separately.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -6891,17 +6891,15 @@ impl FactStyleGenerator {
             _ => return Ok(()),
         };
 
-        // Is there a byte-granular (ptr, len) param or result — a plausible
-        // string whose bytes the async emitter would raw-copy?
-        let is_byte_granular = |cl: &crate::resolver::CopyLayout| -> bool {
-            match cl {
-                crate::resolver::CopyLayout::Bulk { byte_multiplier } => *byte_multiplier == 1,
-                crate::resolver::CopyLayout::Elements { element_size, .. } => *element_size == 1,
-            }
-        };
-        // A ptr-pair position with no recorded layout defaults to byte-granular
-        // in the emitters (`byte_mult` falls back to 1), so treat a bare
-        // position (no parallel layout) as a string too.
+        // Is there a byte-granular (ptr, len) param or result whose bytes the
+        // async emitter would raw-copy — at ANY nesting depth? A nested string
+        // (e.g. `list<string>`) has a non-byte-granular TOP-LEVEL layout
+        // (`Elements{element_size: 8}`) but a byte-granular inner pointer that
+        // `emit_patch_nested_indirections` raw-copies; the detection recurses
+        // into `inner_pointers` to match the emitter's depth (LS-F-27
+        // under-trip fix). A ptr-pair position with no recorded layout defaults
+        // to byte-granular in the emitters (`byte_mult` falls back to 1), so
+        // treat a bare position (no parallel layout) as a string too.
         let param_has_string = site
             .requirements
             .pointer_pair_positions
@@ -6911,7 +6909,7 @@ impl FactStyleGenerator {
                 site.requirements
                     .param_copy_layouts
                     .get(i)
-                    .map(is_byte_granular)
+                    .map(Self::layout_bears_byte_granular_buffer)
                     .unwrap_or(true)
             });
         let result_has_string = site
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos delta-pass disposition — NO PROVABLE FINDING (after a found-and-fixed under-trip)

Two adversarial clean-room passes on the async cross-encoding guard.

**Pass 1** — found a confirmed **under-trip** (Kani + concrete test): a string nested in `list<string>` (top-level `Elements{element_size:8}`, byte-granular `Bulk{1}` inner) slipped past the original top-level-only check, so `emit_patch_nested_indirections` raw-copied the inner string with no transcode — the H-4.4 silent mis-transcode the guard exists to stop, still reachable for nested strings. **Fixed** (commit b5f64f9): the detection now recurses through `Elements.inner_pointers` to any depth (`layout_bears_byte_granular_buffer`), matching the emitter's depth. New tests: `ls_f_27_async_cross_encoding_nested_list_string_{param,result}_fails_loud` + `..._nested_no_string_ok` (recursion must not over-trip on `list<list<u32>>`).

**Pass 2 (re-pass on the fix)** — **NO PROVABLE FINDING**. Enumerated every async emitter string-raw-copy site and confirmed the recursive guard is a conservative super-set of each:
- `emit_param_copy_step` — guard iterates the same `pointer_pair_positions`, recursing layouts; coverage equal-or-stronger (the param nested case is actually a safe over-trip — the param emitter doesn't even deep-copy nested strings).
- `emit_ptr_pair_result_writeback` — for a genuine `string`/`list<string>` result the resolver records `Bulk{1}` / `Elements{8,[Bulk{1}]}`, so the recursive guard trips (verified to depth-3 `list<list<string>>`).
- `emit_patch_nested_indirections` — its WIT-type walk (`collect_indirections`) is shallower than the guard's `inner_pointers` recursion (which descends option/result/variant payloads + deeper list nesting), so guard ≥ emitter.
- `inner_resources` ignored → sound (resource handles, not string buffers). `Bulk{>1}` / `Elements{element_size≠1}` → element-wise, encoding-independent → correctly not flagged. Conditional/optional pairs the guard doesn't iterate → the async emitter doesn't raw-copy them either (`option<string>` lowers to ≥3 result globals; the writeback fires only for exactly 2 i32 globals).
- `None`-encoding (`(_,_) => Ok`) and recursion termination (`CopyLayout` is a finite acyclic tree) refuted.

Verdict: the guard is **complete for its scope** — a conservative super-set of the async emitters' string-copy depth. The pre-existing `result_type=None` opaque-bytes fallback (merger.rs:179-181) is documented-by-design, untouched here, and not a cross-encoding-guard defect.

Disposition: **NO PROVABLE FINDING** on the current diff; the under-trip was found, fixed, and re-verified.

PR #275, opened 2026-06-14
feat(adapter): async UTF-8→UTF-16 string-param transcoding (#272 inc 1)

avrabe:
## What

Increment 1 of #272 (async cross-encoding string transcoding), staged from a scoping analysis. The async path previously **failed loud** on all cross-encoding strings (LS-F-27); this makes a **top-level UTF-8→UTF-16 string PARAM** crossing memory actually transcode. Result-side, nested `list<string>`, and all other directions **still fail loud** (guard narrowed in lockstep).

## How

- The sync per-direction transcoders can't be reused (they hardcode locals 0/1 and emit their own `Call`), so adds `emit_utf8_to_utf16_transcode_param` — a parameterized free fn built from the shared `emit_utf8_decode_codepoint` + `emit_utf16_encode_codepoint` helpers.
- Wired into `emit_param_copy_step` (the seam shared by **both** callback and stackful async variants): for a `caller=Utf8, callee=Utf16`, top-level `Bulk{1}` string param, replaces the raw `memory.copy` with the transcode loop (realloc `2*len`, align 2, LS-A-7 overflow/null guards) and **rewrites the forwarded `len` operand** to the output UTF-16 code-unit count.
- Scratch budget grown (callback 16→22, stackful 12→18) so the loop's locals don't collide with the nested-patch region.
- **Guard narrowed in lockstep:** allows through ONLY `crosses_memory + Utf8→Utf16 + top-level byte-granular param + no result strings + no nested strings`; every other combo still fails loud.

## Verification posture (please note)

- **Transcode loop: runtime-verified** under wasmtime (`tests/async_cross_encoding.rs`): `hello`→500, supplementary `"A😀"`→112254, malformed→U+FFFD, empty→0 — via a synthetic two-memory module driving the exact production emitter.
- **Integration** (`emit_param_copy_step` wiring, len-rewrite, scratch budget) is **inspection- and unit-verified, not e2e-run**: meld's full async-lift e2e (the callback/stackful trampoline under a real async runtime) is gated on **kiln**, out of scope per the repo's existing async-lift tests. The adversarial Mythos pass (posted separately) is the compensating control for the integration.
- Guard narrowing verified by `ls_f_27_*` unit tests (10 pass) + the dispatch test.
- lib 375/0; `async_cross_encoding` 4/0; LS-N gate 57/0/0; clippy/fmt clean.

## Note
Adds a `#[doc(hidden)] pub` test-support module builder (`build_utf8_to_utf16_transcode_test_module`) so the integration-test crate can drive the emitter under wasmtime.

## Tier-5
`adapter/fact.rs` → Mythos delta-pass applies.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -686,6 +686,281 @@ fn emit_utf16_encode_codepoint(
     func.instruction(&Instruction::End);
 }
 
+/// Emit a UTF-8 → UTF-16 transcode of a single top-level `(ptr, len)` string
+/// param on the async-lift cross-memory copy path (#272 inc 1).
+///
+/// Unlike the sync [`FactStyleGenerator::emit_utf8_to_utf16_transcode`], this
+/// helper is FULLY parameterised by local indices and MemArgs (no hidden
+/// local-0/1 / param-shape assumptions) and emits NO `Call` — it is composed
+/// from the shared, validated [`emit_utf8_decode_codepoint`] +
+/// [`emit_utf16_encode_codepoint`] codepoint helpers so it cannot drift from
+/// the sync decoder/encoder.
+///
+/// Contract:
+/// * `ptr_local` / `len_local` hold the CALLER-memory source pointer and the
+///   source byte length. On return, `ptr_local` is rewritten to the freshly
+///   reallocated CALLEE-memory output pointer and `len_local` is rewritten to
+///   the output UTF-16 **code-unit count** (NOT a byte count) — the operand a
+///   UTF-16 lifting callee reads. Failing to rewrite `len_local` would
+///   silently mis-size the callee's view of the string.
+/// * `out_ptr_local` receives the realloc result (the callee output buffer).
+/// * `src_idx_local`, `dst_idx_local`, `cp_local`, `byte_local`,
+///   `cont_local` are dedicated scratch i32 locals the caller must reserve;
+///   they must not alias `ptr_local`/`len_local` or each other.
+///
+/// Sizing (mirrors the sync emitter's LS-A-7 guards): each UTF-8 byte yields
+/// at most one UTF-16 code unit, so the worst case is `len` code units =
+/// `2 * len` bytes. `2 * len` is computed with an i32-overflow guard before
+/// the multiply (trap if `len > u32::MAX/2`), and the realloc return is
+/// null-checked, both via the shared helpers.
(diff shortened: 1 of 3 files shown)


avrabe:
## Mythos delta-pass disposition — found-and-fixed integration bug; verified

The adversarial pass was tasked to compensate for the verification gap (the runtime oracle drives the transcode *loop* via a synthetic module, not the real async adapter). It did exactly that and found a **confirmed integration bug** (failing Kani + a failing wasmparser-validation `#[test]` generating the real adapter):

**Finding (hazard 1 — callback async budget):** the callback adapter declared 22 i32 locals (top addressable `caller_params+21`), but the inc-1 UTF-8→UTF-16 transcode loop reaches `caller_params+25` (`transcode_base = l_p2+16`, uses `..+4`). So the generated **callback** adapter referenced undeclared locals → wasm validation failure (`unknown local 24`) for exactly the inc-1 string-param case. The synthetic oracle never hit this because its hand-built function declares enough locals. **Stackful was sized correctly.**

**Fix (commit 946ad77):** grow the callback budget 22 → 26 so the transcode tail locals (offsets 21..25 — previously past the budget, hence unused) are declared. Provably collision-free: the locals stay put and just get declared (vs moving `transcode_base` into possibly-occupied slots). Both async budgets re-verified: callback 26 (top offset 25 ≤ 25), stackful 18 (top offset 16 ≤ 17); no third async variant.

**Regression test (permanent, closes the integration gap):** `inc1_callback_adapter_transcode_locals_within_budget` generates the REAL callback adapter for the inc-1 site and asserts every referenced local is addressable — fails pre-fix (local 27 > 23 addressable), passes post-fix. This is the integration-level check the loop-only runtime oracle couldn't reach.

**Other hazards cleared by the pass (on logic unchanged by the fix):** (2) len/ptr rewrite — the transcode rewrites the same `ptr_local`/`len_local` the lift call reads, so forwarding is correct (stackful validates clean end-to-end); (3) guard-allow predicate == emitter-transcode predicate for the top-level `Bulk{1}` shape (no divergence → no silent-corruption under-trip); (4) realloc `2*len` with the overflow guard before the multiply, `len` is the UTF-8 byte count, CompactUtf16 can't reach this path (caller must be Utf8).

Disposition: the one finding is fixed and regression-tested at the integration level; remaining hazards cleared. The fix is a provably-safe budget constant on logic the pass already examined.

PR #276, opened 2026-06-14
feat(adapter): async UTF-16→UTF-8 string-param transcoding (#272 inc 2)

avrabe:
## What
Increment 2 of #272: top-level **UTF-16→UTF-8** string **param** crossing memory in an async-lift adapter now transcodes (mirror of inc 1). Result-side, nested, and all other directions stay fail-loud.

## How
- **Extracts a shared `emit_utf8_encode_codepoint`** from the sync `emit_utf16_to_utf8_transcode`'s inline 1/2/3/4-byte encoder (~170 lines of duplication → a call). The sync emitter's hardened suite (SR-17 utf16→utf8 supplementary / lone-surrogate / mid-string / malformed-matrix, LS-P-16/19, LS-A-7) **all still pass** — behavior-preserving.
- Adds `emit_utf16_to_utf8_transcode_param` (parameterized, no `Call`): source `emit_utf16_decode_codepoint` (lossy U+FFFD), dest the shared encoder. Wired into `emit_param_copy_step` for `caller=Utf16, callee=Utf8, Bulk{1}`. Realloc `3*len` (worst case 3 UTF-8 bytes/UTF-16 code unit) with the sync overflow guard; rewrites forwarded `len` to the output UTF-8 byte count.
- **Budgets computed, not assumed** (the inc-1 callback-budget bug lesson): 5 transcode locals; top index fits both budgets unchanged (callback +25 < 26; stackful +16 < 18) — proven by **two** new integration local-bounds tests (callback AND stackful; inc 1 only had callback).
- Guard narrowed in lockstep: now allows Utf8→Utf16 OR Utf16→Utf8 top-level param; result/nested/latin1 still fail loud (predicate == emitter triggers).

## Verification
- 5 loop runtime tests under wasmtime (ASCII→500, é→C3A9, 😀 surrogate pair→F0 9F 98 80, lone surrogate→U+FFFD, empty).
- 2 integration local-bounds tests (real callback + stackful adapters).
- Sync emitter refactor safety: full `adapter_safety` SR-17 utf16→utf8 + malformed-surrogate suite green.
- lib 379/0; `async_cross_encoding` 9/0; `adapter_safety` 23/0; LS-N gate 57/0/0; clippy/fmt clean.

Verification posture unchanged from inc 1: transcode loop runtime-verified via synthetic module; full async-lift e2e gated on kiln (out of scope); integration covered by the local-bounds tests + the Mythos pass.

## Tier-5
`adapter/fact.rs` → Mythos delta-pass applies (posted separately).

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -686,6 +686,192 @@ fn emit_utf16_encode_codepoint(
     func.instruction(&Instruction::End);
 }
 
+/// Emit the code-point → UTF-8 encoder for one code point.
+///
+/// Writes `cp_local` to `mem8[out_ptr_local + dst_idx_local]` as 1–4 UTF-8
+/// bytes (cp<0x80 → 1; <0x800 → 2; <0x10000 → 3; else → 4), advancing
+/// `dst_idx_local` by the number of bytes written. Extracted verbatim from the
+/// sync [`FactStyleGenerator::emit_utf16_to_utf8_transcode`] encode tail (#272
+/// inc 2) so the sync transcoder and the async param transcoder
+/// ([`emit_utf16_to_utf8_transcode_param`]) share ONE validated encoder and
+/// cannot drift. Uses no scratch beyond `cp_local` / `dst_idx_local`.
+fn emit_utf8_encode_codepoint(
+    func: &mut Function,
+    dst_mem8: wasm_encoder::MemArg,
+    out_ptr_local: u32,
+    dst_idx_local: u32,
+    cp_local: u32,
+) {
+    // if code_point < 0x80: 1-byte
+    func.instruction(&Instruction::LocalGet(cp_local));
+    func.instruction(&Instruction::I32Const(0x80));
+    func.instruction(&Instruction::I32LtU);
+    func.instruction(&Instruction::If(wasm_encoder::BlockType::Empty));
+    {
+        // out[dst_idx] = code_point
+        func.instruction(&Instruction::LocalGet(out_ptr_local));
+        func.instruction(&Instruction::LocalGet(dst_idx_local));
+        func.instruction(&Instruction::I32Add);
+        func.instruction(&Instruction::LocalGet(cp_local));
+        func.instruction(&Instruction::I32Store8(dst_mem8));
+        // dst_idx += 1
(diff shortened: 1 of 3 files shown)


avrabe:

Adversarial clean-room pass, sharpest focus on the encoder-extraction refactor of hardened sync code. All five hazards refuted with executable evidence (every probe oracle PASSED on the unfixed branch — no defect to fail on):

1. **Encoder-extraction regression (highest priority):** the extracted `emit_utf8_encode_codepoint` is **byte-identical** to the removed inline encoder (138 vs 138 instructions, IDENTICAL). A runtime boundary differential over the 1/2/3/4-byte UTF-8 boundaries the existing fixtures miss — U+007F/0080, U+07FF/0800, U+D7FF, U+FFFF, U+10000, U+10FFFF — matched `char::encode_utf8` exactly at every point. No drift in masks (0x80/0x3F), prefixes (0xC0/0xE0/0xF0), shifts (6/12/18), or dst advances (1/2/3/4).
2. **Realloc undersize:** `3*len` with the overflow guard applied BEFORE the multiply (identical to the sync emitter); 3 bytes/code-unit is the tight max (surrogate pair = 2 units → 4 bytes = 2/unit; BMP → ≤3). A worst-case 64×U+0800 probe (exactly 3*len) ran without trapping.
3. **Guard vs the two emitter triggers:** direction compared identically (global caller/callee encoding) in the guard and both `transcode_*` triggers; `byte_mult==1` aligns with the guard's `all_param_strings_top_level`. No allow-but-not-transcode (silent-corruption) gap.
4. **Budget (independently computed, not trusting the claim):** callback budget 26 → top addressable `caller_params+25`; transcode top `+25` (exact fit). Stackful budget 18 → top `+17`; transcode top `+16` (1 headroom). Both `inc2_*_locals_within_budget` integration tests pass.
5. **len-forwarding:** the param fn rewrites the same `ptr_local`/`len_local` the lift call forwards — not stale.

Existing suites green: `async_cross_encoding` 9/9, `adapter::fact` lib 44/44, full `adapter_safety` SR-17 utf16→utf8 + malformed-surrogate suite unaffected by the extraction.

Disposition: **NO PROVABLE FINDING.** The hardened-code refactor is byte-identical + boundary-verified; the inc-2 integration matches inc-1's now-proven pattern with budgets computed and tested for both async variants.

PR #277, opened 2026-06-14
feat(adapter): async string-RESULT transcoding, utf8↔utf16 (#272 inc 3)

avrabe:
## What
Increment 3: a top-level **string RESULT** crossing memory in an async-lift adapter now transcodes (utf8↔utf16). Completes the practically-common utf8↔utf16 case (params were inc 1+2). Nested results and latin1/CompactUtf16 directions stay fail-loud.

## How
- **Direction subtlety:** a result is produced by the callee (`callee_enc`) and read by the caller (`caller_enc`), so the result transcode is `callee_enc → caller_enc` — reverse of params. `emit_ptr_pair_result_writeback` gains a 3-way branch (Utf8→Utf16 realloc 2*len; Utf16→Utf8 realloc 3*len; else raw-copy+tag), SRC=callee mem, DST=caller mem, realloc in caller mem. The retptr receives the OUTPUT length.
- **Budget (computed, not assumed):** reuses the param `transcode_base` block — param transcode (pre-call) and result transcode (post-call) are never simultaneously live, and that block doesn't overlap the result-writeback/nested-patch region live during the result step. Top index fits both budgets unchanged (callback cp+25 < 26; stackful cp+16 < 18).
- **Guard:** allow = `param_ok && result_ok`; `result_ok` permits a top-level `Bulk{1}` result whose callee→caller direction is utf8↔utf16. Allow-predicate ≡ union of param+result emitter triggers.

## Verification
- `inc3_{callback,stackful}_adapter_result_string_locals_within_budget` generate the REAL adapters and assert local addressability.
- Guard tests: result string in utf8↔utf16 now allowed; nested result + CompactUtf16-direction result still fail loud.
- lib 382/0; `async_cross_encoding` 9/0; `adapter_safety` 23/0 (sync result-copy path intact); LS-N gate 57/0/0; clippy/fmt clean.

**Reviewer focus / known limit:** the load-bearing concern is **local-reuse collision** — the result-transcode scratch shares indices with the param `transcode_base` block, justified by pre-call vs post-call liveness. A bounds check (the integration tests) catches *overflow* but NOT a *collision* (two live locals at one index = valid-but-wrong wasm). The Mythos pass is tasked to verify no result-transcode local collides with any local live during `emit_ptr_pair_result_writeback`. Verification posture otherwise matches inc 1/2 (loops runtime-verified; full async-lift e2e gated on kiln).

## Tier-5

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -6544,6 +6544,15 @@ impl FactStyleGenerator {
         // undeclared → the generated callback adapter failed wasm validation
         // ("unknown local 24") for exactly the inc-1 UTF-8→UTF-16 string-param
         // case (the stackful variant was sized correctly). #272 inc 1 fix.
+        //
+        // #272 inc 3 (RESULT-side transcode): the result-writeback transcode
+        // REUSES the same `transcode_base = l_p2 + 16` block (offsets 21..=25,
+        // already declared). The param transcode (step 0.5) and the result
+        // transcode (step 3) are never simultaneously live, and that block does
+        // not overlap the result-writeback/nested-patch region `l_p2 + 1 ..=
+        // l_p2 + 10` (offsets 6..=15) live during step 3 — so no growth is
+        // needed; top addressable index stays offset 25 < 26 (proven by
+        // `inc3_callback_adapter_result_string_locals_within_budget`).
         let mut body = Function::new([(26, wasm_encoder::ValType::I32)]);
 
         // Step 0.5: copy string/list params from caller to callee memory.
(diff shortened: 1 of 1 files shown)


avrabe:

Adversarial pass with load-bearing focus on the **local-reuse collision** (which bounds checks can't catch). All five hazards refuted with executable evidence:

1. **Collision (primary) — proven ABSENT.** Full liveness enumeration during the result-transcode step: *callback* — callback-loop locals `cp..cp+5` are dead after the loop `End`; writeback-live = `cp+6/7/8`; retptr = `callee_param_count < cp`; result-transcode scratch = `cp+21..cp+25` → **disjoint**. *Stackful* — writeback-live `cp+1/2/3`, scratch `cp+12..cp+16` → disjoint. Confirmed concretely: generated the REAL callback AND stackful adapters for an `f(s:string)->string` site (BOTH param + result transcodes wired, both reusing `transcode_base`), assembled each into a full module, and `wasmparser::Validator::validate_all` **passed** for caller=Utf16/callee=Utf8 and the mirror — a clobber would surface as an OOB-local/stack/type error. (The shipped `inc3_*_within_budget` tests only check max-index; this probe closed the stack-typing gap and still passed.)
2. **Param-before-call** — param transcode rewrites caller param locals (0..cp), consumed by the lift `Call`; nothing reads `transcode_base` between that and the result step.
3. **retptr** — after the result loop, `l_dst_ptr` = realloc output ptr, `l_src_len` = OUTPUT count; store is `retptr[0]=l_dst_ptr, retptr[4]=l_src_len`; realloc uses `caller_realloc` + caller memory, align 2/1. Correct.
4. **Guard vs emitter** — consistent for all reachable types: `String`→guard allows + emitter transcodes; `list<u8>`→guard allows + emitter raw-copies (correct, byte lists encoding-independent). **Latent fragility (not a provable bug):** if `result_type` resolves to `None` (name/flat-match failure) while the layout stays `Bulk{1}`, the emitter's None-fallback transcodes — correct for the only constructible such case (a genuine string), but no valid input was found yielding `Bulk{1}`+`None`+non-text data. Recorded as a hardening candidate (key the emitter's transcode decision off the same layout the guard uses, or assert `result_type.is_some()`).
5. **Realloc sizing** — utf8→utf16 `2*len`, utf16→utf8 `3*len`; overflow guard precedes the multiply (verified in dumped WAT); null-checked realloc.

Disposition: **NO PROVABLE FINDING.** The collision is rigorously proven absent (real adapter validates for the both-transcodes case); the one latent fragility is non-constructible and noted for hardening.

PR #278, opened 2026-06-14
feat(adapter): async latin1-SOURCE param transcoding (#272 inc 4a)

avrabe:
## What
Inc 4a of #272: async PARAM-side transcoding where **caller=CompactUtf16** (latin1+utf16) → callee **Utf16** or **Utf8**. First slice of the latin1/CompactUtf16 tail. Dest-latin1 (`X→CompactUtf16`, two-phase) and result-side latin1 stay **fail-loud** as later sub-increments.

## How
Two parameterized loop fns mirroring the hardened sync latin1 emitters' **tag dispatch** (mask `UTF16_TAG`, branch latin1-source vs utf16-source):
- `emit_latin1_to_utf16_transcode_param`: tag-clear → zero-extend Latin-1 byte → UTF-16 unit; tag-set → verbatim u16 copy. Realloc `2*count`, align 2.
- `emit_latin1_to_utf8_transcode_param`: tag-clear → Latin-1 byte via shared `emit_utf8_encode_codepoint`; tag-set → `emit_utf16_decode_codepoint` + utf8 encode (lossy U+FFFD). Realloc `3*count`, align 1.
Both overflow-guarded; `len` rewritten to output count. Wired into `emit_param_copy_step`.

**Budget (computed):** Latin1→Utf8 uses 6 scratch (vs 5 for utf8↔utf16) → callback grew 26→27 (one extra tail local, collision-free); stackful unchanged (cp+17 < 18). Proven by `inc4a_{callback,stackful}_adapter_latin1_source_locals_within_budget`.

**Guard:** `param_ok` now also allows CompactUtf16→Utf16/Utf8 top-level; dest-latin1 params, all latin1 results, nested still fail loud.

## Verification
- 6 runtime loop tests (both directions × tag-clear "café" / tag-set "日本" / empty) — exercises BOTH the latin1-source and utf16-source tag branches.
- 2 integration local-bounds + 5 guard scope tests.
- lib 389/0; `async_cross_encoding` 15/0; `adapter_safety` 23/0 (sync SR-17 latin1 + #253 intact); LS-N gate 57/0/0; clippy/fmt clean.

**Reviewer focus:** the tag-dispatch port of hardened sync logic (both branches), realloc sizing per tag branch, and the callback budget 26→27. Verification posture matches prior increments (loops runtime-verified; full async-lift e2e gated on kiln; integration via local-bounds tests + Mythos pass).

## Tier-5

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -1067,6 +1067,320 @@ fn emit_utf16_to_utf8_transcode_param(
     body.instruction(&Instruction::LocalSet(len_local));
 }
 
+/// Emit a `latin1+utf16` (CompactUTF16) → UTF-16 transcode of a single
+/// top-level `(ptr, len)` string param on the async-lift cross-memory copy path
+/// (#272 inc 4a).
+///
+/// The CALLER encoding is `latin1+utf16` — meld's [`StringEncoding::Latin1`] —
+/// whose length operand is **tag-encoded** (see [`UTF16_TAG`]). This helper
+/// mirrors the SYNC [`FactStyleGenerator::emit_latin1_to_utf16_transcode`] but
+/// is FULLY parameterised by local indices / MemArgs (no hidden local-0/1
+/// assumptions) and emits NO `Call`, so it composes into the async param-copy
+/// step exactly like the inc-1/2 transcoders.
+///
+/// Runtime tag dispatch on `len_local`:
+/// * tag **CLEAR** → the source is pure Latin-1: `count` (= `len`) bytes, each
+///   0x00–0xFF zero-extended to one UTF-16 code unit (1:1, total). Output =
+///   `count` code units.
+/// * tag **SET** → the source is already UTF-16: `count` (= `len & !UTF16_TAG`)
+///   code units, copied VERBATIM (2 bytes per unit, surrogate pairs preserved
+///   byte-for-byte). Output = `count` code units.
+///
+/// Contract:
+/// * `ptr_local` / `len_local` hold the CALLER-memory source pointer and the
+///   TAGGED source length. On return, `ptr_local` is rewritten to the freshly
+///   reallocated CALLEE-memory output pointer and `len_local` is rewritten to
+///   the output UTF-16 **code-unit count** (UNTAGGED — a UTF-16 lifting callee
+///   reads a plain code-unit count).
+/// * `out_ptr_local` receives the realloc result.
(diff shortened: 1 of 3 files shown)


avrabe:

Adversarial pass, sharpest focus on the tag-dispatch port of hardened sync code. All five hazards refuted with executable evidence:

1. **Tag-dispatch byte-equivalence (both branches, both fns):** 8 runtime differentials vs an independent Rust reference — latin1 0x00/0x7F/0x80/0xFF (tag-clear), tag-set BMP + UTF-8-length boundaries (U+07FF/0800, U+FFFF/10000), supplementary surrogate pair (verbatim for latin1→utf16), lone-high-at-EOI, mid-string lone surrogate, lone-low — **all pass**. Tag mask, count computation, per-branch stride (1 byte/char latin1 vs 2 bytes/unit utf16), zero-extend, and lossy-U+FFFD (tag-set utf8 arm reuses the shared decode+encode; latin1→utf16 tag-set is verbatim, correctly no FFFD) all match the hardened sync emitters.
2. **Realloc bound per branch:** Latin1→Utf16 `2*count` (exact both arms); Latin1→Utf8 `3*count` (tag-clear ≤2/byte; tag-set ≤3/BMP-unit, surrogate pair 2 units→4 bytes = 2/unit). The SAME masked `count` drives both realloc size and loop bound; high bit masked before `*k` and the loop compare. No overrun.
3. **Budget:** measured the REAL callback latin1→utf8 adapter — max referenced local 26 < 27 declared (the +1 growth necessary, sufficient, collision-free; the 6 scratch locals are contiguous and past the live writeback/nested-patch region). Stackful top +17 < 18.
4. **Guard vs 4 emitter triggers:** `direction_is_implemented` = {Utf8→Utf16, Utf16→Utf8, CompactUtf16→Utf16, CompactUtf16→Utf8} on raw `CanonStringEncoding`, identical to the 4 `transcode_*` triggers; `canon_to_string_encoding` injective over the 3 canon values, so no spurious collapse. No allow-but-not-transcode gap.
5. **Still-fail-loud holds:** Utf8/Utf16→CompactUtf16 params, all latin1 results, nested latin1-source list<string> → Err.

**Test-quality note (not a finding):** `assert_locals_within_budget` hardcodes `caller_param_count=2`, loosening the slack — masks nothing here (the real budget was verified independently), but I'll tighten it to derive the real count before inc 4b, since it's the integration control that adds value as later increments grow the local set.

Disposition: **NO PROVABLE FINDING.** Tag-dispatch byte-verified against hardened sync code; budget measured on the real adapter.

PR #279, opened 2026-06-14
feat(adapter): async dest-latin1 (two-phase) param transcoding (#272 inc 4b)

avrabe:
## What
Inc 4b: async PARAM-side transcoding where **callee=CompactUtf16** (latin1+utf16), caller=**Utf8** or **Utf16** — the two-phase, tag-PRODUCING encoders. **Completes all 6 param directions.** Result-side latin1 and nested strings stay fail-loud.

## How
`emit_utf8_to_latin1_transcode_param` / `emit_utf16_to_latin1_transcode_param` (parameterized), mirroring the #253-hardened sync two-phase emitters:
- **Phase A** scans source code points (shared decode helpers); any cp > 0xFF ⇒ output is UTF-16 (`len = code_units | UTF16_TAG`, surrogate pairs for >0xFFFF), else Latin-1 (`len = char count`, tag clear).
- **Phase B** writes that representation. Scan and write use the **same decode logic** so the representation decision and the bytes written agree.
- Realloc `2*len` (utf16 worst case bounds both arms), align 2, overflow-guarded; ptr/len rewritten to out buffer + tagged length.

**Budget:** 6 scratch reused across scan + both write loops (same six inc-4a's Latin1→Utf8 uses) — callback top cp+26 < 27, stackful cp+17 < 18, **no growth**.

**Guard:** all 6 param directions now allowed; latin1 results + nested still fail loud.

**Helper accuracy fix** (inc-4a Mythos note): replaces hardcoded `caller_param_count=2/1` in `assert_locals_within_budget` with `site_caller_param_count` (derives the real count from the site's func type) at all 12 call sites — integration bounds are now tight.

## Verification
- 8 runtime loop tests (both directions × latin1-fits "café" / non-latin1 "日本" / supplementary 😀 / empty) — assert the **tag bit** (clear for latin1, set for non-latin1/supplementary) AND output bytes.
- 2 integration local-bounds (accurate helper).
- lib 392/0; `async_cross_encoding` 23/0; `adapter_safety` 23/0 (sync SR-17 + #253 latin1 intact); LS-N gate 57/0/0; clippy/fmt clean.

**Reviewer focus:** scan/write representation agreement (a cp near the 0xFF boundary classified differently between passes would truncate or mis-tag), the `2*len` bound, the tag-set output, and the helper fix. Verification posture matches prior increments.

## Tier-5

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -1381,6 +1381,386 @@ fn emit_latin1_to_utf8_transcode_param(
     body.instruction(&Instruction::LocalSet(len_local));
 }
 
+/// Emit a UTF-8 → `latin1+utf16` (CompactUTF16) transcode of a single top-level
+/// `(ptr, len)` string param on the async-lift cross-memory copy path
+/// (#272 inc 4b).
+///
+/// The CALLEE encoding is `latin1+utf16` — meld's [`StringEncoding::Latin1`] —
+/// a **two-phase / tag-PRODUCING** encoder: it must decide a representation for
+/// the WHOLE string before writing it, then emit a length operand carrying the
+/// [`UTF16_TAG`] bit to tell the callee which representation was chosen. This
+/// helper mirrors the SYNC
+/// [`FactStyleGenerator::emit_utf8_to_latin1_transcode`] but is FULLY
+/// parameterised by local indices / MemArgs (no hidden local-0/1 assumptions)
+/// and emits NO `Call`, so it composes into the async param-copy step exactly
+/// like the inc-1/2/4a transcoders. It is composed from the shared, validated
+/// [`emit_utf8_decode_codepoint`] (source) + [`emit_utf16_encode_codepoint`]
+/// (UTF-16-arm output) helpers so it cannot drift from the sync decoder/encoder.
+///
+/// Two phases over the UTF-8 source:
+/// * **Phase A (scan):** decode every code point; if ANY code point is
+///   `> 0xFF`, the string cannot be Latin-1, so set `needs_utf16`.
+/// * **Phase B (write):**
+///   * `needs_utf16` SET → re-decode and UTF-16-encode each code point
+///     (surrogate pair for `> 0xFFFF`); output length = code-unit count with
+///     [`UTF16_TAG`] **set**.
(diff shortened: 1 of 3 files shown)


avrabe:

Adversarial pass, sharpest focus on the two-phase scan/write agreement. All six hazards refuted with executable evidence:

1. **Scan/write decode equivalence (primary) — structurally guaranteed.** Phase A (scan) and phase B (write) call the **same** shared `emit_utf8_decode_codepoint`/`emit_utf16_decode_codepoint` with **identical** args (same `src_idx`/`len`/scratch), so they cannot diverge on representation OR source advance. Runtime differential vs an independent Rust reference (Canonical-ABI U+FFFD + `>0xFF ⇒ utf16`): U+00FF (latin1, tag clear), U+0100 (utf16, tag set), mixed "aé好" (whole string utf16), invalid UTF-8 (0xFF lead, truncated 3-byte, overlong 0xC1 0x81), UTF-16 lone-high-at-EOF / lone-low / high+non-low — **9/9 pass** (tag + exact bytes).
2. **Realloc `2*len`** — output code units ≤ source units/bytes ≤ 2*len; all-supplementary "😀😀" (4 units → 8 bytes = exactly 2*len) no overrun. Same masked `len` drives realloc `*2` and the write-loop bound; overflow-guarded + null-checked.
3. **Tag-on-code-unit-count — correct.** utf16 arm `len = dst_idx | UTF16_TAG` (dst_idx = code-unit count); latin1 arm `len = dst_idx` (byte count), tag clear.
4. **Budget — confirmed.** Callback `transcode_base = cp+21`, top `cp+26 < 27`; stackful `cp+12`, top `cp+17 < 18`; scan + both write loops reuse the same 6 scratch (not disjoint).
5. **`site_caller_param_count` == emitter count** — byte-for-byte the emitters' derivation (`import_func_type_idx` → `type_index_map` → `params.len()`); the integration bounds now check the exact emitter layout.
6. **Guard vs 6 triggers / fail-loud** — allow-set = exactly the 6 param directions, matching the emitter triggers; latin1 results + nested dest-latin1 still fail loud.

Full suite: 32/32 runtime differentials (23 existing + 9 probes) + 57 fact.rs unit tests.

Disposition: **NO PROVABLE FINDING.** The two-phase agreement is structurally guaranteed (shared decode helpers); budget and helper-fix verified against the real emitter layout. This completes all 6 async param directions.

PR #282, opened 2026-06-14
feat(adapter): async nested list<string> RESULT transcoding, utf8→utf16 (#272 inc 5a)

avrabe:
## What
First nested-string slice: a `list<string>` **RESULT** crossing memory in an async-lift adapter, direction callee=Utf8 → caller=Utf16, now transcodes its inner strings (instead of fail-loud). Other directions, nested params, records/deeper, and nested `list<u8>` stay fail-loud / raw-copy.

## The blocker, resolved correctly (not papered over)
A nested `string` and a nested `list<u8>` both lower to `Bulk{1}` — indistinguishable in `CopyLayout`. Transcoding a `list<u8>` as UTF-8 would silently corrupt arbitrary binary. Fix: `collect_indirections` still has the WIT type, so it now returns `is_string` per inner indirection; the emitter transcodes **only** when `is_string`. A nested `list<u8>` keeps the verbatim raw-copy — **proven** by `inc5a_async_nested_list_u8_result_not_transcoded` (byte-sum unchanged).

## How
- `collect_indirections` → `Vec<(offset, sub_elem_size, is_string)>`; callers updated (lib.rs stabilizing-shim path ignores it — byte-preserving).
- `emit_patch_nested_indirections` gains caller/callee encoding + transcode_base; for `is_string` + Utf8→Utf16 replaces the inner `memory.copy` with `emit_utf8_to_utf16_transcode_param`. **MemArgs are the RESULT direction** (src=callee, dst=caller — opposite the flat param call).
- **NEW per-element header len store**: transcoded code-unit count at `rec_dst+offset+4` (overwriting the stale callee byte length) — proven by the non-ASCII len-rewrite test.
- **Budget under simultaneous liveness** (transcode runs *inside* the nested loop): nested region and transcode block disjoint — callback top offset 25 < 27, stackful 16 < 18, no growth. Machine-verified on the real adapter.
- **Guard**: adds `result_wit_types` to `AdapterRequirements` (resolver-populated from the same WIT results the emitter uses) so the guard's nested-allow predicate reads the **same** `is_string` signal as the emitter — no allow-but-raw-copy gap.

## Verification
- Runtime: `list<string>` transcodes (["Hi","yo"]→409), non-ASCII len rewrite (é→1 unit), `list<u8>` NOT transcoded (verbatim).
- Integration local-bounds (both variants, real adapter, simultaneous liveness).
- Guard tests: allowed only for is_string + utf8→utf16; list<u8>, other directions, nested params all fail-loud.
- lib 400/0; `async_cross_encoding` 26/0; `adapter_safety` 23/0 (sync result-copy + stabilizing shim intact); LS-N gate 57/0/0; full suite green.

**Reviewer focus:** (1) the `list<u8>` non-corruption guarantee; (2) guard `result_wit_types` ≡ emitter `collect_indirections` signal (no allow-but-raw-copy); (3) the header len-store (transcoded, not source); (4) simultaneous-liveness budget; (5) callee→caller MemArg direction.

## Tier-5
`adapter/fact.rs` + `resolver.rs` → Mythos delta-pass (posted separately).

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -2646,6 +2646,312 @@ fn build_x_to_latin1_transcode_test_module(utf16_source: bool) -> Vec<u8> {
     module.finish()
 }
 
+/// Build a self-contained two-memory wasm module that exercises the EXACT
+/// async-path NESTED-result inner copy emitter
+/// ([`emit_patch_nested_indirections`]) for a `list<string>` RESULT crossing
+/// memory in the callee=Utf8 → caller=Utf16 direction, for the #272 inc-5a
+/// runtime-differential oracle.
+///
+/// This is the RESULT direction (callee PRODUCES, caller READS): SRC = callee
+/// memory 1 (where the async-lift callee wrote the UTF-8 inner strings + the
+/// outer list of `(ptr, len)` records), DST = caller memory 0 (where the caller
+/// reads UTF-16). The host pre-writes BOTH memories to simulate the post-outer-
+/// bulk-copy state:
+/// * callee memory 1: inner UTF-8 strings, and the outer list of 8-byte
+///   `(inner_ptr, inner_len)` records pointing at them.
+/// * caller memory 0: a verbatim copy of the outer records (the stale callee
+///   ptrs + byte-lengths the outer `memory.copy` produced).
+///
+/// The exported `patch_and_sum(outer_caller, outer_callee, count) -> i32` runs
+/// the production [`emit_patch_nested_indirections`] (UTF-8 → UTF-16), which for
+/// EACH record transcodes the inner string into a freshly `cabi_realloc`'d
+/// caller-memory buffer, rewrites the caller-side inner `(ptr, len)` header to
+/// `(out_ptr, out_code_unit_count)`, then SUMS the UTF-16 code units of every
+/// inner string by re-reading the PATCHED `(ptr, len)` headers from caller
+/// memory. A raw byte-copy (the pre-inc-5a behaviour) would leave UTF-8 bytes
+/// the summing loop reads as 16-bit units (a different sum) AND leave the inner
+/// len as the UTF-8 byte count — so a passing assertion proves real transcoding
+/// AND the NEW header-len rewrite, not a copy.
+///
+/// `#[doc(hidden)] pub` purely so the `async_cross_encoding` runtime test can
+/// instantiate it; not a supported API.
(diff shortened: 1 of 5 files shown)


avrabe:

Adversarial pass, sharpest focus on the `list<u8>` corruption blocker + guard/emitter divergence. All six hazards refuted:

1. **`list<u8>` vs string (the blocker) — sound.** `collect_indirections` sets `is_string=true` ONLY for `CVT::String`; `CVT::List` (incl. `list<list<u8>>` one level deeper) → `false`, never recursed/transcoded. Probes through the REAL `emit_patch_nested_indirections`: `list<record{u32,string}>` (string at offset 4) and `list<tuple<string,string>>` (offsets 0/8) transcode exactly the strings (é→0x00E9); `list<u8>` raw-copied verbatim. Blocker resolved correctly.
2. **Guard vs emitter signal — converge** from the same source-component WIT (`result_wit_types` ← resolver `results`; emitter `info.result_type` ← shim). *Low-severity imprecision (not a finding, not corruption):* a multi-result `(list<string>, u32)` passes the guard's `.all()` vacuously (the u32 has no inner indirections), but the emitter is driven by `result_type`/`result_copy_layouts`/`result_globals`, not `result_wit_types`, so no mis-transcode. Future-tightening note.
3. **Header len store — correct:** transcoded UTF-16 code-unit count at `rec_dst+offset+4` in CALLER memory, conditional on transcoding; raw-copied buffers keep their original len. (A stale byte-len would over-read — the non-ASCII test would catch it.)
4. **Simultaneous liveness — disjoint:** nested `l_p2+5..+10`, transcode `l_p2+16..+20`, `l_new_ptr=l_p2+9` untouched by the transcode scratch. Machine-verified on the real adapter (top index < 27/18).
5. **MemArg direction — correct:** src=callee, dst=caller (opposite the param call); a swap would read garbage.
6. **Resolver/arity — correct:** `result_wit_types` populated at both resolver sites; all 3 `collect_indirections` callers updated; lib.rs stabilizing shim ignores `is_string` (byte-preserving) — sync path intact.

*Second low-severity note (fail-safe, not a finding):* the transcode-inner branch traps on a garbage callee `(ptr,len)` rather than skipping (the raw-copy path skips); a trap is fail-safe and trusts the same callee header inc-3's top-level result transcode already does.

Disposition: **NO PROVABLE FINDING.** The blocker is verified absent across nested/mixed shapes; the two notes are non-corrupting and recorded for future tightening.

PR #283, opened 2026-06-15
feat(adapter): async nested list<string> RESULT — remaining 5 directions (#272 inc 5b)

avrabe:
## What
Completes nested-result transcoding: extends inc-5a's single-direction (Utf8→Utf16) nested slot to a full **6-way** result-direction dispatch, reusing the verified loop fns. **All 6 nested-`list<string>`-RESULT directions now transcode.** Nested params + records/deeper stay as-is; nested `list<u8>` stays raw-copied (all directions).

## How
Added Utf16→Utf8, CompactUtf16→Utf16/Utf8, Utf8/Utf16→CompactUtf16. The **dest-latin1** directions write the **tagged** inner-header length (Latin-1 tag-clear / UTF-16 `| UTF16_TAG` tag-set), matching the flat dest-latin1 result path; source-latin1 handles a tag-set inner source buffer. Each loop rewrites `l_buf_len` to its (possibly tagged) output length, stored at `rec_dst+offset+4`.

**Budget (widest 6-scratch loop, simultaneous liveness):** the nested call passes the same `transcode_base` the flat writeback reserves → 6-scratch block at offsets 21..26 (callback) / 12..17 (stackful), disjoint from the nested region (10..15 / 5..10) and `l_new_ptr` (9). Top 26 < 27 / 17 < 18 — no growth. Proven on the real dest-latin1 nested adapter.

**Guard:** `result_nested_string_ok = result_dir_is_implemented (all 6) && result_inner_all_strings`; allow-set ≡ emitter's 6 nested triggers. Nested `list<u8>` + nested params still fail loud.

## Verification
- 5 runtime tests (per direction, latin1-fits + tag-driven; dest-latin1 asserts the TAGGED inner len on a non-latin1 string).
- Guard tests (all-directions-allowed; `list<u8>`-not-transcoded over all 6 dirs).
- 2 integration local-bounds (real dest-latin1 nested adapter, both variants).
- lib 403/0; `async_cross_encoding` 31/0; `adapter_safety` 23/0; LS-N gate 57/0/0; clippy/fmt clean.

**Reviewer focus:** the tagged inner-header length for dest-latin1 nested results, the 6-scratch budget under simultaneous liveness, and the `list<u8>`-not-transcoded guarantee across all 6 directions.

## Tier-5

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -2705,6 +2705,145 @@ pub fn build_nested_list_u8_result_not_transcoded_test_module() -> Vec<u8> {
     )
 }
 
+/// Build a two-memory wasm module that exercises the production
+/// [`emit_patch_nested_indirections`] emitter for a `list<string>` RESULT in an
+/// ARBITRARY result direction (`callee_enc → caller_enc`), for the #272 inc-5b
+/// runtime-differential oracle covering the five non-utf8→utf16 directions.
+///
+/// Unlike the inc-5a `patch_and_sum` builder, this one does NOT sum in wasm
+/// (the dest-latin1 tag and the per-direction read width make a wasm sum
+/// awkward). It only runs the patcher, then exports `caller_memory` so the host
+/// reads back the PATCHED outer `(ptr, len)` records and the transcoded inner
+/// buffers and asserts directly in Rust. The exported `patch(outer_caller,
+/// outer_callee, count)` runs the production emitter (SRC = callee memory 1,
+/// DST = caller memory 0) for the given direction; for dest-latin1 directions
+/// the inner len header is rewritten to the TAGGED output length, which the
+/// host inspects.
+///
+/// `#[doc(hidden)] pub` purely so the `async_cross_encoding` runtime test can
+/// instantiate it; not a supported API.
+#[doc(hidden)]
+pub fn build_nested_list_string_result_test_module(
+    caller_enc: crate::parser::CanonStringEncoding,
+    callee_enc: crate::parser::CanonStringEncoding,
+) -> Vec<u8> {
+    use crate::parser::ComponentValType;
+    use wasm_encoder::{
+        CodeSection, ConstExpr, ExportKind, ExportSection, FunctionSection, GlobalSection,
+        GlobalType, MemorySection, MemoryType, Module, TypeSection, ValType,
+    };
+
+    let callee_compact_utf16 =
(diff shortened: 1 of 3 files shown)


avrabe:

Adversarial pass, sharpest focus on the 6-scratch simultaneous-liveness collision (the wider latin1 loops are new here). All five hazards refuted, with the **full `wasmparser::Validator`** run on the REAL adapters (the strongest collision oracle, expressible here):

1. **6-scratch collision — proven disjoint, both variants, real-adapter-validated.** Callback: nested-loop locals `l_p2+5..=l_p2+10` (offsets 10..15); 6-scratch transcode `l_p2+16..=l_p2+21` (offsets 21..26); out-ptr sink `l_new_ptr=l_p2+9` is intentionally the nested new_ptr. Disjoint (gap 16..20), 6th local offset 26 < 27. Stackful: nested 5..10, transcode 12..17, disjoint, 17 < 18. **Generated the real callback + stackful adapters for all 6 directions and ran `Validator::validate_all` — all validate** (catches OOB/aliased locals); no transcode-scratch index equals any of the 6 live nested-loop locals.
2. **Tagged inner-header len (dest-latin1):** each element's `len_local` is rewritten from its OWN Phase-A scan (tag-clear char count / `code_units | UTF16_TAG`) and stored at `rec_dst+offset+4`. Mixed-element differential (record 0 latin1-fitting tag-CLEAR, record 1 non-latin1 tag-SET) passes both dest-latin1 directions — per-element tagging is independent.
3. **Per-direction loop-fn/realloc/MemArg:** all 6 callsites match (5/5/4/6/6/6 scratch; correct align; src=callee/dst=caller). No mis-wiring/swap.
4. **`list<u8>` not transcoded, all 6:** `transcode_inner = is_string && inner_transcode_dir`; nested `List` → `is_string=false` → raw-copy (no tag, original len) in every direction; guard rejects nested list<u8> in all 6.
5. **Guard ≡ emitter:** `result_dir_is_implemented` = the same 6-predicate set as the emitter's `inner_transcode_*`; `result_inner_all_strings` mirrors the `is_string` gate. No allow-but-raw-copy.

`adapter::fact` lib (68) + `async_cross_encoding` (31) all pass.

Disposition: **NO PROVABLE FINDING.** Collision proven absent AND real-adapter-validated for all 6 directions; per-element tagging, dispatch, and the list<u8> guarantee verified. Completes all 6 nested-result directions.

PR #284, opened 2026-06-15
feat(adapter): async nested list<string> PARAM transcoding + deep-copy (#272 inc 5c-a, closes #281)

### LINKED ISSUE #281: candidate: async adapter may not deep-copy nested string/list pointers in list<string> params (pre-existing, reachability unconfirmed)
**Status: CANDIDATE — surfaced by code-reading during #272 inc-5 scoping; reachability NOT runtime-confirmed.** Filing for investigation, not as a confirmed bug (a prior #272 code-read reachability claim — item 1 — turned out unreachable, so this needs a runtime differential before any fix).

## The suspected gap
The async param-copy path `emit_param_copy_step` (`meld-core/src/adapter/fact.rs` ~8151) appears to do only a **flat bulk `memory.copy`** of a pointer-bearing aggregate param (e.g. `list<string>`), reading `element_size` for the byte count, and **never walks the inner pointers** — unlike the SYNC memory-copy adapter, which calls `emit_inner_pointer_fixup` (~5445) to deep-copy + re-point inner `(ptr,len)` buffers, and unlike the async RESULT path, which calls `emit_patch_nested_indirections` (~2858).

If that reading is correct, then for a `list<string>` (or `record{…string…}`) PARAM crossing memory in an async-lift adapter, the inner string `(ptr,len)` headers are bulk-copied **verbatim** and still point into the **caller's** memory — so the callee, reading from its own memory at those pointers, gets garbage / OOB. This would be a **same-encoding** correctness gap (independent of cross-encoding); the #272 fail-loud guard only covers the cross-encoding case.

## Why it's only a candidate
- Reachability unverified: does the resolver/compose actually route a nested-pointer async-lift param through `emit_param_copy_step`'s flat-copy path, or is it handled/rejected elsewhere? (The #272 item-1 lesson: a statically-plausible path can be unreachable.)
- Needs a `compose → fuse → execute` differential: an async-lift callee taking a `list<string>` param across `MemoryStrategy::MultiMemory`, asserting the callee reads the correct strings (vs garbage). Note meld's async-lift e2e is kiln-gated, so this may need the same compensating controls (generate adapter + inspect for the absence of an inner-pointer walk).

## Relationship to #272
#272 inc-5c (nested-string async PARAM transcoding) will need to BUILD this nested param-copy machinery on the async path regardless — so fixing this gap and implementing 5c overlap. If confirmed, the same-encoding fix should land first (or together).

## Next step
Confirm or refute with a runtime/inspection differential before any fix. If refuted (the path isn't reached, or nested param copy happens elsewhere), close.

avrabe:
## What
The foundational nested-PARAM slice — and it **closes #281** (a confirmed pre-existing correctness bug). Delivers nested `list<string>` PARAM transcoding for utf8→utf16, plus the same-encoding deep-copy that #281 was missing.

## #281 confirmed + fixed
Verified by inspection: pre-fix, a `list<string>` param had exactly **1** `memory.copy` (the outer element array) — the inner string buffers were bulk-copied verbatim and left **dangling into caller memory** (the callee would read caller memory cross-memory = garbage). The async param path simply had no nested-copy machinery (unlike the result/sync paths).

## How
Builds the missing machinery, mirroring the result-side `emit_patch_nested_indirections` but **param direction caller→callee**:
- Adds `param_wit_types` to `AdapterRequirements` (resolver) — mirror of `result_wit_types` — so the guard + emitter get `is_string` via `collect_indirections` (no descriptor change).
- New `emit_param_nested_indirections`, slotted into `emit_param_copy_step` after the outer bulk copy. Per inner element: `is_string` + Utf8→Utf16 → realloc+transcode into **callee** memory (src=caller where the inner ptr points, dst=callee) + store new ptr/len; else (`list<u8>` or same-encoding) → **deep-copy verbatim** caller→callee + re-point.
- The else branch **closes #281**: same-encoding `list<string>` params now deep-copy inner buffers (2 `memory.copy`) instead of dangling. `list<u8>` deep-copied verbatim, NOT transcoded.

**Budget:** param-nested walk (offsets 10..14) + transcode (21..25) disjoint, top 25 < 27 / 16 < 18, no growth (verified on real adapters).

**Guard:** `param_nested_string_ok = Utf8→Utf16 && param_inner_all_strings`; other directions / nested `list<u8>` / records still fail loud.

## Verification
- Step-0 premise test (confirms + pins the #281 close: 1 → 2 `memory.copy`).
- Runtime: transcode (["Hi","yo"]→409), non-ASCII len rewrite, `list<u8>` deep-copied verbatim (493).
- Integration local-bounds (both variants), guard scope tests.
- lib 410/0; `async_cross_encoding` 34/0; `adapter_safety` 23/0; **full suite green** (the same-encoding behavior change doesn't regress sync/wit_bindgen); LS-N gate 57/0/0.

**Reviewer focus:** the #281 deep-copy correctness (inner buffers re-pointed into callee memory), the caller→callee direction (opposite the result side), the `list<u8>`-deep-copied-not-transcoded guarantee, the budget, and the same-encoding behavior change.

## Tier-5
`adapter/fact.rs` + `resolver.rs` → Mythos delta-pass (posted separately).

Closes #281.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -3091,6 +3091,309 @@ fn build_nested_list_result_test_module(
     module.finish()
 }
 
+/// Build a self-contained two-memory wasm module that exercises the EXACT
+/// async-path NESTED-PARAM inner copy emitter ([`emit_param_nested_indirections`])
+/// for a `list<string>` PARAM crossing memory in the caller=Utf8 → callee=Utf16
+/// direction, for the #272 inc-5c-a runtime-differential oracle.
+///
+/// This is the PARAM direction (caller PROVIDES, callee READS): SRC = caller
+/// memory 0 (where the caller wrote the UTF-8 inner strings), DST = callee
+/// memory 1 (where the callee reads UTF-16, and where the deep-copy/transcode
+/// arena lives). The host pre-writes the post-outer-bulk-copy state:
+/// * caller memory 0: the inner UTF-8 (or `list<u8>`) source buffers.
+/// * callee memory 1: the SHALLOW outer copy of the 8-byte `(inner_ptr,
+///   inner_len)` records — `inner_ptr` STILL points into CALLER memory 0 (the
+///   bulk copy was shallow; gap #281).
+///
+/// The exported `patch_and_sum(outer_callee, count) -> i32` runs the production
+/// [`emit_param_nested_indirections`] (caller=Utf8 → callee=Utf16), which for
+/// EACH record transcodes (string) or deep-copies (`list<u8>`) the inner buffer
+/// into a freshly `cabi_realloc`'d CALLEE-memory buffer, rewrites the
+/// callee-side inner `(ptr, len)` header, then SUMS the inner units of every
+/// inner buffer by re-reading the PATCHED `(ptr, len)` headers from CALLEE
+/// memory. For the string case a passing UTF-16 sum proves real transcode +
+/// header-len rewrite + re-point into callee memory; for the `list<u8>` case a
+/// matching byte-sum proves verbatim deep-copy + re-point (the #281 close), NOT
+/// a transcode.
+///
+/// `#[doc(hidden)] pub` purely so the `async_cross_encoding` runtime test can
+/// instantiate it; not a supported API.
+#[doc(hidden)]
+pub fn build_nested_list_string_utf8_to_utf16_param_test_module() -> Vec<u8> {
+    use crate::parser::ComponentValType;
+    // Element type of the outer `list<string>` is `string` (is_string == true) —
+    // transcoded UTF-8 → UTF-16; the summing reads u16 code units.
+    build_nested_list_param_test_module(ComponentValType::String, true)
+}
+
+/// Build the #272 inc-5c-a PARAM negative oracle: the SAME nested-param emitter,
+/// but for a `list<list<u8>>` PARAM in the caller=Utf8 → callee=Utf16 direction.
+/// A nested `list<u8>` (`is_string == false`) is arbitrary binary; it MUST be
+/// DEEP-COPIED VERBATIM into callee memory and re-pointed (closing #281), NEVER
+/// transcoded. The module sums the inner BYTES via the PATCHED header; a
+/// verbatim copy preserves both the bytes and the length.
+///
+/// `#[doc(hidden)] pub` purely so the `async_cross_encoding` runtime test can
+/// instantiate it; not a supported API.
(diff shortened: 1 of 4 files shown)


avrabe:
## Mythos disposition — found-and-fixed index-alignment bug; re-verified

**Pass 1 found a confirmed bug** (failing `#[test]` + Kani): `param_wit_types` was populated per-WIT-param but the emitter indexed it per-pointer-pair — for a `(u32, list<string>)` param the spaces diverge, so `emit_param_nested_indirections` was silently skipped while the guard still allowed the site → inner strings dangle into caller memory (re-opens #281) + transcode skipped (H-4.4). Shipped single-param tests missed it.

**Fixed** (commit `356c8b5`): `param_wit_types` is now per-pointer-pair via `collect_param_pointer_wit_types`, which walks types EXACTLY like `collect_type_copy_layouts`; both resolver sites `debug_assert_eq!` the three vectors' lengths. Regression tests `inc5c_a_mixed_param_nested_list_string_after_scalar_*` fail pre-fix (89 vs 525 ops / 1 vs 2 copies), pass post-fix.

**Pass 2 (tight re-confirmation, walker-equivalence) — NO PROVABLE FINDING:**
- **Walker order/count equivalence:** `collect_param_pointer_wit_types` (resolver ~821), `collect_type_copy_layouts` (~754), and the parser's `collect_pointer_positions` (~2247) are arm-by-arm identical — same count + order for every `ComponentValType`: String→1; List(_)→1 (no recursion into T); Record→fields in order; Tuple→elems; FixedSizeList<_,N>→N; Type(idx)→same alias unwrap; Primitive/Flags/Variant/Option/Result/Own/Borrow→no pair. Verified by unit probes over tuple/record/fixedlist/mixed/nested shapes AND a Kani proof over bounded symbolic type trees (with teeth — fails when a recursion arm is dropped). The `debug_assert_eq!` invariants hold by construction.
- **Guard vs emitter (mixed/nested allowed shapes):** no allow-but-skip; both read `param_wit_types` at the same aligned `pair_idx`. `list<string>`→allowed+transcoded; `record{string,list<u8>}`→correctly rejected (`.all(is_string)` false); `list<record{string,string}>`→allowed; `record{string,u32}`→u32 bears no pair, shifts nothing.
- **Option/Result/Variant:** all walkers + `collect_indirections` skip them → zero pointer pairs → fail-loud by absence, no divergence.

Disposition: **NO PROVABLE FINDING** on the fixed branch; the index-alignment invariant is structurally guaranteed (three aligned walkers) + Kani-proven + regression-tested. Closes #281.

PR #285, opened 2026-06-15
feat(adapter): async nested list<string> PARAM all 6 directions + fix same-encoding deep-copy under-copy (#272 inc 5c-b)

avrabe:
## What
Completes nested-PARAM transcoding (all 6 directions), AND fixes a confirmed pre-existing #281-completeness bug the adversarial pass found in the shared deep-copy branch.

With this, **all 6 directions transcode for flat AND one-level-nested strings, params AND results** — the full practical matrix. Deeper list nesting (`list<list<string>>`) stays fail-loud.

## inc-5c-b (the 6-way nested-param dispatch) — clean
Extends `emit_param_nested_indirections` to all 6 directions (caller→callee), reusing the verified loop fns; dest-latin1 writes the tagged inner-header length; 6-scratch budget proven disjoint (callback 26<27, stackful 17<18, no growth); index alignment holds in the new directions (mixed `(scalar, list<string>)` regression test). The Mythos pass cleared all of these (hazards 2–5).

## The bug it found + fixed (hazard 1)
The same-encoding/`list<u8>` deep-copy branch used `l_buf_len * sub_elem_size`, and `collect_indirections` hardcodes `sub_elem_size=1` for a string. So a **same-encoding UTF-16** nested-param string was under-copied (`len*1` vs `len*2`) — confirmed by runtime: `"é"` (`[0xE9,0x00]`) read back as `[0xE9,0x48]` (trailing byte dropped, `0x48` bled from the adjacent buffer); CompactUtf16 tag-set tripped the bounds guard → dangling. Reachable since same-encoding params pass the cross-encoding guard.

**Fix** (mirrors the result-side `emit_patch_nested_indirections`): for a same-encoding inner *string*, the deep-copy byte count uses the encoding width — CompactUtf16 → `emit_copy_byte_count(.., true)` (tag-aware, also un-inflates the guard) align 2; Utf16 → `len*2` align 2; Utf8 → `len*1`. `list<u8>`/`list<uN>` (`is_string=false`) → `len*sub_elem_size`, unchanged. Applied to bounds guard + realloc + memory.copy consistently.

## Verification
- inc-5c-b: 5 per-direction runtime + guard + budget + mixed-param-alignment tests.
- Deep-copy fix: `inc5c_b_same_encoding_{utf16,compact_utf16_tagset}_nested_param_*` (fail pre-fix, pass post-fix) + utf8/`list<u8>` unchanged-behavior tests.
- async_cross_encoding 44/0; lib 420/0; adapter_safety 23/0; full suite green; LS-N gate 57/0/0; clippy/fmt clean.

## Tier-5
`adapter/fact.rs` → Mythos delta-pass (disposition below — hazard 1 found-and-fixed, 2–5 clean).

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -3146,6 +3146,148 @@ pub fn build_nested_list_u8_param_deep_copied_test_module() -> Vec<u8> {
     )
 }
 
+/// Build a self-contained two-memory wasm module that exercises the production
+/// nested-PARAM emitter ([`emit_param_nested_indirections`]) for a
+/// `list<string>` PARAM in an ARBITRARY param direction (`caller_enc →
+/// callee_enc`), for the #272 inc-5c-b runtime-differential oracle covering the
+/// five non-utf8→utf16 directions — the PARAM mirror of
+/// [`build_nested_list_string_result_test_module`].
+///
+/// PARAM direction (caller PROVIDES, callee READS): SRC = CALLER memory 0 (where
+/// the host wrote the inner SOURCE strings), DST = CALLEE memory 1 (where the
+/// transcoded buffers + the patched records live, and where `cabi_realloc`
+/// allocates). The outer `(ptr, len)` record array lives ONLY in CALLEE memory 1
+/// (the post-outer-bulk-copy duplicate); its inner ptrs still point into CALLER
+/// memory 0 (the shallow bulk copy — gap #281).
+///
+/// It does NOT sum in wasm (the dest-latin1 tag + per-direction read width make
+/// a wasm sum awkward — same choice as the inc-5b result oracle). It runs the
+/// patcher, then exports `callee_memory` so the host reads back the PATCHED inner
+/// `(ptr, len)` records + transcoded inner buffers (both in CALLEE memory) and
+/// asserts directly in Rust. For dest-latin1 directions the inner len header is
+/// rewritten to the TAGGED output length.
+///
+/// `#[doc(hidden)] pub` purely so the `async_cross_encoding` runtime test can
+/// instantiate it; not a supported API.
+#[doc(hidden)]
+pub fn build_nested_list_string_param_test_module(
+    caller_enc: crate::parser::CanonStringEncoding,
+    callee_enc: crate::parser::CanonStringEncoding,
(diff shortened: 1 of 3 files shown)


avrabe:
## Mythos delta-pass disposition — found-and-fixed (hazard 1); 5c-b additions clean

The adversarial pass cleared the inc-5c-b additions (hazards 2–5) and FOUND a confirmed pre-existing bug (hazard 1) in the shared deep-copy branch, now fixed.

**Hazard 1 — same-encoding nested-param string deep-copy UNDER-COPY (CONFIRMED → FIXED).** Both oracles failed pre-fix: a Kani proof (`len*1` vs the true `len*2` for utf16) and a wasmtime runtime test — same-encoding (Utf16,Utf16) `list<string>` `["é","Hi"]` read `"é"` back as `[0xE9,0x48]` (trailing `0x00` dropped, `0x48` bled from the adjacent source); CompactUtf16 tag-set tripped the inflated-len bounds guard → re-point skipped → dangling. Reachable: same-encoding params pass the cross-encoding guard early. **Fixed** (commit `0bce074`) by making the deep-copy byte count encoding-aware for same-encoding strings (mirror of the result-side `emit_patch_nested_indirections`): CompactUtf16 → `emit_copy_byte_count(.., true)` align 2; Utf16 → `len*2` align 2; Utf8 → `len*1`; `list<u8>`/`list<uN>` → `len*sub_elem_size` unchanged. Regression tests fail pre-fix / pass post-fix; utf8 + list<u8> paths confirmed byte-identical.

**Hazards 2–5 (the inc-5c-b diff) — clean:**
- dest-latin1 tagged inner-header length per-element correct (mixed list);
- 6-scratch budget disjoint under simultaneous liveness (callback 26<27, stackful 17<18), real-adapter validated;
- all 6 direction→loop-fn/realloc/MemArg mappings correct (src=caller/dst=callee), no swap;
- index alignment holds in the new directions (`(scalar, list<string>)` Utf16→Utf8 fires the walk + guard agrees).

Disposition: **NO PROVABLE FINDING remaining** — hazard 1 found, fixed, and regression-tested; the 5c-b additions verified clean. Completes all 6 nested-param directions + closes the same-encoding deep-copy under-copy.

Issue #286, opened 2026-06-15
#272 follow-up: deeper list nesting (list<list<string>>) — needs a recursive-copy restructure (scoped)


avrabe:
Split off from #272 (now practically complete). The ONE remaining cross-encoding-transcoding gap: strings two-or-more list-levels deep (`list<list<string>>`). **Currently fail-loud (SAFE)** — the guard's `result_inner_all_strings`/`param_inner_all_strings` call `collect_indirections` on the outer element `list<string>`, which yields a single `is_string=false` leaf (the inner list pointer), so `.all(is_string)` is false → not allowed. The allow-set exactly equals the depth-1 transcode-set; this is correct, not a bug.

A scope-first investigation (5d) found this is a **structural restructure, not a bounded increment** — three independent changes, each substantial:

## (a) `collect_indirections` descriptor redesign — ~9 sites, 3 files
`collect_indirections` (fact.rs ~4616) returns `Vec<(offset, sub_elem_size, is_string)>` and treats `CVT::List(elem)` as ONE `is_string=false` leaf without recursing into `elem` — the flattened tuple discards which CVT subtree produced each indirection, so the emitter can't know an inner indirection is a pointer-bearing list needing recursion. Fix: a type-carrying (recursive or 4-tuple) descriptor. Blast radius: ~9 destructuring sites across fact.rs (emitters ×2, guard predicates ×4) + `lib.rs:1220` (the async result shim-stabilization deep-copy, a separate consumer).

## (b) Budget restructure
Depth-2 needs a 2nd nested loop's 6 locals live simultaneously with depth-1's 6 + the leaf transcode's 5–6. Free slots today = 2 of 6 → **stackful 18→~24, callback 27→~33**, and every `inc*_{callback,stackful}_*_within_budget` invariant re-derived. Worse, depth-1 reuses `l_new_ptr` as the leaf-transcode out-ptr sink; at depth-2 that local holds the inner-array base for the whole loop → can't double as the sink → forces a re-layout, not an append.

## (c) Recursion shape
The honest fix is general depth-N recursion parameterized over `(elem_ty, scratch_base, depth)` — the emitter is a single ~410-line fn. A depth-2-only inline copy that mis-shares `l_new_ptr` is exactly the fragile-recursive-copy-that-corrupts to avoid (worse than the current fail-loud).

## Recommended split
[...]

Issue #168, opened 2026-05-19
smithy: sanitizer/libc config drift recurrence on fuzz smoke jobs


avrabe:
## Summary

The "sanitizer is incompatible with statically linked libc" error originally documented in #139 §3 continues to surface intermittently on fuzz smoke jobs. Affects `pulseengine-ci-01-5` and possibly siblings; `-7` was previously identified as the only consistently-working runner.

## Observed (today, 2026-05-19)

PR #166 first run failed BOTH `fuzz_parse_component` AND `fuzz_resolver_terminates` with:

```
error: sanitizer is incompatible with statically linked libc, disable it using `-C target-feature=-crt-static`
error: could not compile `equivalent` (lib) due to 1 previous error
...
```

`gh run rerun --failed` re-queued both jobs. They sat queued ~25 min waiting for the rust-cpu pool (7/7 busy) to free up so they could land on a different runner.

## Pattern recurrence

| Date | PR | Symptom |
|---|---|---|
| 2026-05-03 to 2026-05-11 | v0.7.0 cycle | Original observation (#139 §3) |
| 2026-05-16 (yesterday's session) | #153 (PR #153 rerun) | Same error, cleared on rerun |
| 2026-05-19 | #166 | Both fuzz_parse and fuzz_resolver fail simultaneously |

This isn't new, but it's recurring frequently enough that meld's CI is unreliable without per-PR `gh run rerun --failed` intervention. #139's "Three consecutive PRs merge without --admin bypass" acceptance bar keeps getting reset.

## Asks

1. **Standardize the rust-cpu runner image** to eliminate sanitizer/libc per-runner config drift. The working `-7` runner should be the template; `-5`/`-6` (and siblings if affected) should be re-imaged.
2. **Add a sanity-check smoke test** that runs at runner boot: `cargo +nightly fuzz build` against a known-good corpus on each rust-cpu runner. Fail health-check and remove from pool if it errors on sanitizer/libc.

## References

- #139 §3 documented the original observation
- PR #166 fuzz run today: https://github.com/pulseengine/meld/actions/runs/26076704174

## Repro

Push any commit that triggers `.github/workflows/fuzz.yml`; ~20-30% chance a fuzz_parse_component or fuzz_resolver_terminates job lands on a config-drifted runner.

avrabe:
Investigated the runner side and the drift theory doesn't hold up.

Compared runner5/6/7 state — all three have **identical** toolchains
[...]

PR #290, opened 2026-06-15
refactor(adapter): #286 5d-pre — type-carrying Indirection descriptor (no behavior change)

avrabe:
## What & why

`collect_indirections` returned a flat `Vec<(offset, sub_elem_size, is_string)>` that treated `CVT::List(elem)` as **one** `is_string=false` leaf and **discarded the element type**. That discard is the structural blocker for depth-N nested-string transcoding (#286 5d — `list<list<string>>`): the emitter can't recurse into an inner pointer-bearing list if the descriptor never told it the list's element type.

This is **5d-pre**: a behaviour-preserving descriptor redesign that carries the type, setting up 5d cleanly.

## Change

```rust
struct Indirection { offset, sub_elem_size, kind }
enum IndirectionKind { String, List { elem: Box<ComponentValType> } }  // elem = the 5d recursion seam
Indirection::is_string() == matches!(kind, String)                     // the former flag
```

All consumer sites moved to field access (`offset`/`sub_elem_size` kept as references so `*offset`/`*sub_elem_size` body reads are untouched):
- `emit_patch_nested_indirections` (result), `emit_param_nested_indirections` (param)
- the 4 guard allow-predicate closures → `.all(|ind| ind.is_string())`
- `lib.rs` async result shim-stabilization deep-copy + its `&[Indirection]` param

## No behaviour change (the 5d-pre invariant)

Both `list<list<string>>` and `list<list<u8>>` still collect to a single `List` leaf with `is_string() == false`, so the guard's `.all(is_string)` predicate still **rejects** them → they stay **fail-loud (SAFE)**, exactly as before. 5d-pre only *adds* the carried `elem`; it does not widen the allow-set. `elem` is `#[allow(dead_code)]` until 5d consumes it.

## Verification

- New unit test `collect_indirections_carries_nested_list_element_type_5d_pre` pins the descriptor (`list<string>` vs `list<u8>`, depth-1 vs depth-2) and **reads the carried `elem`**, proving the type is genuinely carried.
- Full `meld-core` suite green — incl. **44** `async_cross_encoding` + **23** `adapter_safety` nested/SR-17 oracles (the behaviour-preservation oracle).
- `clippy --all-targets -D warnings` clean; `fmt` clean; LS-N gate **57/0/0**.

## Scope

Tier-5 (`adapter/fact.rs`, `lib.rs`) → Mythos clean-room delta-pass + `mythos-pass-done` label before merge.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -3873,7 +3873,13 @@ fn emit_patch_nested_indirections(
     body.instruction(&Instruction::I32Add);
     body.instruction(&Instruction::LocalSet(l_rec_src));
 
-    for (offset, sub_elem_size, is_string) in &indirections {
+    for ind in &indirections {
+        // #286 5d-pre: field-access over the type-carrying descriptor. `offset`
+        // / `sub_elem_size` stay references so the `*offset`/`*sub_elem_size`
+        // body reads are unchanged; `is_string` is the former leaf flag.
+        let offset = &ind.offset;
+        let sub_elem_size = &ind.sub_elem_size;
+        let is_string = ind.is_string();
         let dst_mem_arg_ptr = wasm_encoder::MemArg {
             offset: *offset as u64,
             align: 2,
@@ -3910,7 +3916,7 @@ fn emit_patch_nested_indirections(
         // `sub_elem_size == 1` but is arbitrary binary — explicitly does NOT
         // take this branch; it keeps the raw-copy below so its bytes are never
         // mis-transcoded.
-        let transcode_inner = *is_string && inner_transcode_dir;
+        let transcode_inner = is_string && inner_transcode_dir;
 
         if transcode_inner {
             // RESULT direction (callee PRODUCES, caller READS): SRC = callee
@@ -4303,7 +4309,12 @@ fn emit_param_nested_indirections(
     body.instruction(&Instruction::I32Add);
     body.instruction(&Instruction::LocalSet(l_rec));
 
-    for (offset, sub_elem_size, is_string) in &indirections {
+    for ind in &indirections {
+        // #286 5d-pre: field-access over the type-carrying descriptor (see the
+        // result-side mirror). `offset`/`sub_elem_size` stay references.
+        let offset = &ind.offset;
+        let sub_elem_size = &ind.sub_elem_size;
+        let is_string = ind.is_string();
         // Both the inner-header read and write are in CALLEE memory (the
         // record array was bulk-copied there); the inner BUFFER itself is read
         // from CALLER memory (the shallow ptr) and written to CALLEE memory.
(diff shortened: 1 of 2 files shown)


avrabe:
## Status — pre-merge gating

**Required CI**: Clippy / Format / LS-N (57/0/0) / Coverage / Detect-Tier-5 → all green.

**Non-blocking**:
- *Fuzz smoke* (4 jobs) fail with `ToolNotFound: x86_64-linux-musl-g++` — the known #168 musl/sccache infra flake on the runner, not a code failure and not a required gate.
- *Mythos delta-pass gate* fails pre-label as expected; it goes green once the clean-room Tier-5 pass completes and the `mythos-pass-done` label fires the `labeled` webhook.

**Merge is gated on the Mythos clean-room delta-pass** (Tier-5 discipline). That adversarial pass is queued; the model API is transiently overloaded (529), so it will run on the next attempt. Inline pre-audit (compiler-verified complete under `-D warnings`; the 44 `async_cross_encoding` + 23 `adapter_safety` runtime oracles covering the exact touched emitter/guard paths all pass) supports behaviour-preservation, but the fresh-context adversarial pass is still required before the label is applied.

Not merging until the clean-room pass returns REFUTED.

avrabe:
## Mythos clean-room delta-pass — VERDICT: REFUTED (behaviour-preserving)

A fresh-context adversarial verifier attacked the claim that the tuple→struct refactor of `collect_indirections` changes no emitted adapter bytes or guard decision, across all four surfaces (guard closures, both emitter loops, the lib.rs shim, and the `collect_indirections` arms). It constructed a **differential oracle**: a verbatim reimplementation of the pre-559d4be tuple logic (`collect_indirections_OLD`), asserting `old(ty, base) == new(ty, base)` leaf-for-leaf.

**Evidence (run to completion):**
- **Concrete differential test** — `old == new` (offset, sub_elem_size, is_string) over `string`, primitives, `list<string>`, `list<u8>`, `list<list<string>>`, `list<list<u8>>`, records, tuples, nested record/tuple, and the dropped `FixedSizeList`/`Option`/`Variant` arms, at bases {0,4,16,1000}. **PASS.**
- **Existing runtime oracles** — 44 `async_cross_encoding` + 23 `adapter_safety` (real adapters fused + executed under wasmtime, exercising the exact emitter/guard paths). **PASS.**
- **Compiler completeness** — clean build under `-D warnings` proves no consumer site was left on the old tuple shape.

**Symbolic half — disclosed limitation:** the two `#[cfg(kani)]` equivalence proofs are **intractable** under CBMC — they time out unwinding the recursive, allocation-heavy `cabi_size_align` over a symbolic `ComponentValType` tree (`Box`/`Vec`/`Option` allocator models). This is the same CBMC-on-allocating-code limit already recorded for SR-40 (awaits an allocation-free layout-core refactor), **not** a counterexample: CBMC produces no failing check, it simply cannot decide. Reported honestly rather than claimed as a passing proof.

**No finding** (no failing test, no failing proof). The verification scaffolding was scratch and is not committed; the permanent guard is the committed `collect_indirections_carries_nested_list_element_type_5d_pre` unit test.

PR #291, opened 2026-06-17
feat(adapter): #286 5d — depth-2 list<list<string>> RESULT cross-encoding transcoding

avrabe:
## What

Closes the last #272 cross-encoding gap on the RESULT side: depth-2 nested strings (`list<list<string>>`) now transcode across encodings, instead of failing loud. Builds on 5d-pre (#290, the type-carrying `Indirection` descriptor).

## How (three verified increments)

1. **Codegen-time recursion** (3ada8da) in `emit_patch_nested_indirections`: a `List { elem }` leaf with an all-string subtree, in a transcodable direction, copies the inner array into caller memory + recurses one level (reading source headers from the callee original, writing patched headers to the caller copy). Each depth `d` uses a disjoint 6-local block at `l_first_scratch + 6·depth`; the transcode scratch is **shared** because only the deepest leaf transcodes at any instant — so each depth's `l_new_ptr` is its own local and never collides with the leaf sink (the trap #286 warned about, avoided structurally). Depth-1 byte-identical.

2. **Depth-2 runtime oracle** (d06b22f): `d5d_depth2_...utf8_to_utf16_transcodes` drives the real emitter over `list<list<string>>`, follows the doubly relocated pointers in caller memory, asserts each deepest string transcoded UTF-8→UTF-16. Validator-valid + runtime-correct.

3. **Production-active** (7c2b510): guard widened via `indirection_result_transcodable_depth2` (allow-set ≡ emitter transcode-set); async-adapter budgets grown (callback 27→28, stackful 18→23) with the result `transcode_base` shifted past the depth-1 recursion block.

## Verification

- **Mandatory real-adapter integration tests** (the budget-bug class — unknown-local overflow AND valid-but-wrong local-reuse collision — that the synthetic oracle can't reach): `d5d_{callback,stackful}_adapter_depth2_nested_result_within_budget` generate the REAL async adapters and assert they validate within the grown budgets.
- **Allow-set ≡ transcode-set**: `d5d_async_nested_list_list_string_result_allowed` (depth-2 ALLOWED) + `d5d_async_nested_list_list_u8_result_fail_loud` + `d5d_async_nested_depth3_result_fail_loud` (binary `list<u8>` and depth-3 stay FAIL-LOUD — the emitter recurses exactly one level; budget sized for depth-2).
- Full `meld-core` suite 0 failed; clippy `-D warnings` + fmt clean.

## Scope

Tier-5 (`adapter/fact.rs`) → Mythos clean-room delta-pass + `mythos-pass-done` label before merge. PARAM-side depth-2 and result directions beyond utf8↔utf16/latin1 (already wired via the shared leaf dispatch) are covered; param-side depth-2 remains fail-loud (future increment).

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -4896,6 +4896,26 @@ impl Indirection {
     }
 }
 
+/// #286 5d: the per-indirection RESULT recurse/transcode decision, shared by the
+/// emitter and the result guard so the guard's allow-set is EXACTLY what
+/// `emit_patch_nested_indirections` transcodes or recurses into (allow-set ≡
+/// transcode-set). A leaf is depth-bounded transcodable iff it is a `string`
+/// (transcoded directly) OR a nested `list<elem>` whose immediate inner
+/// indirections are all strings (the depth-2 `list<list<string>>` case the
+/// emitter recurses one level for). Returns `false` for a nested `list<u8>`
+/// (raw-copied binary) and for deeper-than-2 nesting (whose inner subtree is a
+/// `list`, not all strings) — both remain fail-loud, matching the emitter's
+/// recursion guard and the async-adapter local budget (sized for depth-2).
+pub(crate) fn indirection_result_transcodable_depth2(ind: &Indirection) -> bool {
+    match &ind.kind {
+        IndirectionKind::String => true,
+        IndirectionKind::List { elem } => {
+            let sub = collect_indirections(elem, 0);
+            !sub.is_empty() && sub.iter().all(|s| s.is_string())
+        }
+    }
+}
+
 /// For a given element type, find every field offset that holds a (ptr, len)
 /// pair that needs cross-memory copying. See [`Indirection`].
 pub(crate) fn collect_indirections(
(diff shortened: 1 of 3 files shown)


avrabe:
## Mythos clean-room delta-pass — VERDICT: REFUTED

A fresh-context adversarial verifier attacked the three gaps the in-PR tests don't cover, extending the production-driven depth-2 oracle (`build_nested_list_list_string_result_test_module`) with throwaway probes (reverted; tree clean):

1. **Other directions at depth-2** — utf16→utf8 and latin1→utf16 (CompactUtf16) both transcode the deepest strings correctly, pointers relocated (≥2048), output lengths correct. The shared leaf dispatch works under recursion.
2. **Runtime aliasing** — `[["Hi","é"],["ok"]]` and `[["Hi","é"],[],["é","ok"]]` (empty inner list sandwiched between multibyte-bearing lists): no cross-element corruption; the depth-0 `l_new_ptr`/`l_rec_src` survive the inner recursion (disjoint +6 block + shared transcode scratch hold).
3. **Bulk-copy + recurse** — inner length headers carry element COUNT (not byte count) in every case; empty inner list (count 0) does not trap (zero-length realloc/copy + zero-iteration recursive loop well-formed).

No failing runtime test or Validator rejection could be constructed → **REFUTED** (no finding). Combined with the in-PR real-adapter budget-validation tests and the depth-2 runtime oracle, the depth-2 RESULT recursion is verified valid + correct.

PR #292, opened 2026-06-17
feat(adapter): #286 5d PARAM — depth-2 list<list<string>> PARAM cross-encoding transcoding (closes #286)

### LINKED ISSUE #286: #272 follow-up: deeper list nesting (list<list<string>>) — needs a recursive-copy restructure (scoped)

avrabe:
## What

The PARAM-side mirror of the merged result-side depth-2 work (#291), **closing #286**: depth-2 nested strings (`list<list<string>>`) now transcode across encodings as PARAMS too, instead of failing loud.

## How (three verified increments)

1. **Codegen-time recursion** (9cad68f) in `emit_param_nested_indirections`: a `List { elem }` leaf with an all-string subtree deep-copies the inner array caller→callee, stores the relocated header, and recurses one level to patch the inner array **in-place** (transcoding each string caller→callee). Param-specific: 5-local-per-depth blocks (`l_first_scratch + 5·d`), inverted memory direction. Depth-1 byte-identical.

2. **Depth-2 runtime oracle** (c848cad): `d5d_depth2_..._param_utf8_to_utf16_transcodes` — full two-level value in caller memory, shallow outer in callee; after `patch`, follows the doubly relocated pointers within callee memory, asserting each deepest string transcoded. Validator-valid + runtime-correct.

3. **Production-active** (2662daa): param guard widened via the now-shared, direction-agnostic `indirection_transcodable_depth2` (allow-set ≡ transcode-set). **No budget growth needed** — proven by the real-adapter validator tests: the result-side budget growth (callback 28, stackful 23) + the existing param `transcode_base` headroom already accommodate the depth-1 param recursion block.

## Verification

- **Real-adapter integration tests**: `d5d_{callback,stackful}_adapter_depth2_nested_param_within_budget` generate the REAL async adapters and validate (the budget-bug class the synthetic oracle can't reach — confirming the no-growth claim).
- **Allow-set ≡ transcode-set**: `d5d_async_nested_list_list_string_param_allowed` + `d5d_async_deeper_list_nesting_param_allowed_all_directions` (all 6 directions) + `d5d_async_nested_list_list_u8_param_fail_loud` + `d5d_async_nested_depth3_param_fail_loud`. The stale `inc5c_b_..._param_fails_loud_all_directions` (old fail-loud invariant) was repurposed to assert the new ALLOWED behavior.
- Full `meld-core` suite 0 failed; clippy `-D warnings` + fmt clean.

## Scope

Tier-5 (`adapter/fact.rs`) → Mythos clean-room delta-pass + `mythos-pass-done` label before merge. With this, both RESULT and PARAM depth-2 `list<list<string>>` cross-encoding transcode across all directions — **#286 fully closed**. Depth-3+ stays fail-loud (emitter recurses one level; budget sized for depth-2).

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -5116,17 +5116,20 @@ impl Indirection {
     }
 }
 
-/// #286 5d: the per-indirection RESULT recurse/transcode decision, shared by the
-/// emitter and the result guard so the guard's allow-set is EXACTLY what
-/// `emit_patch_nested_indirections` transcodes or recurses into (allow-set ≡
-/// transcode-set). A leaf is depth-bounded transcodable iff it is a `string`
-/// (transcoded directly) OR a nested `list<elem>` whose immediate inner
-/// indirections are all strings (the depth-2 `list<list<string>>` case the
-/// emitter recurses one level for). Returns `false` for a nested `list<u8>`
-/// (raw-copied binary) and for deeper-than-2 nesting (whose inner subtree is a
-/// `list`, not all strings) — both remain fail-loud, matching the emitter's
-/// recursion guard and the async-adapter local budget (sized for depth-2).
-pub(crate) fn indirection_result_transcodable_depth2(ind: &Indirection) -> bool {
+/// #286 5d: the per-indirection recurse/transcode decision, shared by BOTH the
+/// result and param emitters/guards so each guard's allow-set is EXACTLY what
+/// the emitter transcodes or recurses into (allow-set ≡ transcode-set). The
+/// decision is DIRECTION-AGNOSTIC — it inspects only the type structure, which
+/// is identical for `emit_patch_nested_indirections` (result) and
+/// `emit_param_nested_indirections` (param). A leaf is depth-bounded transcodable
+/// iff it is a `string` (transcoded directly) OR a nested `list<elem>` whose
+/// immediate inner indirections are all strings (the depth-2 `list<list<string>>`
+/// case the emitter recurses one level for). Returns `false` for a nested
+/// `list<u8>` (raw-copied binary) and for deeper-than-2 nesting (whose inner
+/// subtree is a `list`, not all strings) — both remain fail-loud, matching the
+/// emitter's recursion guard and the async-adapter local budget (sized for
+/// depth-2 on both sides).
+pub(crate) fn indirection_transcodable_depth2(ind: &Indirection) -> bool {
     match &ind.kind {
         IndirectionKind::String => true,
         IndirectionKind::List { elem } => {
@@ -11955,7 +11958,7 @@ impl FactStyleGenerator {
                 // makes the whole result NOT nested-string-allowed (fail-loud).
                 inner_indirections
                     .iter()
-                    .all(indirection_result_transcodable_depth2)
+                    .all(indirection_transcodable_depth2)
             })
             // At least one result must actually carry a nested string for this
             // predicate to widen the allow-set (a pure `list<u32>` has no inner
@@ -11966,7 +11969,7 @@ impl FactStyleGenerator {
                 | crate::parser::ComponentValType::FixedSizeList(elem, _) => {
                     let inner = collect_indirections(elem, 0);
                     !inner.is_empty()
-                        && inner.iter().all(indirection_result_transcodable_depth2)
+                        && inner.iter().all(indirection_transcodable_depth2)
                 }
                 _ => false,
             });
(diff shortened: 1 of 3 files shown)


avrabe:
## Mythos clean-room delta-pass — VERDICT: REFUTED

A fresh-context adversarial verifier attacked the three gaps the in-PR tests don't cover (throwaway probes, reverted; tree clean):

1. **Other directions at depth-2 PARAM** — utf16→utf8 ("Hi"→48 69, "é"→C3 A9, "ok"→6F 6B) and latin1→utf8 (CompactUtf16 tag-clear "é" 0xE9 → C3 A9) both transcode correctly via the real `emit_param_nested_indirections`, inner-list + string pointers relocated ≥2048 into CALLEE.
2. **No-budget-growth claim holds** — verified the param region (step 0.5) and result region (step 3) are time-disjoint; param depth-2 tops at offset 26, result depth-2 at 27, both < budget 28 (callback) / fit 23 (stackful). The param `transcode_base` (l_p2+16) sits past the depth-1 param block (l_p2+10..14); the index overlap is benign time-multiplexed reuse, not a collision. The `d5d_*_within_budget` real-adapter validator tests confirm.
3. **In-place aliasing + empty inner list** — `[["a","b"],[],["c"]]` (UTF-8→UTF-16): empty inner list (count 0) doesn't trap; siblings uncorrupted; the depth-0 loop reads its locals fresh each iteration before recursing, so the in-place inner patch (block at l_first_scratch+5) can't clobber a live depth-0 value.

No failing runtime test or Validator rejection → **REFUTED** (no finding). Working tree clean (3 commits only).

Issue #295, opened 2026-06-17
resource_graph: two latent resource-attribution weaknesses (reachability unconfirmed) — Mythos sweep 2026-06-17


avrabe:
## Summary

A bounded Mythos discovery sweep of `meld-core/src/resource_graph.rs` (`ResourceGraph::build` + queries) surfaced **two structurally-real attribution weaknesses**, both **verified in code** but with **reachability UNCONFIRMED** — I could not construct the compose→fuse→execute differential that proves an observable H-class misroute, so per the validate.md bar this is **NOT a confirmed bug**. Filing to capture the lead for a dedicated, budgeted investigation that either confirms (→ fix) or refutes (→ document why safe). **Per-component `defines_resource` is correct**, which bounds severity (see below).

## Weakness 1 — `definer_cache` is `(iface, name)`-keyed → last-wins on textual collision

`resource_definer()` (resource_graph.rs:404-420) reads `definer_cache: Map<(String iface, String name), Option<usize>>` and returns a single `Option<usize>`. `build` inserts into it in a loop (resource_graph.rs:307), so two components that define the *same textual* `(iface, name)` — spec-legal, since the Component Model identifies resources **nominally** (by definition site), not textually — collapse **last-wins**: `resource_definer` silently returns only the last component, dropping the first.

- Probe (sweep): two comps each with their own `ResourceRep` + `[resource-rep]input-stream` from textually-identical `wasi:io/streams` → `defines_resource(0)=true` AND `defines_resource(1)=true` (per-component path **correct**), but `resource_definer(...) = Some(1)` (drops comp 0).

## Weakness 2 — terminal-exporter pass attributes a definer WITHOUT a rep/Defines check

The terminal-exporter second pass (resource_graph.rs:300-309): when `to_comp` is classified terminal (not a re-exporter for `import_name`), it inserts `defines_cache[(to_comp, iface, rn)] = true` and `definer_cache[(iface,rn)] = Some(to_comp)` for **every** matching resource node — **without verifying `to_comp` actually holds a canonical `ResourceRep`/`Defines` edge** for that resource. A component that merely exports an interface *name* and is the export-resolution target can thus be recorded as a definer.

- Probe (sweep): comp A imports `iface:i/mod` resolved to comp B (empty — no rep/new/imports); comp C holds the real `ResourceRep`. Result: `resource_definer = Some(B)` and `defines_resource(B)=true`, though B owns nothing and C is the true rep holder.

## The reachability question (what a follow-up must close)

[...]

PR #296, opened 2026-06-20
fix(resource_graph): #295 — rep-backed Defines authoritative over export-inferred terminal (LS-A-17 hardening)

avrabe:
## What

Closes the #295 follow-up: defensive hardening of `ResourceGraph::build`'s definer attribution. The terminal-exporter pass ran `defines_cache.insert` / `definer_cache.insert` **before** the rep-backed `defines_candidates` were merged with `.or_insert`, so an export-name terminal holding **no** `ResourceRep` could silently override a genuine rep-backed definer for the same `(iface, rn)`.

## Why it's hardening, not a correctness fix

Reachability to a live misroute was **REFUTED** (#295): the resolver consumes `resource_definer` only as `.is_some()`, and the routing override at `resolver.rs:3116` is downgrade-only + heuristic-gated, so a wrong definer value can't steer a route. This makes the resource queries robust against *future* consumers; it changes no current routing.

## Fix

Compute `rep_backed_resources` (the `(iface,rn)` set with a rep-backed `Defines` candidate) and skip export-inferred attribution for those — rep-backed `Defines` is authoritative; export-inference fills only the genuine no-rep (flattened-canonical) case it exists for. **Behaviour-unchanged** where the terminal exporter *is* the rep holder (the verified-correct LS-A-17/18 + live-routing cases); only the rep-vs-export conflict resolves toward the rep holder.

## Oracle (proven red→green)

`ls_a_17_rep_backed_definer_authoritative_over_export_terminal`: a rep holder (comp 0) + a no-rep export terminal (comp 1) for the same interface → asserts `resource_definer == Some(0)`, `defines_resource(0)=true`, `defines_resource(1)=false`. **Confirmed FAILING (`Some(1)`) on the pre-fix code**, passing after.

## Traceability & verification

- **LS-A-17** Fix section extended with the #295 hardening + a residual note (the no-rep-collision last-wins determinism, `.is_some()`-immune at every consumer). Honest framing: robustness, reachability-refuted — *not* a new loss path.
- resource_graph 4/0; full `meld-core` suite **0 failed**; clippy `-D warnings` + fmt clean; LS-N gate **57/0/0** (LS-A-17 already gated; the new `ls_a_17_` test adds coverage); `rivet validate` **0 broken cross-refs**.

## Scope

Tier-5 (`resource_graph.rs`) → Mythos clean-room delta-pass + `mythos-pass-done` label before merge.

### DIFF
--- a/meld-core/src/resource_graph.rs
+++ b/meld-core/src/resource_graph.rs
@@ -272,6 +272,22 @@ impl ResourceGraph {
             }
         }
 
+        // #295 (LS-A-17 hardening): the set of `(iface, rn)` resources that
+        // have a rep-backed `Defines` candidate (a canonical `ResourceRep`
+        // holder). The export-inferred terminal-exporter pass below runs with
+        // `.insert(...)` BEFORE the rep-backed candidates are merged in with
+        // `.or_insert(...)`, so without this guard an export-name terminal
+        // component holding NO rep could silently override a genuine rep-backed
+        // definer for the same resource. Rep-backed `Defines` is AUTHORITATIVE;
+        // export-inference must fill ONLY the genuine no-rep (flattened-
+        // canonical) case it exists for. Reachability to a live misroute was
+        // refuted (#295); this is robustness hardening, behaviour-unchanged
+        // where the terminal exporter IS the rep holder.
+        let rep_backed_resources: std::collections::HashSet<(String, String)> = defines_candidates
+            .keys()
+            .map(|(_idx, iface, rn)| (iface.clone(), rn.clone()))
+            .collect();
+
         // Second pass: use ResolvesTo edges for definer detection.
         // For each interface, find the terminal component — the one that
         // exports but doesn't import (no outgoing ResolvesTo for this interface).
@@ -302,7 +318,13 @@ impl ResourceGraph {
                 for (iface, rn) in resource_nodes.keys() {
                     let iface_matches = iface == import_name
                         || iface.strip_prefix("[export]") == Some(import_name.as_str());
-                    if iface_matches {
+                    // #295: skip export-inferred attribution for any resource
+                    // that already has a rep-backed `Defines` candidate — the
+                    // rep holder is the authoritative definer and is merged in
+                    // below. Only attribute the export terminal for the genuine
+                    // no-rep case.
+                    if iface_matches && !rep_backed_resources.contains(&(iface.clone(), rn.clone()))
+                    {
                         defines_cache.insert((*to_comp, iface.clone(), rn.clone()), true);
                         definer_cache.insert((iface.clone(), rn.clone()), Some(*to_comp));
                     }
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos clean-room delta-pass — VERDICT: REFUTED (hardening sound, behaviour-preserving)

A fresh-context adversarial verifier attacked the change (bounded; throwaway probes reverted, tree clean):

1. **Oracle honesty** — disabled ONLY the new guard (left the committed test) → `ls_a_17_rep_backed_definer_authoritative_over_export_terminal` fails with exactly `left: Some(1), right: Some(0)` (the claimed pre-fix value); the other `ls_a_17_*`/`ls_a_18_*` still pass. Genuinely red→green, non-tautological.
2. **Behaviour-preservation** — full `meld-core` suite passes unchanged post-fix (0 failed). No real resolver-produced topology changes its `resource_definer`/`defines_resource` answers.
3. **Cleanup-purge attack** (the dangerous case for this change) — built the exact input: a rep holder whose own same-interface import resolves to a peer, so the re-exporter cleanup purges its rep-backed entry and the new guard has already skipped export-inference → post-fix leaves the resource with **no** definer (`None`) vs pre-fix `Some(1)`. **Not a confirmed regression**: the input is spec-contradictory and not resolver-producible (a genuine rep holder *exports* the interface, it doesn't wire its own same-interface import to a sibling), and the pre-fix `Some(1)` was itself the *wrong* re-exporter attribution — so it's a change between two degenerate answers on an unreachable input, corroborated by 0 integration-suite failures.

**Additional safety note:** even if that degenerate input were somehow reachable, post-fix (`None`) degrades to a **loud** missing-table trap at `resolver.rs:1891`, whereas pre-fix (`Some(1)`) was a *silent* wrong-attribution — so the change is **safe-or-better** (fail-loud beats silent-wrong) in the worst case.

No failing reachable test could be constructed → **REFUTED**. The hardening is sound. Working tree confirmed clean.

avrabe:
## Iteration: mythos-auto caught an `[export]`-prefix gap my manual clean-room pass missed

The mythos-auto discover pass flagged a real incompleteness in the first commit (c29d962): the `rep_backed_resources` guard used **exact** interface-string keys, but `defines_resource`/`resource_definer` and the terminal-exporter `iface_matches` are `[export]`-prefix-**insensitive**. So a rep holder registered under `svc:lib/api` failed to match a terminal whose resource node is keyed `[export]svc:lib/api`, defeating the guard in the `[export]`-alias case → `defines_resource(no_rep_terminal, "svc:lib/api", "thing")` wrongly returned true.

**Verified the finding** (not taken on faith): added `ls_a_17_rep_backed_guard_is_export_prefix_insensitive`, confirmed it FAILS on c29d962, then fixed by `[export]`-normalising the prefix on both the set keys and the membership check (matching `defines_resource`'s semantics). Now red→green (commit f964a6b).

This is the dual-gate working as intended — the automated discover caught what the bounded manual pass missed. Honest note for the record: my earlier "REFUTED" verdict was sound *for the cases it tested* but incomplete; the `[export]` variant is now covered by a dedicated oracle.

Re-verification on f964a6b: resource_graph 5/0; full `meld-core` suite 0 failed; clippy `-D warnings` + fmt clean; LS-N 57/0/0; rivet 0 broken cross-refs. The fresh auto-discover on this push should report NO FINDINGS.

Issue #305, opened 2026-06-23
opaque-rep resource: construct+drop traps during drop teardown (fuse oracle exists on feat/explore-I-opaque-rep, not in main)


avrabe:
## Why this is filed (rescued from a forgotten branch)

A branch-usefulness audit found **unrecovered work** on the un-merged, no-PR branch `feat/explore-I-opaque-rep` (commit `78d0fd7`, 2026-04-24). The opaque-rep *feature* landed (`FuserConfig.opaque_resources` is in `main`), but this branch's **test oracle never made it into `main`**, and it documents a **real meld-side bug that is not tracked in any open issue**. Capturing both here so they aren't lost; the branch can then be cleaned up once the test is cherry-picked.

## The bug

A 3-component re-exporter chain whose intermediate exports a resource opted into opaque-rep (built with `pulseengine/wit-bindgen@feat/opaque-rep-attribute` `--opaque-export-resources float`) **fuses cleanly and runs construct-only**, but **construct+drop traps with a memory fault during drop teardown** under meld fusion.

Opaque-rep makes the constructor return the leaf's inner handle as the rep verbatim, skipping the `Box::into_raw` + `assume(ptr.is_aligned())` chain — which is what *removes* the `& 7` alignment trap of the standard `resource_floats` fixture (the long-standing #69/#92 trap). Removing that alignment trap **unmasks** a separate, deeper bug in the drop path.

Documented fix direction (from the branch commit message):
1. **explicit inner-handle drop forwarding** in the opaque-rep destructor, AND
2. **per-resource handle-table discrimination** in `meld-core/src/merger.rs` (the re-exporter handle-table allocation — relates to the parked `feat/per-component-handle-tables` Option-A work and the closed #107 path-B routing).

## Repro / oracle (to cherry-pick into main)

`feat/explore-I-opaque-rep` adds, in `meld-core/tests/wit_bindgen_runtime.rs`:

```rust
fuse_only_test!(
    test_fuse_wit_bindgen_resource_floats_opaque,
    "resource_floats_opaque"
);
```

- The `fuse_only` oracle **passes today** (meld fuses 3→1, validates with wasm-tools). It should be cherry-picked into `main` as the standing fuse oracle.
- The runtime construct+drop path is what traps — that's the bug; it warrants a `fuse_only`→`runtime` upgrade once fixed.
[...]

Issue #304, opened 2026-06-22
Fusion emits identity trampolines that never inline (dead inline_adapters flag); de-export fusion-vestigial symbols


avrabe:
### Context
gale's dissolved-codegen work (pulseengine/gale#97) found that un-inlined forwarding wrappers cost ~2× cycles on tiny functions (an extra `bl` + stack round-trip + second prologue). For the **meld-fused** demonstrator path (gale-app-demo + gale-kiln → `meld fuse --memory shared --address-rebase` → loom → synth → 668 B native, gale `benches/gust/build-fused.sh`), meld introduces the same pathology structurally.

### Finding 1 — identity trampolines + a dead `inline_adapters` flag
meld emits **one adapter function per cross-component call edge** and rewires the caller to it (`meld-core/src/lib.rs:wire_adapter_indices:372`; `adapter_base` :380). For the common shared-memory, no-transcode, no-resource case meld emits a pure forwarding thunk — `generate_direct_adapter` (`adapter/fact.rs:552-567`): re-push every param, `call $target`, end. Left un-inlined into native, each becomes the `bl`-chain + frame round-trip we measured.

`AdapterConfig.inline_adapters` (`adapter/mod.rs:35`) is **set `true`** (lib.rs:296) but **read nowhere** in meld-core (grep finds only the definition + assignment). *Ask:* honor it — when `generate_adapter` selects the identity-trampoline case (analyze_call_site → direct adapter, same memory, no post-return/resource ops), **rewire the import directly to `target_func`** instead of emitting a thunk. Sound (the body only forwards), and it's the leanest output — fewer functions for loom/synth, no thunk for loom to inline (and loom currently doesn't inline small multi-site funcs anyway — see pulseengine/loom issue).

### Finding 2 — de-export fusion-vestigial symbols (generalize meld#298)
gale's `build-fused.sh` must manually `grep -vE '\(export "(cabi_realloc|gale:kernel|__data_end|__heap_base)'` AFTER meld so loom/synth can DCE the now-dead canonical-ABI/realloc path. meld emits these exports (`merger.rs` cabi_realloc :1218-1256). Once fusion closes the import graph, any export that existed only to satisfy a now-resolved cross-component import is vestigial. *Ask:* de-export them at fusion time (a `--strip-vestigial` flag or default) — generalizes meld#298 beyond cabi_realloc, removes the manual strip step.

### Not meld's gap (honest)
[...]

Issue #303, opened 2026-06-22
Traceability debt: 180 rivet validate errors — requirements trace only to GitHub issues, host controllers unmodeled, dangling UCA refs, verification right-side unmodeled


avrabe:
## Summary

`rivet validate` on meld reports **180 errors** (34 of them broken links). Investigating the broken links (prompted by pulseengine.eu#91's traceability-visibility thread) revealed this is **structural traceability debt requiring safety-engineering judgment**, not a mechanical cleanup. Filing so it's tracked rather than fabricated-over.

Grounded in `rivet 0.17.0` against the current tree.

## The 34 broken links — three classes, all needing judgment

**Class A — requirements trace ONLY to GitHub issues (18 links, SR-32…SR-41).**
e.g. `SR-33` has exactly two links: `derives-from → #94`, `tracked-by → #141`. rivet resolves `#NNN` as artifact IDs (they don't exist) → broken. The deeper problem: these SRs have **no `derives-from` to any real STPA artifact** (constraint/scenario/hazard). Fixing means *authoring* the real upstream link (which constraint/scenario each requirement derives from — a modeling decision) and moving the issue ref to a metadata field (`upstream-ref`).

**Class B — UCAs reference host-side controllers absent from meld's control structure (12 links).**
`UCA-RT-* → "CM Runtime (embedded)"`, `UCA-HB-* → "host-wit-bindgen"`, `UCA-FM-* → "Fiber Manager (host intrinsic)"`. meld's control structure has 7 controllers (CTRL-ADAPTER/BUILD/CLI/MERGER/PARSER/RESOLVER/WRAPPER) — none of these. STPA **scope decision**: model the host/runtime boundary as (external) controllers, or these UCAs are out of meld's control scope.

**Class C — loss-scenarios reference UCA IDs that never existed (4 links).**
`LS-A-17`/`LS-A-18 → UCA-F-2`, `LS-A-9 → UCA-F-3`, `LS-CP-5 → UCA-CP-1`. `git log -S` shows `UCA-F-*`/`UCA-CP-*` were never in `ucas.yaml` (existing prefixes: A/FM/HB/M/P/R/RT/W). Need the correct UCA identified or the missing UCA authored.

## The other ~146 errors + the right-side gap

- `rivet coverage`: **`requirement-coverage 0/45 = 0.0%`** — the rule wants each requirement `satisfied` by a `design-decision`/`feature`, and the project has **zero** of those artifact types.
[...]

Issue #300, opened 2026-06-21
Coordinate the dissolved-library-OS inter-component isolation model (gale#86/#404, synth#404)


avrabe:
**[cross-repo coordination — synth maintainer side; not a meld bug]** gale's BYO-OS work (gale#86, gale#404) is choosing the inter-component memory-isolation model for the dissolved library-OS, and the fusion-structure half of that decision is meld's — so meld should be in the loop before gale commits.

## The fork

1. **Single shared memory + MPU-carve** — `meld fuse --memory shared` → one linear memory; the embedder/TCB carves MPU sub-regions per component by hand. This is meld's current direction (meld#172 closed in favor of shared; meld#298 / meld#299 actively enabling `--memory shared`).
2. **Multi-memory → one native region per wasm memory** — meld preserves each component's memory; synth places each at a distinct native base so the memory boundary *is* the MPU/PMP region boundary (protection coincides with semantic boundaries — a component can't even *form* a pointer into a neighbor; under model 1 a miscomputed address lands in a neighbor's sub-range and only the MPU catches it).

## synth's stance (synth#404)

synth lowers whichever structure meld emits; it will **not** be the blocker for model 2 — concretely: carry `memidx` through the IR instead of dropping it, place each wasm memory at a distinct native base, emit cross-memory ops explicitly (closing synth#369), and expose per-memory base/size so the embedder programs one MPU/PMP region per memory. Today synth is single-memory, so **model 1 is what's lowerable now**; model 2 is forward-looking on the synth side too.

## The question for meld

Is `--memory shared` the **committed** dissolved-MCU path (isolation = embedder MPU-carve on one region), or does meld intend to support a fuse mode that **preserves component memories as distinct regions** for structural isolation? gale#86's isolation design forks on this. Note the MPU/PMP region cap (~8 on ARMv7-M, PMP similar) bounds the component count N either way — so it's not a differentiator between the models.

Refs: gale#86, gale#404, synth#404, synth#369, meld#298, meld#299.

avrabe:
## meld-side answer: commit to **model 2 (multi-memory structural isolation)** as the dissolved-MCU isolation path

Decision (meld maintainer): the committed dissolved-library-OS isolation model is **structural** — meld preserves each component's memory as a distinct region so the MPU/PMP boundary *is* the semantic boundary. `--memory shared` + embedder MPU-carve (model 1) is retained as a **secondary** mode (lean, lowerable-now), not the committed isolation path.

[...]

PR #306, opened 2026-06-23
release: v0.34.0 — identity-trampoline adapter inlining (#304) + multi-memory isolation commitment (#300/ADR-4)

avrabe:
## Release v0.34.0 scope

Bundles the verified session work (SR-42, ADR-4, plan SR-43/44).

### Headline
- **#304 — identity-trampoline adapter inlining (SR-42).** Honors the previously-dead `inline_adapters` flag: a pure identity `Direct` adapter (`local.get*; call target; end`) is bypassed — the caller's import is wired straight to the target, the dead thunk left for loom to DCE. No re-index (all adapter index maps preserved). Fail-safe condition: `Direct && no resource ops && callee_post_return.is_none()`. New `FusionStats.adapters_inlined`.
- **#300 — multi-memory structural isolation committed as the dissolved-MCU model (ADR-4).** `Auto` left unchanged but a warning fires when it resolves a strategy on an attested build ("explicit, not auto" for functional safety). Multi-memory isolation verified to scale to N=3 components. `#298`/`#299` (shared-memory de-export) reclassified secondary; their oracle/verdict foundation rides along (inert, sound).

### Release plan (rivet)
- SR-42 [v0.34.0, implemented], SR-43 [v0.35.0, opaque-rep drop #305], SR-44 [v0.36.0, enforced traceability #303].

## Verification
- Full suite green per-commit (clippy + cargo test). Cross-component/adapter/runtime/e2e: `cross_component_call`, `adapter_safety`, `runtime_intra_adapter`, `runtime_from_exports`, `wit_bindgen_runtime` (73), `golden_e2e` (Tier-B behavioural equivalence).
- #304 (Tier-5: `adapter/fact.rs`, `lib.rs`) **clean-room adversarially verified — SOUND** across 5 angles; guard↔trampoline-body coupling pinned by reciprocal invariant comments. Awaiting the repo Mythos gate / `mythos-pass-done`.

## Falsification statement
If meld inlined a *non-identity* adapter (one that transcodes, converts resources, or runs post-return), the cross-component runtime suites would mis-execute or trap. They pass — including `golden_e2e` Tier-B behavioural equivalence on the real wac-composed consumer→provider fixture, where ≥1 identity adapter is inlined.

Refs: #304, #300, #298, #299. Plan: SR-42/43/44.

### DIFF
--- a/meld-core/src/adapter/fact.rs
+++ b/meld-core/src/adapter/fact.rs
@@ -5306,6 +5306,28 @@ impl FactStyleGenerator {
             (t, b, AdapterClass::Direct)
         };
 
+        let target_function = self.resolve_target_function(site, merged)?;
+
+        // #304: a Direct adapter is a *pure identity trampoline* — body is
+        // exactly `local.get*; call target; end` — only when it carries no
+        // resource conversions AND no post-return. (The Direct `else` branch in
+        // `generate_direct_adapter` still emits `call post_return` when
+        // `has_post_return && result_count == 0`, so post-return must be
+        // excluded entirely, not just the result>0 case — wiring the caller
+        // straight to the target would otherwise skip the post-return cleanup.)
+        // When inlining is enabled and that holds, record the target so
+        // `wire_adapter_indices` bypasses this thunk. Fail-safe: any doubt → None.
+        let inline_target = if self.config.inline_adapters
+            && matches!(class, AdapterClass::Direct)
+            && options.resource_rep_calls.is_empty()
+            && options.resource_new_calls.is_empty()
+            && options.callee_post_return.is_none()
+        {
+            Some(target_function)
+        } else {
+            None
+        };
+
         Ok(AdapterFunction {
             name,
             type_idx,
@@ -5314,8 +5336,9 @@ impl FactStyleGenerator {
             source_module: site.from_module,
             target_component: site.to_component,
             target_module: site.to_module,
-            target_function: self.resolve_target_function(site, merged)?,
+            target_function,
             class,
+            inline_target,
         })
     }
 
@@ -10299,6 +10322,9 @@ impl FactStyleGenerator {
             target_module: site.to_module,
             target_function: target_func,
             class: AdapterClass::Async,
+            // Async-lift adapters transform (await/result delivery) — never a
+            // pure identity trampoline, so never inlined (#304).
+            inline_target: None,
         })
     }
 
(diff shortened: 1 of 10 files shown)


avrabe:
## Mythos delta-pass — NO FINDINGS (4/4 Tier-5 files)

Ran the `scripts/mythos/discover.md` protocol on each touched Tier-5 file, scoped to this PR's delta. Per the protocol's "both a Kani harness AND a failing PoC, or do not report" rule, no divergence could be constructed for any file → NO FINDINGS.

### `meld-core/src/adapter/fact.rs` — NO FINDINGS
Delta: #304 `inline_target` computation (`generate_adapter`) + async sites set `None` + a doc comment. Fresh-context adversarial discover pass attacked all three corruption vectors:
1. **Guard ⊇ complex-branch trigger** — the inline guard (`Direct && resource_rep_calls.empty() && resource_new_calls.empty() && callee_post_return.is_none()`) is provably *stronger* than `generate_direct_adapter`'s pure-forward condition; it can only fire when the body is exactly `local.get*; call target; end`. Guard and body-generator read the *same* `options` (computed once, passed by ref) — no recomputation divergence.
2. **Type validity** — `AdapterFunction.type_idx` is the target's type; the thunk already exposed that type to the caller, so inline preserves the identical caller-facing contract. P3 i32→i64 widening goes through the wrapper, which takes precedence over inline in `wire_adapter_indices`.
3. **Target index** — `inline_target` is the *same* `resolve_target_function` value the thunk's `call` uses; nothing for the thunk to "mask."

### `meld-core/src/rewriter.rs` — NO FINDINGS
Delta: #298 part (b) `IndexMaps::defer_grow_under_rebase` (default `false`) + a deferred `memory.grow` → `Unreachable` branch under address-rebasing. Fresh-context pass: the flag has **no production setter** (4 crate-wide hits, all in rewriter.rs; 2 in `mod tests`); the sole production constructor `build_index_maps_for_module` never sets it. So the default path is byte-identical (memory.grow under rebasing still hard-errors); the `Unreachable` branch is statically dead in production. `IndexMaps` derives only `Debug/Clone/Default` — no eq/hash/serde perturbation; all construction is `new()`+assign (no positional/exhaustive hazard).

### `meld-core/src/adapter/mod.rs` — NO FINDINGS
Delta: a single struct field `pub inline_target: Option<u32>` on `AdapterFunction`. No production logic; no attackable surface.

### `meld-core/src/merger.rs` — NO FINDINGS
Delta: +359 lines, **all under `mod tests`** (the #298 `test_298_*` oracles). No production code changed.

Method: two fresh-context discover sessions (fact.rs, rewriter.rs — the files with production-logic deltas); the other two have no production delta. This corroborates the in-session clean-room verification of #304 (SOUND, 5 angles).

Issue #301, opened 2026-06-22
fuse: pass through embedder import `cabi-arena-realloc` instead of trying to satisfy it


avrabe:
## Context

BYO-OS lean-MCU dissolve pipeline (gale#89):

```
gale app + kernel-primitive components
  → meld fuse --memory shared
  → loom → synth (native ELF, no memory.grow)
```

Grow-free components (forked wit-bindgen `cabi-realloc-extern`, pulseengine/wit-bindgen#4) route `cabi_realloc` to an embedder-provided `env::__cabi_arena_realloc`, satisfied at native link / synth dissolve — not by the component model.

## What wasm-tools now produces (contract)

pulseengine/wasm-tools#2 adds `component new --import-passthrough env::__cabi_arena_realloc`, which emits a **valid** component carrying a world-root import:

```wit
world root {
  import cabi-arena-realloc: func(p0: u32, p1: u32, p2: u32, p3: u32) -> u32;
  export greet: func() -> string;
}
```

(extern name = `kebab_case(core_field_name)`; `__cabi_arena_realloc` → `cabi-arena-realloc`. Internally it's `canon lower`ed and wired to the core module's `env::__cabi_arena_realloc`.)

This import is **embedder-provided** — meant to survive down to synth, not be resolved during fusion.

## Ask (meld)

When fusing components — especially `fuse --memory shared` (meld#298/#300) — recognize this embedder import and **pass it through to the fused artifact unsatisfied**:

- do **not** error that the import is unresolved;
- do **not** attempt to map it to a provider component in the fusion set;
- preserve it (by extern name `cabi-arena-realloc`, or as the underlying `env::__cabi_arena_realloc` core import if meld emits a core module) so synth can bind it at dissolve (see synth#418).

Open question for us to settle across meld/synth/wasm-tools: how meld should mark "this is embedder-provided, leave it alone." Options: match on the well-known name `cabi-arena-realloc`, or introduce a dedicated namespaced interface (e.g. an `embedder` ABI interface) that wasm-tools emits instead of a bare world-root func. The wasm-tools fork currently emits the bare kebab name; happy to switch to a namespaced interface if meld prefers a non-name-matching signal.

## Refs
pulseengine/wasm-tools#2, pulseengine/wit-bindgen#4, pulseengine/synth#418, meld#298, meld#300, gale#89.

> Pulseengine fork pipeline only — nothing goes upstream to bytecodealliance.

avrabe:
[...]

PR #315, opened 2026-06-26
feat(resolver): explicit embedder-passthrough recognition (#301 inc 1)

avrabe:
## What

Closes the meld side of #301 inc 1: **explicit recognition of embedder-provided passthrough imports** in the `pulseengine:embedder` WIT namespace.

The BYO-OS lean-MCU dissolve pipeline (gale#89) routes `cabi_realloc` to an embedder-provided seam (e.g. `pulseengine:embedder/arena`'s `__cabi_arena_realloc`), satisfied downstream at native link / synth dissolve — **never** by a component in the fusion set.

Per avrabe's meld-side answer on #301: core-module passthrough already worked, but only **incidentally** (the fusion resolver runs `allow_unresolved: true`). A future strict/aggressive resolver could silently drop or reject these seams, and a bare `cabi-arena-realloc` magic-name contract is collision-prone and invisible to review. This makes the passthrough **explicit** — same "explicit, not auto" stance that settled #300.

## Changes
- `EMBEDDER_PASSTHROUGH_NAMESPACE` (`pulseengine:embedder`) + `is_embedder_passthrough()`: matches by **package namespace**, version-tolerant, no prefix false-positives (`embedderx`/`embed` rejected).
- `resolve_component_imports_with_hints` recognizes such imports as intentional passthrough — never bound to a coincidental provider, **exempt from the strict-mode `UnresolvedImport` error**. The two error sites collapse into one with an explicit embedder arm.
- **LS-R-17** (approved): documents the mis-bind (UCA-R-3, H-1/H-3.1) and false-rejection hazards + the namespace-recognition mitigation.

## Tests / gates
- `ls_r_17_embedder_passthrough_survives_strict_resolution` — strict resolver keeps the embedder import but **still errors on a non-embedder unexported import** (exemption is namespace-scoped, not over-broad).
- `test_is_embedder_passthrough_recognizes_namespace_only` — namespace match incl. version suffixes; rejects look-alikes.
- `test_298_fork_arena_realloc_...` now also asserts the retained arena import is recognized by namespace (ties the real fork fixture to the predicate).
- **443 lib tests pass; `rivet validate` PASS; LS gate `LS-R-17 [OK]`.**

## Scope / follow-up
The `--output component` (P2 wrap) exposure of embedder imports remains a **tracked follow-up** — needs a `wasm-tools component new --import-passthrough`-produced fixture (fork-domain, pulseengine/wasm-tools#2).

Refs: #301, gale#89, pulseengine/wasm-tools#2, pulseengine/wit-bindgen#4, pulseengine/synth#418.

### DIFF
--- a/meld-core/src/merger.rs
+++ b/meld-core/src/merger.rs
@@ -5557,6 +5557,26 @@ mod tests {
                 .map(|i| format!("{}/{}", i.module, i.name))
                 .collect::<Vec<_>>()
         );
+
+        // #301: the retained import's module is in the recognized
+        // `pulseengine:embedder` namespace, so meld preserves it as INTENTIONAL
+        // embedder passthrough — not merely incidental survival of an unresolved
+        // import. This ties the real fork fixture's module name to the explicit
+        // recognition predicate the strict-resolution carve-out relies on.
+        assert!(
+            merged
+                .imports
+                .iter()
+                .any(|i| i.name == "__cabi_arena_realloc"
+                    && crate::resolver::is_embedder_passthrough(&i.module)),
+            "retained arena import must be recognized as embedder passthrough by namespace, \
+             got imports: {:?}",
+            merged
+                .imports
+                .iter()
+                .map(|i| format!("{}/{}", i.module, i.name))
+                .collect::<Vec<_>>()
+        );
     }
 
     // -- SR-31: Multiply-instantiated module detection -------------------------
--- a/meld-core/src/resolver.rs
+++ b/meld-core/src/resolver.rs
@@ -1587,6 +1587,40 @@ fn resolve_resource_positions(
     resolved
 }
 
+/// The well-known WIT package namespace for embedder-provided seams.
+///
+/// Imports in this namespace (e.g. `pulseengine:embedder/arena`'s
+/// `__cabi_arena_realloc`) are satisfied by the *embedder* downstream — at
+/// native link / synth dissolve — NOT by another component in the fusion set.
+/// meld must recognize them as intentional passthrough: never bind them to a
+/// provider, never treat them as a resolution failure, and preserve them into
+/// the fused artifact (meld#301, gale#89).
+///
+/// Recognition is by **namespace**, not by the bare field name
+/// `cabi-arena-realloc`: a magic-name match would be collision-prone and
+/// invisible to review, whereas a namespaced interface is an explicit,
+/// greppable contract (the "explicit, not auto" stance that settled #300).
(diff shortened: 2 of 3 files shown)


avrabe:
## Mythos delta-pass — NO FINDINGS

Per `scripts/mythos/HOWTO-gate.md`, the Tier-5 discover protocol was run on each touched Tier-5 file (fresh adversarial session, `discover.md` framing, scoped to this PR's delta).

### `meld-core/src/resolver.rs` — **NO FINDINGS**
The substantive change: `EMBEDDER_PASSTHROUGH_NAMESPACE` + `is_embedder_passthrough`, the collapse of the two `UnresolvedImport` error sites into one computed `resolved` + an explicit embedder arm, and the strict-mode carve-out. Four adversarial hypotheses, all refuted concretely:

1. **Refactor behavior drift (UCA-R-3)** — refuted. Truth table over all four cases (hint-hit, stale-hint→fallback, no-hint-other-export, no-exports) shows the new `export_index.get(&import.name).and_then(|exports| hinted.or_else(...))` chain binds identically to the original two-branch logic, including hint precedence. `get → None` now flows through `and_then` to the same arms — identical outcome for non-embedder names.
2. **Over-broad recognition / swallowing a satisfiable import (UCA-R-1)** — refuted. The bind arm (`if let Some(resolved)`) runs strictly before the embedder arm, so any real in-set provider still wins; only genuinely-unresolved imports reach the predicate. (A component naming `pulseengine:embedder/*` to pass strict resolution is the *intended* #301 contract.)
3. **Parsing soundness** — refuted by direct execution: `pulseengine:embedderx/foo`, `pulseengine:embed/foo`, `x/pulseengine:embedder`, uppercase, and empty all return `false`; no cross-namespace false positive.
4. **Downstream / merger interaction** — refuted. The component-level path never populated `unresolved_imports` (only the module-level paths do); the new arm touches neither `resolved_imports` nor `unresolved_imports` — exactly the prior non-strict fall-through. Production fusion uses `allow_unresolved: true`, so the only true behavioral delta is in `Resolver::strict()`, which has **no production caller** (test-only) and cannot reach a fused artifact.

### `meld-core/src/merger.rs` — **NO FINDINGS**
This PR's delta to merger.rs is **test-only**: a single added assertion inside the existing `test_298_fork_arena_realloc_fuses_under_shared_rebase_today` (asserting the retained arena import is recognized by `is_embedder_passthrough`). No production logic changed, so the semantic-preservation discover protocol has no new surface.

Adding `mythos-pass-done`.

Issue #319, opened 2026-07-01
meld fuse produces DWARF that fails llvm-dwarfdump --verify (534 out-of-parent + 480 overlapping ranges) while inputs verify clean


avrabe:
**Filed by the PulseEngine challenge harness** (meld v0.37.0, aarch64-apple-darwin). Confirmed by a differential against `llvm-dwarfdump --verify` with a clean input baseline.

## Defect

`meld fuse` produces a core module whose merged `.debug_*` sections **fail `llvm-dwarfdump --verify`** — hundreds of overlapping / out-of-parent DIE address ranges and invalid locations — even though the inputs' DWARF is valid. The DWARF `.debug_*` relocator/merger (#130, #208, #144, all closed) is emitting inconsistent address ranges for the fused/reduced code.

## Repro

```
F=tests/wit_bindgen/fixtures
meld fuse $F/records.wasm $F/variants.wasm -o fused.wasm     # 41.4% reduction
llvm-dwarfdump --verify fused.wasm
```

```
error: Aggregated error counts:
error: DIE address ranges are not contained by parent ranges occurred 534 time(s).
error: DIEs have overlapping address ranges occurred 480 time(s).
error: Invalid DWARF expressions occurred 18 time(s).
error: Invalid DW_AT_location occurred 17 time(s).
Errors detected.
```

Single-component fusion is also affected: `meld fuse records.wasm -o single.wasm` (32.1% reduction) → 272 overlapping + 230 not-contained.

## Clean-room baseline (rules out "inputs were already broken")

`llvm-dwarfdump --verify` can't read the component wrapper directly (reports a bogus `invalid version number: 65549`), so I extracted the inputs' inner core modules with wasm-tools (not meld) and verified those:

```
wasm-tools component unbundle $F/records.wasm --module-dir mods --output /dev/null
llvm-dwarfdump --verify mods/unbundled-module0.wasm   # -> No errors.
llvm-dwarfdump --verify mods/unbundled-module1.wasm   # -> No errors.
```

Both inner core modules verify **clean**. So the overlapping/uncontained ranges are introduced by meld's fuse transform, not inherited from the inputs.

## Why it matters

The fused output is what Loom optimizes and Synth compiles; a debugger (or witness MC/DC / source-map tooling that consumes this DWARF) resolving an address to a DIE gets ambiguous or wrong results. For a toolchain aimed at source-level debugging of safety-critical firmware, silently-invalid DWARF after fusion defeats the purpose of the relocator feature.

[...]

PR #320, opened 2026-07-01
fix(dwarf): remap DW_AT_high_pc lengths to the fused layout (#319 inc 1)

avrabe:
## What

Fixes the dominant cause of **#319** — `meld fuse` producing DWARF that fails `llvm-dwarfdump --verify` with hundreds of out-of-parent + overlapping DIE ranges.

## Root cause
gimli routes only *address*-form values through `convert_address`; `DW_AT_high_pc` encoded as a **length** (`DW_FORM_udata`, the common Rust/LLVM form) is a constant it copies verbatim. Fused bodies change length (LEB re-encode of the merged index space), so every subprogram/lexical-block/inlined DIE became `[remapped_low, remapped_low + STALE_input_len)` — overrunning its parent (grew) or the next function (shrank).

## Fix
After the gimli conversion, walk the write DIEs and rewrite each `low_pc`+`high_pc`-length pair via a new `AddressRemap::corrected_high_pc`: invert the remap for `low_pc` within its output body, then use the span's `output_body_end` when the range ends at the function end (avoids the aliased-boundary `translate()` None) or `translate` for interior ends. Tombstoned (dropped) `low_pc` is left untouched — llvm-dwarfdump ignores those DIEs.

## Measured (records+variants repro)
| | before | after |
|---|---|---|
| overlapping | 480 | **0** |
| not-contained | 534 | 11 |
| invalid-expr / invalid-loc | 18 / 17 | 18 / 17 |
| **total** | **1049** | **46** |

## Scope
Inc 1 clears ~96% (all overlapping + 98% of out-of-parent). Residual is **distinct** mechanisms tracked as follow-ups on #319: nested `DW_AT_ranges` on inlined/lexical DIEs (11/4 not-contained), and the `.debug_loc` tombstone-vs-base-selection collision (Cause #2, invalid-expr/loc).

Test: `corrected_high_pc_rewrites_length_to_fused_layout` (grew/shrank/interior/tombstone) — deterministic, no llvm-dwarfdump dependency. 24 dwarf tests pass.

Refs: #319.

### DIFF
--- a/meld-core/src/dwarf.rs
+++ b/meld-core/src/dwarf.rs
@@ -236,6 +236,60 @@ impl AddressRemap {
             .unwrap_or(0);
         self.input_code_extent.max(registered_max)
     }
+
+    /// Inverse of [`Self::translate`] within one span: map an **output**
+    /// code address at an instruction boundary back to its **input**
+    /// address. Used to correct `DW_AT_high_pc`-as-length (#319), whose
+    /// stale value is an input-relative length that gimli copies verbatim
+    /// (it is a constant, not an address, so it never reaches the forward
+    /// remap). Returns `None` when `out_addr` does not land on a recorded
+    /// instruction boundary.
+    fn reverse_in_span(span: &FunctionSpan, out_addr: u32) -> Option<u32> {
+        let out_rel = out_addr - span.output_body_start;
+        // Region 1: locals/prefix — preserved verbatim, linear inverse.
+        if out_rel < span.locals_prefix_len {
+            return Some(span.input_start + out_rel);
+        }
+        // Region 2: instruction stream — invert the offset map. DIE ranges
+        // always start on an instruction boundary, so require an exact
+        // `new` match; a miss means "not a boundary", return None.
+        let instr_new = out_rel - span.locals_prefix_len;
+        for e in &span.instr_offsets.entries {
+            if e.new == instr_new {
+                return Some(span.input_start + span.locals_prefix_len + e.old);
+            }
+        }
+        None
+    }
+
+    /// Correct a `DW_AT_high_pc` encoded as a **length** (#319).
+    ///
+    /// `out_low` is the DIE's already-remapped `low_pc`; `stale_len` is the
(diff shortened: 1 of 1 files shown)


avrabe:
## Mythos delta-pass — NO FINDINGS (meld-core/src/dwarf.rs)

Ran the discover protocol on this PR's `dwarf.rs` delta (a fresh adversarial session). Five hypotheses tested, four refuted **empirically**:

1. **Wrong-span reverse map at a boundary** — refuted: output bodies are non-overlapping, so a concrete `out_low` at a shared `output_body_end == next.output_body_start` boundary resolves to the next function's start (its own low_pc), not the prior's exclusive end. The input-space aliasing `translate` must tombstone doesn't exist in output space.
2. **`reverse_in_span` mis-inversion** — refuted: `InstrOffset.new` is strictly increasing (each operator ≥1 output byte), so the exact `new == instr_new` match is unambiguous; the region-1/2 split is the exact inverse of `translate`.
3. **`orig_end` past the function** — the one theoretical wrong-length path (a `high_pc` exceeding its body would `translate` into the next function). Unreachable from valid toolchains (subprograms end at `input_end`; nested blocks/inlines stay interior; CUs use `DW_AT_ranges`, not `high_pc`). **Hardened anyway** (commit 594ba57): bail when `orig_end > span.input_end`, so it's now provably impossible even on malformed input (LS-D-1).
4. **Silent no-op on real DWARF** (the key risk: clang/rustc emit `high_pc` as `DW_FORM_data4`, my match is on `Udata`) — refuted end-to-end: gimli **normalizes** the constant-class `high_pc` to `Udata` in the read→`WriteDwarf::from` path, so the match fires and the length is corrected. Verified with hand-assembled `data4` bytes.
5. **Mutation walk / adapter unit** — refuted: `correct_high_pc_lengths` runs *before* `append_adapter_unit`, so the synthetic `<meld-adapter>` unit is never walked; the DFS collects ids then mutates per id (no cycles, each once).

No failing oracle for any wrong-attribution defect. Adding `mythos-pass-done`.

PR #321, opened 2026-07-01
fix(dwarf): remap DW_AT_ranges lists to the fused layout (#319 inc 2)

avrabe:
## What
Second increment for **#319**. Inc 1 fixed `DW_AT_high_pc` lengths; this fixes the residual **"DIE address ranges are not contained in its parent's ranges"** errors.

## Root cause
DWARF v4 `.debug_ranges` entries are offset pairs **relative to a base** (the CU's `DW_AT_low_pc`, nonzero for wasi-libc CUs). gimli parses them as `AddressOrOffsetPair` and routes the *offsets* through `convert_address`, which treats a small offset (e.g. `0x141`) as an absolute code address and remaps it to an unrelated output location — so the resolved range escapes its parent (the `.debug_ranges` analogue of the inc-1 high_pc bug, one level down). Verified empirically: input inlined child ranges resolve inside `write` `[0xb398,0xb53c)`; output rendered `[0x1c439,…]` outside `[0x133b6,0x1353d)`.

## Fix
`correct_die_ranges` — a post-convert pass that walks read↔write DIEs in lockstep pre-order, resolves each DIE's absolute input ranges via gimli's `die_ranges` (base already applied), `translate`s both endpoints to the fused layout, and re-emits them as `OffsetPair` relative to the output CU base. Unmappable sub-ranges (dropped/tombstoned code) are dropped; an emptied list deletes `DW_AT_ranges` — never a wrong/escaping range (LS-D-1).

## Effect
The **not-contained class is eliminated**: records-only 4→**0**, records+variants 11→**0**. Cumulative with inc 1: **1049→35** (records+variants), **~510→8** (records-only). The remaining 35/8 are **Cause #2** (`.debug_loc` tombstone vs base-address-selection escape) = **inc 3**.

## Test
`inc2_die_ranges_stay_within_enclosing_subprogram` — fuses `records.wasm`, re-parses the fused DWARF, asserts every `DW_AT_ranges` entry sits inside its enclosing subprogram (the invariant that failed). No llvm-dwarfdump dependency. 2 witness + 24 dwarf unit tests pass.

Refs: #319.

### DIFF
--- a/meld-core/src/dwarf.rs
+++ b/meld-core/src/dwarf.rs
@@ -744,7 +744,12 @@ fn correct_die_ranges<R: gimli::read::Reader<Offset = usize>>(
                 ) else {
                     continue; // unmappable (dropped/tombstoned) sub-range → drop
                 };
-                if ob >= out_cu_base && oe >= ob {
+                // Require a strictly-positive extent: a degenerate `begin ==
+                // end` OffsetPair is rejected by gimli's range writer
+                // (`InvalidRange`), which would abort the whole DWARF write
+                // and strip. A zero-length range carries no information, so
+                // dropping it is lossless.
+                if ob >= out_cu_base && oe > ob {
                     new_ranges.push(Range::OffsetPair {
                         begin: (ob - out_cu_base) as u64,
                         end: (oe - out_cu_base) as u64,
--- a/meld-core/src/dwarf.rs
+++ b/meld-core/src/dwarf.rs
@@ -575,6 +575,9 @@ fn rewrite_debug_sections(
     // `llvm-dwarfdump --verify` flags as out-of-parent / overlapping. Fix
     // the lengths to the fused layout now that `low_pc` is remapped.
     correct_high_pc_lengths(&mut write_dwarf, remap);
+    // #319 inc 2: fix DW_AT_ranges lists (base-relative offsets that gimli
+    // mis-routes through convert_address) from the read side.
+    correct_die_ranges(&mut write_dwarf, &read_dwarf, remap);
     // #144 inc 3: the synthetic `<meld-adapter>` unit rides the SAME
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos delta-pass — NO FINDINGS (meld-core/src/dwarf.rs)

Ran the discover protocol on `correct_die_ranges` / `write_dies_preorder`. Five hypotheses; the highest-risk one cleared by source **and** empirically:

1. **Read↔write DIE lockstep misalignment** (would mis-attribute ranges to the wrong DIE) — refuted. gimli 0.31.1 `UnitTable::from`→`convert_entry` is a faithful, order-preserving 1:1 pre-order copy with no tag-based skip/add/reorder (any unsupported DIE fails `from()` wholesale, no partial-skip desync). Instrumented the real code on the fused fixture: **261 units, 195,497 DIEs, 0 tag desyncs, 0 count mismatches**.
2. **Wrong/absent CU base** — the 23/284 units skipped for a non-constant root `low_pc` keep gimli's converted ranges; `--verify` shows none escape (pre-existing boundary, not a regression).
3. **Offset guard** — `ob >= out_cu_base` drops pre-base sub-ranges (fidelity, never mis-attribution). The `begin==end` degenerate that gimli's writer rejects (`InvalidRange`, would abort→strip) is now hardened to strict `oe > ob` (commit in this PR).
4. **low/high reprocessing** — the `DW_AT_ranges.is_some()` gate correctly excludes low/high-only subprograms.
5. **Tombstone sub-ranges** — dropped, so a list only ever describes *less* code, never wrong code (LS-D-1).

**Oracle:** `llvm-dwarfdump --verify` on the fused output → **0** "not contained" errors (the inc-2 target class gone); residual is the untouched `.debug_loc` class (inc 3). Adding `mythos-pass-done`.

Issue #339, opened 2026-07-11
#326 Finding A: path-F gate misses absolute addresses used as values (no-reloc, no direct load/store)


avrabe:
Follow-up from the Mythos discover pass on the #326 reloc-consumer (PR pending).

## The gap

The path-F hard-error gate (`module_has_direct_memory_access` in `merger.rs`) rejects a **no-reloc, non-zero-base** module only when it performs a direct `load`/`store`. It does NOT catch a module that references an absolute address **purely as a value** — `i32.const <abs>; call $import` (a canonical-ABI `&BUF` handed to an imported `memcpy`/`fd_write`) or `i32.const <abs>; return`. With no reloc metadata, the legacy path emits the const verbatim (un-rebased), so such a module hands an un-rebased pointer to the host / another component → silent cross-component corruption. This is exactly datapoint A's pattern (a buffer pointer passed to an import).

## Why the naive fix is wrong (empirically confirmed)

Broadening the gate to reject any `i32.const`/`i64.const` **regresses the legitimate bulk-only case**: `test_address_rebasing_end_to_end` uses `i32.const` values that are *safely consumed by rebased bulk-memory ops* (`memory.copy/fill/init`, dynamically rebased by `append_rebased_address`). The naive broadening rejected it (test FAILED). So a bare const is indistinguishable from an integer, and bulk-op-consumed consts are safe — a sound fix needs **data-flow analysis** to tell an escaping address-const from a rebased-bulk-consumed or plain-integer const.

## Scope / severity

- **Medium.** Silent memory corruption, but only on the **no-reloc** path (already the unsupported path — meld hard-errors the load/store subset and the error advises `--emit-relocs`).
- Does **NOT** affect the supported path: `--emit-relocs` inputs (e.g. gale's) carry reloc metadata, so their address consts are correctly rebased via `reloc.CODE`.
- The residual gap is documented in `module_has_direct_memory_access`'s doc comment (KNOWN LIMITATION).

## Proposed fix

Data-flow-aware gate: reject a no-reloc non-zero-base module iff it has an `i32.const`/`i64.const` that reaches a `call`/`call_indirect`/`return`/`store` WITHOUT being consumed by a dynamically-rebased bulk-memory op. (Or: require reloc metadata for ANY no-reloc non-zero-base module with a non-trivial code section, accepting the loss of the bulk-only-without-relocs convenience — a scope decision.)

Refs #326, ADR-6 (path-F), SR-48.

avrabe:
[...]

Issue #334, opened 2026-07-10
fuse --memory shared leaves 2 MCU-dissolve blockers on a wac-composed node: N __stack_pointer globals + a dead lowered-import shim trampoline


avrabe:
## Summary

Fusing a **wac-composed multi-provider Component-Model node** with `meld fuse --memory shared [--address-rebase]` for an MCU dissolve leaves **two artifacts** that block the downstream `synth --native-pointer-abi --shadow-stack-size` lowering. Both are fixable in the fused wasm by hand — doing so makes the node dissolve **cleanly** (no warnings) to a budget-fitting object — so meld should emit them fused.

Context: this is gale's gust:os v0.4.0 buffer node — an app importing `gust:os` (time + log, `log.line` carries `list<u8>`) `wac plug`ged with a time-provider + log-provider, fused to one shared-memory core, dissolved to a Cortex-M3 relocatable `.o` on an 8 KiB-SRAM part (STM32F100).

## Artifact 1 — N `__stack_pointer` globals not coalesced

`--memory shared` (with or without `--address-rebase`) merges the linear memories but leaves **one mutable `__stack_pointer` global per fused component**, all `init == sp_init` (0x100000):
```
(global (;0;) (mut i32) i32.const 1048576)   ; app-tl SP
(global (;3;) (mut i32) i32.const 1048576)   ; time-provider SP (unused after fuse)
(global (;6;) (mut i32) i32.const 1048576)   ; log-provider SP
```
A shared memory should have **one** shadow stack. synth then can't pick which global to re-base (`--shadow-stack-size: could not uniquely identify the shadow-stack global`, synth#707). Worse, **leaving them separate is unsound**: two components both starting their SP at the shared-memory top would clobber each other's frames on a cross-component call. Manual fix that works: redirect all `global.get/set` of the extra SP globals to global 0 and drop them → one descending shared stack.

## Artifact 2 — dead wac lowered-import shim trampoline not DCE'd

The pre-meld composite (valid) contains wac's lowered-import shim: an `$imports` table + a `call_indirect (type 0)` trampoline, populated at instantiation. After meld flattens to core wasm, the **real** cross-component calls become correct *direct* calls (e.g. `log.line` → `read32`), but a **vestigial trampoline survives**:
```
(func (;14;) (type 19) (param i32 i32)   ; exported as "0"
  local.get 0  local.get 1  i32.const 0  call_indirect 2 (type 9))
[...]

Issue #326, opened 2026-07-10
fuse --memory auto/shared silently corrupts memory across components: load/store dynamic addresses are not rebased (no memory.grow; distinct from #172/#299)


avrabe:
## Summary

`meld fuse` with the **default** `--memory auto` strategy (which resolves to `shared` + "address rebasing" when no input has `memory.grow`) **silently corrupts memory across fused components**: two components that each access their own linear memory at overlapping addresses end up sharing the same physical bytes. One component's private store clobbers the other's data, producing wrong results with **no error and no trap**.

This also reproduces with the explicit `--memory shared --address-rebase`. It is **distinct** from #172/#299 (rebase vs `memory.grow`) — there is **no `memory.grow`** here — and from #298 (adapter internalization).

Faithful `--memory multi` fusion of the same input produces the correct result, which pins the shared path as the incorrect one.

## Root cause

The "address rebasing" only rebases the **static memarg offset** field and the operands of **bulk-memory** ops; it never rebases the **dynamic address operand** of ordinary `i32.load`/`i32.store`.

- `meld-core/src/rewriter.rs` `convert_memarg()` adds `memory_base_offset` to `memarg.offset` only. The effective address is `dynamic_operand + memarg.offset`; the dynamic operand (e.g. a computed pointer, or an `i32.const` pushed on the stack) is left untouched.
- `rewriter.rs` sets `needs_rebase_locals = true` **only** when a function contains `MemoryCopy | MemoryFill | MemoryInit`. Plain loads/stores get no base adjustment.

Real components address their heap/shadow-stack via **computed** pointers (not folded into the static memarg offset), so any non-trivial pair of memory-using components collides. In the repro below the provider's accesses even land at physical base 0 (same as the consumer).

## Minimal reproducer (fully offline: wasm-tools + wac + wasmtime)

`provider_core.wat` — writes a private sentinel `0xABCD` to its own mem[8]:
```wat
(module
  (memory (export "memory") 1)
  (func (export "golden:math/lib@0.1.0#add")(param i32 i32)(result i32)
    (i32.store (i32.const 8)(i32.const 43981))   ;; 0xABCD, provider-private
    (i32.add (local.get 0)(local.get 1))))
```

`consumer_core.wat` — writes `0x1234` to *its* mem[8], calls provider.add, re-reads its mem[8]:
```wat
(module
  (import "golden:math/lib@0.1.0" "add" (func $add (param i32 i32)(result i32)))
  (memory (export "memory") 1)
[...]

PR #340, opened 2026-07-11
fix(326): reloc-driven shared-memory address rebasing (sound --memory shared)

avrabe:
Makes `meld fuse --memory shared` **sound** for relocatable inputs by consuming their relocation metadata to rebase absolute addresses into each module's shared-memory window — fixing the meld#326 silent memory corruption (gale's gust:os buffer + `static mut` datapoints).

## What it does
- **Contract (ADR-6 / SR-48):** input = **relocatable** (`linking` + `reloc.*`, produced by `-C link-arg=--emit-relocs`). PIC is architecturally excluded (`wasm-tools component new` refuses core modules importing memory/`__memory_base` globals). Settled by a spike (see #326 comment); gale-side producer resolved (gale#168, one flag).
- **Reader (increment 1):** `reloc.rs` hand-parses `linking` v2 + `reloc.CODE`/`reloc.DATA` (wasmparser 0.246 dropped these readers). Verified the metadata survives `component new` byte-for-byte.
- **Consumer (increment 2, Tier-5):** per module, translate each `R_WASM_MEMORY_ADDR_*` site through the rewriter `offset_map` (the #331 AddressRemap seam), add the module's `memory_base` to the `i32.const`/`i64.const` literal (code) or 4-byte LE pointer (data), strip consumed relocs. memarg rebasing switched blanket→reloc-driven (also fixes latent over-rebasing of struct-field offsets). No-reloc + non-zero-base + direct memory access → **hard-error** (path-F), never silent corruption.

## Verification (isolated target)
- **457 lib + 4 integration tests green.** Key oracles: `test_326_reloc_const_rebasing_end_to_end` (address rebased at **runtime** via wasmtime), `test_326_shared_rebase_without_relocs_hard_errors` (path-F), and the pre-existing `test_address_rebasing_end_to_end` (bulk-only dynamic rebasing) still green. clippy + fmt clean.

## Mythos discover pass (Tier-5)
Core rebasing arithmetic verified **sound** (coordinate alignment across all functions, rebase-exactly-once, LEB/overflow, reloc-type filtering). Three gate/scope findings, all resolved:
- **A** (value-as-pointer gate miss) — naive fix regresses the legitimate bulk-only case (empirically confirmed), so reverted to a documented KNOWN LIMITATION + **#339** (needs data-flow-aware gate). Does NOT affect `--emit-relocs` inputs (their consts are rebased via `reloc.CODE`).
- **B** (memory64 8-byte data pointers) — reject via `has_unhandled_data_addr_relocs` (+ unit test) rather than mis-rebase.
- **C** — corrected the `MissingRelocMetadata` message (`--emit-relocs` on final link; a `wasm-ld -r` object is addends-only).

## Falsification
If reloc rebasing regressed, `test_326_reloc_const_rebasing_end_to_end` (asserts the address literal equals `base + addr` at runtime) fails. If path-F regressed, `test_326_shared_rebase_without_relocs_hard_errors` fails. If the bulk path regressed, `test_address_rebasing_end_to_end` fails.

Refs #326, #334, #339, gale#168, ADR-6, SR-48, LS-M-11, SYS-6.

### DIFF
--- a/meld-core/src/error.rs
+++ b/meld-core/src/error.rs
@@ -56,6 +56,26 @@ pub enum Error {
     #[error("memory strategy not supported: {0}")]
     MemoryStrategyUnsupported(String),
 
+    /// A core module must be placed at a non-zero shared-memory base but
+    /// carries no relocation metadata, so meld cannot rebase its absolute
+    /// addresses into the shared window (issue #326, ADR-6 path-F). Emitting
+    /// it unchanged would collide it with a prior module's addresses and
+    /// silently corrupt memory, so fusion hard-fails instead.
+    #[error(
+        "component '{component}' module {module} is placed at a non-zero shared-memory base but \
+         carries no relocation metadata (linking/reloc.*); its absolute addresses cannot be \
+         rebased safely. Rebuild it with relocations retained on the FINAL link \
+         (`-C link-arg=--emit-relocs`) — an unresolved relocatable object \
+         (`wasm-ld -r`) is NOT sufficient, its stored values are addends, not final \
+         addresses — or fuse without shared memory"
+    )]
+    MissingRelocMetadata {
+        /// Display name of the component that owns the offending module.
+        component: String,
+        /// The offending module's index within its component.
+        module: String,
+    },
+
     /// Adapter generation error
     #[error("adapter generation failed: {0}")]
     AdapterGeneration(String),
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -60,6 +60,7 @@ pub mod p3_bridge;
 pub mod p3_stream;
 pub mod parser;
 pub mod provenance;
+pub mod reloc;
 pub mod resolver;
 pub mod resource_graph;
 pub mod rewriter;
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -1022,6 +1022,10 @@ impl Fuser {
                 memory_initial_pages,
                 data_segment_base,
                 elem_segment_base,
+                // #326: this adapter-redirect re-rewrite runs with
+                // `memory_base_offset == 0` (no rebasing), so no reloc
+                // consumption is needed here.
+                None,
             );
 
             let import_func_count = module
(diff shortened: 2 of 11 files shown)


avrabe:
## Mythos discover pass (manual, Tier-5) — done

Ran an adversarial discover pass over the reloc-consumer diff (the automated per-file pass will likely time out on merger.rs/rewriter.rs). Focus areas and verdict:

**Core rebasing arithmetic — VERIFIED SOUND:**
- **Coordinate alignment** (highest risk): `reloc.CODE` offsets and `into_iter_with_offsets()` positions are both code-section-content-relative (`extract_function_body` slices from `CodeSectionStart.range.start`), so operator `old_pos` is content-relative for **every** function, not just the first. The `[old_pos, op_end)` range test correctly attributes a reloc site (which points at the immediate) to its operator; the `usize::MAX` sentinel only over-matches the terminal `End` (which ignores `addr_reloc`). The e2e test exercises 3 consts across 3 functions.
- **Rebase-exactly-once:** const-form and memarg-offset-form are mutually exclusive per effective address; `append_rebased_address` is skipped under reloc mode so bulk operands aren't double-counted; base-0 modules are a clean no-op.
- **LEB/encoding & reloc-type filtering:** `wrapping_add` preserves bit-patterns; `RelocType::Other` never panics; index relocs (function/table/type/global) are correctly excluded from address rebasing.

**Three gate/scope findings — resolved:**
- **A** (path-F misses an absolute address used purely as a *value* — `i32.const &BUF; call $import` — in a no-reloc module): the naive gate broadening regresses the legitimate bulk-only case (`test_address_rebasing_end_to_end`, empirically confirmed — a bulk-consumed const is safe), so a sound fix needs data-flow. Reverted to a documented KNOWN LIMITATION + tracked as **#339**. Does not affect `--emit-relocs` inputs.
- **B** (memory64 8-byte `reloc.DATA` pointers silently dropped while segment placement is rebased): now rejected via `RelocInfo::has_unhandled_data_addr_relocs` (+ unit test `unhandled_data_addr_relocs_flags_non_i32_pointers`).
- **C** (`MissingRelocMetadata` advised `--relocatable`, which yields addends-only modules the consumer mis-rebases): corrected to `--emit-relocs` on the final link.

Findings B/C fixed in this PR; A filed as #339 with rationale. `mythos-pass-done`.

PR #343, opened 2026-07-11
feat(334): MCU-dissolve fixups — coalesce __stack_pointer globals + DCE dead import shim

avrabe:
Makes a `meld fuse --memory shared` wac-composed multi-provider node dissolve cleanly on an MCU (SR-49) — the "ergonomics gate" gale currently hand-works-around with WAT surgery, and the companion to #326.

A post-merge pass (`mcu_dissolve.rs`, gated on `SharedMemory`, run before encoding so DWARF/attestation stay consistent) does two fixups:

1. **Coalesce the N per-component `__stack_pointer` globals into one** shared shadow stack. Leaving them separate is **unsound** (providers descending from the shared-memory top clobber each other's frames on a cross-component call) and blocks synth's shadow-stack rebasing (synth#707). Keep one survivor, redirect `global.get`/`global.set` of the duplicates via the existing rewriter machinery, renumber survivors down, drop extras.
2. **Drop the dead lowered-import shim trampoline's keep-alive export** (the numeric `"0"` export over an unreachable, import-dispatched `call_indirect`) so synth DCEs it instead of loud-declining (synth#642/#676).

## Conservative detection (the risk area)
SP coalescing uses **only** the authoritative `name`-section signal (`__stack_pointer`). An init-value fallback heuristic was **rejected during this branch's Mythos pass** — it could fuse two unrelated mutable-i32 globals in different components that merely share an init value, silently clobbering one component's writes. Absent the name, meld does not guess. Trampoline drop requires numeric name + strict body shape + unreachable + import dispatch table.

## Tests (457 lib + 4 integration green; clippy + fmt clean)
- `stack_pointer_globals_coalesced` — 3→2 globals, exactly one SP remains, redirect + renumber verified.
- `shared_stack_pointer_runtime_non_clobber` — **wasmtime**: pre-coalesce the nested provider call clobbers the caller's frame; post-coalesce it doesn't.
- `dead_import_trampoline_export_dropped` — dead `"0"` dropped; real/vtable/non-trampoline exports kept.
- `conservative_no_coalesce_when_unsure` — single-SP, differing-init, AND unnamed same-init all → 0 coalesced.

## Falsification
If SP coalescing regressed, `shared_stack_pointer_runtime_non_clobber` (frames alias) fails. If detection became unsafe, `conservative_no_coalesce_when_unsure` (unnamed same-init coalesced) fails. If the trampoline drop hit a live export, `dead_import_trampoline_export_dropped` (kept exports) fails.

Refs #334, SR-49, SYS-2, SYS-8, synth#707.

### DIFF
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -53,6 +53,7 @@ pub mod component_wrap;
 pub mod custom_merge;
 pub mod dwarf;
 mod error;
+pub mod mcu_dissolve;
 pub mod memory_probe;
 pub mod merger;
 pub mod p3_async;
@@ -714,6 +715,21 @@ impl Fuser {
         // the adapter trampolines instead of calling the target directly.
         if !adapters.is_empty() {
             stats.adapters_inlined = self.wire_adapter_indices(&mut merged, &adapters, &graph)?;
+        }
+
+        // Step 3.6: #334 MCU-dissolve fixups (SR-49, `--memory shared` only).
+        //
+        // Coalesce the N per-provider `__stack_pointer` globals into one
+        // shared shadow stack, and drop the vestigial lowered-import shim's
+        // keep-alive export so the downstream
+        // `synth --native-pointer-abi --shadow-stack-size` dissolve proceeds
+        // without gale's hand WAT surgery. Runs AFTER adapter wiring (so
+        // direct-call flattening is already reflected in the bodies) and
+        // BEFORE encoding (so all three encode passes, DWARF remap, and the
+        // attestation/provenance hashes see the coalesced module). The
+        // non-shared paths never reach here, keeping their output unchanged.
+        if self.config.memory_strategy == MemoryStrategy::SharedMemory {
+            mcu_dissolve::dissolve_fixups(&mut merged, &self.components)?;
         }
 
         // Step 4: Encode output module.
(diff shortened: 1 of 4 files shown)


avrabe:
## Mythos discover pass (manual) — done, one finding fixed

Adversarial pass over `mcu_dissolve.rs` (new fusion-critical post-merge pass). Cleared: global-renumber coverage is complete (bodies + `export (global)`; defined-global inits / active data-elem offsets reference only *imported* globals which never drop), name-section index mapping is correct, trampoline export-drop is conservative and can't strip a live export, and the pass composes with encode/DWARF/attestation (runs before them; no upstream pass captured stale global indices).

**Finding [MEDIUM] — fixed in this PR.** The original SP-detection had an init-value *fallback* (when no module names `__stack_pointer`): fuse any mutable-i32 globals with equal init spanning ≥2 components. That could coalesce two unrelated non-SP globals (an errno/TLS/counter at the same init) → redirect one component's `global.set` onto the other's storage → silent cross-component corruption. The fallback path was also entirely untested (all fixtures supply a `__stack_pointer` name). **Fix:** dropped the fallback — coalescing now uses ONLY the authoritative `name`-section signal; absent it, meld does not guess (the fused module keeps its several SP globals and synth surfaces the multi-SP condition rather than meld corrupting it). Added falsifying test case `conservative_no_coalesce_when_unsure` (c): two unnamed same-init globals → 0 coalesced.

Lower items noted (not blocking): mixed named/unnamed inputs under-coalesce (fails safe, no corruption); the whole-body re-rewrite couples to the rewriter's MVP-opcode whitelist (latent — no synthetic body emits a non-whitelisted op today). Both are fail-safe and can be follow-ups if they ever bite.

457 lib + 4 integration green post-fix; clippy + fmt clean. `mythos-pass-done`.

Issue #299, opened 2026-06-21
fuse --memory shared --address-rebase rejects components with memory.grow (default wasip2 allocator) — blocks single-address-space MCU lowering (jess/gale Pixhawk)


avrabe:
## Surfaced by the issue-hunt (via gale#89), confirmed as a meld-local issue

`meld fuse --memory shared --address-rebase` (the single-address-space MCU lowering the jess/gale Pixhawk pipeline needs) **fails**:

```
unsupported component feature: memory.grow not supported with address rebasing
```

Both `gale-app-demo` and `gale-kiln` carry one `memory.grow` each — emitted by the **default Rust / wit-bindgen wasip2 allocator**. The multi-memory fallback isn't MCU-lowerable (2 memories × 1088 KB; synth loud-rejects). So the canonical dissolve `meld fuse → loom → synth` cannot currently produce the monolithic single-address-space core for the drone target. (gale#89 documents the gale-side view.)

## This is correct fail-loud, not a bug — but it's a hard capability gap

meld's refusal is **sound-by-construction** (SR-37 + `memory_probe.rs`): `MemoryStrategy::SharedMemory` freezes the static memory layout at fuse time, so a runtime `memory.grow` would corrupt the shared single-address-space layout (merger Bug #7). The `module_uses_memory_grow` probe gates Auto→shared exactly on `memory.grow` *absence*, and a parse failure conservatively counts as "uses grow." So the gate must stay; this issue is about adding a **sound path**, not relaxing it.

## Options (the actual work)

1. **Sound `memory.grow` lowering under address-rebasing (preferred):** reserve a bounded static heap region per fused component and rewrite `memory.grow` → a bounded bump within that reservation, failing loud (trap / `AbiError::Oom`) on exhaustion. Preserves the frozen-layout soundness invariant AND accepts default-allocator components. Touches `segments.rs` (address-rebase), `memory_probe.rs`/`MemoryStrategy`, and the rebase math in `rewriter.rs`. Needs an `ls_*`-gated oracle + a Kani/▢ bound-check.
2. **Documented no-grow input contract + actionable diagnostic:** make the error point to the remediation (build components `no_std`/no-grow, or pre-strip the allocator's grow when a static max is known) — the path gale#89 pursues on the gale side. Cheaper; pushes the requirement upstream.
3. **Allocator pre-pass:** when a static heap max is declared, replace the wasip2 allocator's `memory.grow` with a pre-sized static reservation before fusion.

## Cross-refs

[...]

Issue #298, opened 2026-06-21
fuse: drop vestigial cabi_realloc / canonical-ABI adapter once the inter-component boundary is internalized (unblocks --memory shared)


avrabe:
## Observation
After `meld fuse` of two components whose interface is fully resolved internally (app imports `gale:kernel`, host exports it → fused core has **0 remaining `gale:kernel` imports**), the fused core **still exports `cabi_realloc` (×2, one per input) and still contains `memory.grow` (×2)**:

```
(export "cabi_realloc" (func 13))
(export "cabi_realloc$1" (func 32))
memory.grow occurrences: 2
```

Those `cabi_realloc` functions (wit-bindgen canonical-ABI realloc, backed by Rust `dlmalloc → sbrk → memory.grow`) exist **to marshal data across the component boundary** — which fusion has just *removed*. They're vestigial. Because they're *exports*, loom can't DCE them, so the allocator + `memory.grow` survive.

## Consequence
This is what blocks the lean MCU path: `meld fuse --memory shared --address-rebase` fails with *"memory.grow not supported with address rebasing"* — solely because of the dead canonical-ABI allocator. (The user's logic: meld fusion should make the per-component adapter unnecessary; right now it keeps it.)

## Ask
When fusion internalizes the inter-component boundary and the remaining **external** surface is scalar/flat (no realloc-requiring lifts), meld should **de-export / drop the now-internal `cabi_realloc` and canonical-ABI adapter functions** (or mark them internal so loom DCEs them). Then `dlmalloc/sbrk/memory.grow` DCE away → `--memory shared` becomes viable → one lean single-address-space core for synth (target the wasm-dist 544 B class, not tens of KB).

## Kill-criterion
"After `meld fuse` of an all-scalar-boundary component pair, `cabi_realloc` is still exported / `memory.grow` still present in the fused core" → not fixed.

Repro: `crates/gale-app-demo/dissolve.sh` on gale@feat/kiln-dissolve-libos. Relates: gale#89 (the gale-side tracking; the grow is NOT gale code — it's the canonical-ABI adapter meld should drop on fusion).

avrabe:
[...]

PR #345, opened 2026-07-14
feat(298): drop vestigial cabi_realloc + defer dead memory.grow (unblocks --memory shared --address-rebase)

avrabe:
Wires the previously-INERT `cabi_realloc_drop_provably_safe` verdict so a fully-internalised boundary drops its dead allocator — removing the last dead artifact that blocked the lean single-address-space MCU fuse (SR-50, the #326→#334→#298 MCU-dissolve arc).

## What it does
When the boundary is provably vestigial (core output, no adapters, **all lifts scalar**) AND `address_rebasing` is on AND **the allocator is provably dead**, meld drops the `cabi_realloc*` exports (allocator DCEs downstream) and defers the dead `memory.grow` to `unreachable` — so `--memory shared --address-rebase` (which otherwise hard-errors on `memory.grow`) succeeds. Every other path is byte-identical.

## Mythos discover pass — 2 findings, both fixed on this branch
- **FINDING 1 (HIGH):** module-wide grow-defer was **unsound** — the boundary verdict proves no *marshalling* realloc is needed, not that the allocator is *dead*. A scalar-interface component that allocates internally (`Vec`/`String` → `dlmalloc` → `sbrk` → `memory.grow`, reachable from a live export) would fuse-Ok then **trap at runtime** (a compile error silently downgraded to a trap). **Fix:** a call-graph reachability probe (`memory_probe::module_has_reachable_memory_grow` — roots = live exports minus the dropped `cabi_realloc*`, `start`, `ref.func`/elem targets; edges = `call`/`return_call`; fail-safe on parse error) ANDed into the gate. Now only fires when no `memory.grow` is reachable from a live non-`cabi_realloc` root; a live internal grow keeps `cabi_realloc` and the clean hard-error.
- **FINDING 2 (LOW):** tightened the export match to `cabi_realloc` / `cabi_realloc$<digits>` (was `starts_with`, which could drop a lookalike export).

## Tests (467 lib + 7 integration green; clippy + fmt clean)
`scalar_boundary_drops_realloc_and_defers_grow` (positive), `live_internal_grow_keeps_realloc_and_hard_errors` (FINDING 1), `tight_match_preserves_realloc_lookalike_export` (FINDING 2), 3 verdict-branch negatives (string lift / component output / adapter site all keep realloc + hard-error), non-shared guard, + 11 `memory_probe` unit tests (transitive dlmalloc/sbrk reachability live vs dead, elem/start roots, malformed fail-safe).

## Falsification
If the reachability gate regressed, `live_internal_grow_keeps_realloc_and_hard_errors` (fuses-Ok-then-would-trap) fails. If the drop misfired, a verdict-branch negative fails. If the match broadened, `tight_match_preserves_realloc_lookalike_export` fails.

Refs #298, #299, SR-50, SYS-8, LS-D-3.

### DIFF
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -590,6 +590,29 @@ impl Fuser {
         true
     }
 
+    /// #298 (FINDING 1) — is every input's allocator genuinely dead once the
+    /// vestigial `cabi_realloc*` exports are dropped?
+    ///
+    /// [`Self::cabi_realloc_drop_provably_safe`] proves the *interface boundary*
+    /// needs no realloc; it does NOT prove the allocator unreachable. A
+    /// scalar-interface component may still allocate internally (`Vec`/`String`/
+    /// `Box` → `dlmalloc`/`sbrk` → `memory.grow`) from a live export, and the
+    /// grow-defer is module-wide — so without this gate a live grow would be
+    /// rewritten to `unreachable` and trap at runtime instead of hard-failing
+    /// at fuse time. This ANDs an extra, strictly-stricter condition into the
+    /// drop/defer wiring: it returns `true` only when NO `memory.grow` is
+    /// reachable from any live root (all exports except the dropped
+    /// `cabi_realloc*`, plus `start` and every indirectly-referenced function)
+    /// in ANY core module of ANY input. Fail-safe: an unparseable module counts
+    /// as having a live grow (via [`memory_probe::module_has_reachable_memory_grow`]).
+    fn allocator_grow_is_dead(&self) -> bool {
+        self.components.iter().all(|comp| {
+            comp.core_modules
+                .iter()
+                .all(|module| !memory_probe::module_has_reachable_memory_grow(&module.bytes))
+        })
+    }
+
     /// The fusion pipeline proper. `self.config.memory_strategy` is a
     /// concrete strategy here — `Auto` has been resolved by the caller.
     fn fuse_with_stats_resolved(&self) -> Result<(Vec<u8>, FusionStats)> {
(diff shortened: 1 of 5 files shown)


avrabe:
## Mythos discover pass — done, two findings fixed in this PR

Adversarial pass on the export-drop + grow-defer wiring. Cleared: gating (defer flag only consulted under `address_rebasing`, which the drop already requires; the resource re-rewrite pass hardcodes `address_rebasing:false`), verdict boundary reasoning (`type_contains_pointers` covers string/list params+results), and the shared-vs-multimemory dedup namespaces.

**FINDING 1 (HIGH) — fixed.** The `defer_grow_under_rebase` flag was module-wide, but `cabi_realloc_drop_provably_safe` only proves the *boundary* needs no realloc — NOT that the allocator is dead. A scalar-interface component that allocates internally (`Vec`/`String` → `dlmalloc`/`sbrk` → `memory.grow`, reachable from a live export) would fuse-`Ok` then trap at runtime — a compile error silently downgraded to a runtime trap. Fixed with a call-graph reachability probe (`memory_probe::module_has_reachable_memory_grow`) ANDed into the gate via `allocator_grow_is_dead()`: roots = live exports minus the dropped `cabi_realloc*` + `start` + `ref.func`/elem targets; edges = `call`/`return_call`; fail-safe `true` (assume live) on any parse error. The drop/defer now fires only when every `memory.grow` is unreachable except via the dropped allocator. Regression test `live_internal_grow_keeps_realloc_and_hard_errors`.

**FINDING 2 (LOW) — fixed.** `starts_with("cabi_realloc$")` could drop a legitimately-named export; tightened to `cabi_realloc$` + non-empty ASCII digits (`is_vestigial_realloc_export_name`). Test `tight_match_preserves_realloc_lookalike_export`.

11 new `memory_probe` unit tests cover transitive dlmalloc/sbrk reachability (live and dead-under-realloc), elem/start roots, suffixed/lookalike realloc, and malformed fail-safe. 467 lib + 7 integration green; clippy + fmt clean. `mythos-pass-done`.

Issue #338, opened 2026-07-11
meld fuse silently truncates extended-const exprs beginning with global.get (drops trailing arithmetic) — corrupts global initializers + data/element offsets (PIE base+N)


avrabe:
## Summary

`meld fuse` silently **truncates** any wasm-2.0 extended-const expression that **begins with
`global.get`**: it reads only the leading `global.get` and drops the trailing arithmetic operators.
This corrupts both **global initializers** and **data/element-segment offsets** — exactly the
`__memory_base + N` / `__table_base + N` shape LLVM/emscripten emit for position-independent (PIE)
modules. The fused module still validates, so it's a **silent miscompile**.

Distinct from #152/LS-A-11, which fixed the *const-first* case (`i32.const N; global.get; i32.add`);
the `global.get`-first case was missed. Verified on meld **v0.39.0** (`9a40cff`).

## Root cause (source)

`meld-core/src/segments.rs` (`parse_const_expr_with_value`, data/elem offsets):
```rust
Operator::GlobalGet { global_index } => (ParsedConstExpr::GlobalGet(global_index), None),
```
`meld-core/src/merger.rs:~3155` (`convert_init_expr`, global initializers) has the analogous
`GlobalGet` arm. Both **return immediately** with a bare `GlobalGet`, discarding any following
operators — unlike the `I32Const`/`I64Const` arms, which fold through to `End` via
`fold_extended_const_*`. So `global.get $base; i32.const N; i32.add` is read as just
`global.get $base`.

## Reproduction (executed vs wasmtime)

**Case 1 — global initializer** `g = base + 100` (base = 1000):
```
$ meld fuse ec_comp.wasm -o fused_ec.wasm
golden (unfused component) getg = 1100
fused  (meld)              getg = 1000        # BUG: +100 dropped; validates VALID
```
The input global is `(global $g i32 global.get $base i32.const 100 i32.add)`; meld emits
`(global (;1;) i32 global.get 0)` — the `i32.const 100 i32.add` is gone.

**Case 2 — data-segment offset** `base + 8` (base = 16 → data belongs at 24):
```
golden  peek(24)=4022250974  peek(16)=0        # data at 24, correct
fused   peek(24)=0           peek(16)=...       # data landed at 16 — BUG: +8 dropped
```

## Impact

Any component whose globals/segment offsets use `global.get <imm-base> (+/-/*) const` — the standard
PIE relocation form — is silently mislaid after fusion: global values wrong, data/element segments
[...]

PR #346, opened 2026-07-15
fix(338): preserve global.get extended-const exprs — silent miscompile of PIE base+N offsets

avrabe:
`meld fuse` silently **truncated** any wasm-2.0 extended-const expression containing a `global.get` — both `global.get $base; i32.const N; i32.add` and the operand-swapped `i32.const N; global.get $base; i32.add` became just `global.get $base`, dropping the arithmetic. This corrupted **global initializers** and **data/element-segment offsets** — the position-independent `__memory_base + N` / `__table_base + N` shape. The module still validated → **silent miscompile** (SR-51). #152/LS-A-11 fixed only the const-first-no-`global.get` fold; both `global.get` cases were missed.

## Fix
The value is runtime-dependent, so it can't fold to a constant — meld now PRESERVES the full extended-const operator sequence and re-emits it with global indices remapped. New `ParsedConstExpr::ExtendedGlobalGet` + `ExtConstOp` + `read_extended_const_global_get`; the `fold_extended_const_{i32,i64}` arms return `ExtConstFold::{Value,Extended}` and route an embedded `global.get` into the preserve-and-remap path (instead of erroring → un-remapped raw-bytes fallback). Wired at both sites: `segments.rs` (data/elem offsets) + `merger.rs convert_init_expr` (global inits). Operand order preserved (`sub`/`mul` non-commutative).

**Load-bearing ordering fix** (found by the Mythos pass): the imported-global index-map population was hoisted to run BEFORE the defined-globals loop that calls `convert_init_expr` — previously the remap silently missed (`unwrap_or` kept the local index), so a multi-module fuse that shifts `__memory_base` off import-index 0 read the wrong base.

## Tests (executed vs wasmtime; 737 green; clippy + fmt clean)
- global-init `base+100`→**1100** and multi-module `100+base`→**1100** (base-shift fixture; **pre-fix read `__stack_pointer`=5100**).
- data-offset `base+N` and `N+base` land at base+N (**pre-fix truncated to base+0 / hard-error**).
- bare `global.get` + const-first fold unchanged.

## Falsification
If the preserve-and-remap regressed, `test_338_multimodule_const_first_embedded_globalget_remaps` (asserts the fused global reads the correct base, not the un-remapped 5100) fails. If truncation returned, `test_338_global_initializer_extended_const_preserved` (==1100) fails.

Refs #338, #152, SR-51, SYS-1, LS-A-11.

### DIFF
--- a/meld-core/src/merger.rs
+++ b/meld-core/src/merger.rs
@@ -1582,29 +1582,19 @@ impl Merger {
                 .push(convert_table_type(table, comp_idx, mod_idx, merged));
         }
 
-        // Merge globals (defined globals only; imported globals handled below)
-        let global_offset = merged.globals.len() as u32;
-        for (old_idx, global) in module.globals.iter().enumerate() {
-            let new_idx = merged.import_counts.global + global_offset + old_idx as u32;
-            merged.global_index_map.insert(
-                (comp_idx, mod_idx, import_global_count + old_idx as u32),
-                new_idx,
-            );
-            let init_expr = convert_init_expr(
-                &global.init_expr_bytes,
-                comp_idx,
-                mod_idx,
-                merged,
-                &global.content_type,
-            );
-            let ty = convert_global_type(global, comp_idx, mod_idx, merged);
-            merged.globals.push(MergedGlobal { ty, init_expr });
-        }
-
         // Resolve imported global indices via intra-component module_resolutions.
         // This mirrors how function imports are resolved below: if module A
         // imports a global that module B exports, map A's imported global index
         // to B's defined global's merged index.
+        //
+        // This MUST run before converting THIS module's defined-global init
+        // exprs below: an extended-const initializer may reference an imported
+        // global (`i32.const N; global.get $__memory_base; i32.add`, #338), and
+        // `convert_init_expr` remaps that global through `global_index_map`. If
+        // the imported-global entries were populated only afterwards (the prior
+        // ordering), the remap silently missed and emitted the un-remapped local
+        // index — reading the wrong global whenever fusion shifted the import
+        // off index 0.
         {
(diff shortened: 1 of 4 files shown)


avrabe:
## Mythos discover pass — done (two passes; one incomplete-fix caught + fixed)

**Pass 1** on the initial `global.get`-first fix: the fix itself was sound across all angles (op whitelist with no silent fall-through, correct remap keying, operand order preserved, `value=None` safely errors on the rebasing path, valid re-encode). But it flagged a **MEDIUM incomplete fix**: the operand-swapped sibling `i32.const N; global.get $base; i32.add` (const-FIRST with an embedded `global.get`) was STILL a silent miscompile in global initializers — the const-first fold path errored on the embedded `global.get` and the merger fell back to `ConstExpr::raw(<un-remapped bytes>)`, so in multi-module fusion where `__memory_base`'s merged import index shifts off 0 it read the wrong base. The new test masked it (single-module keeps base at index 0).

**Fix (this PR):** routed the embedded-`global.get` case from the `fold_extended_const_{i32,i64}` arms into the same preserve-and-remap path (`ExtConstFold::Extended`), and — the actual root cause — **hoisted the imported-global index-map population before `convert_init_expr`** so the remap no longer silently misses. Added a multi-module base-shift test (`__memory_base` at a non-zero merged index; pre-fix read `__stack_pointer`=5100, post-fix reads 1100) + a data-offset `N+base` test (pre-fix hard-errored).

**Pass 2 / re-verification:** full suite re-run on the rebased branch — 36 test binaries, 0 failures; clippy + fmt clean; `rivet validate` PASS. Both operand orders, both sites (global inits + data/elem offsets), single- and multi-module, all covered. `mythos-pass-done`.

Issue #341, opened 2026-07-11
fuse --reproducible: output still depends on the input file PATH (identical bytes at different paths -> different sha)


avrabe:
Follow-up to #325 (thank you for `--reproducible` in v0.38.0 — it removes the random-UUID + wall-clock nondeterminism, which was the dominant source). Verifying it against a real component surfaced a **residual**: `--reproducible` output still depends on the input file's **path**, so identical *bytes* at two different paths fuse to different sha256s.

### Repro (meld v0.39.0, macOS arm64)
```
# identical bytes, two paths
mkdir -p /tmp/pa /tmp/pb
cp falcon-flight-v1.112.wasm /tmp/pa/falcon.wasm
cp falcon-flight-v1.112.wasm /tmp/pb/falcon.wasm   # byte-identical to /tmp/pa/falcon.wasm

meld fuse /tmp/pa/falcon.wasm -o a.wasm --reproducible
meld fuse /tmp/pb/falcon.wasm -o b.wasm --reproducible
shasum -a256 a.wasm b.wasm
# f0cf6a82...  a.wasm
# 1cc0f5c2...  b.wasm     <- differ, though the inputs are byte-identical
```

Controls (both as expected):
- **same path, separate invocations → identical** (`f0cf6a82` both) ✅
- **different path, same content → differ** ❌ (above)
- even the *same filename* in a different directory differs, so it's the full path, not just the basename.

### Why it matters (jess / Pixhawk 6X-RT)
We fuse the falcon flight core and want the fused artifact to be a **re-verifiable, sigil-signable** attestation for the FIRST-FLASH image set — two parties (or two CI checkouts / temp dirs) fusing the same component should get the same sha. With path-dependence, the fused sha isn't portable, so we still can't key an attestation on it (we currently key our scry gate on soundness invariants instead).

### Likely cause / suggested fix
Something path-derived is being embedded (a producers/source/name custom section, or the content-hash that seeds the attestation id is computed over a buffer that includes the source path). Deriving the id purely from the **module content** (+ `SOURCE_DATE_EPOCH`) and stripping/normalizing any embedded input path would make identical bytes → identical output regardless of location, completing the #325 intent.

Happy to test a fix against our real falcon core. No urgency — our gate keys on invariants meanwhile.

avrabe:
[...]

Issue #325, opened 2026-07-08
fuse: output is byte-nondeterministic — identical input yields a different sha256 each run


avrabe:
## What

`meld fuse` (v0.33.0) produces a **different `sha256` on every run** for the *same* input component, even though the output size and semantics are identical. For a verification toolchain this breaks reproducible-build / supply-chain attestation (the fused core is what scry, loom, witness and the sigil/rivet evidence chain all key on).

## Repro

```sh
for i in 1 2 3; do meld fuse falcon-flight-v1.112.wasm -o f$i.wasm; done
shasum -a256 f1.wasm f2.wasm f3.wasm
```

Result — same size (101938 B), three different hashes:

```
9e1265bb0bd8a19d…  f1.wasm
cef3c0c189250a02…  f2.wasm
5dd1982a9b999ea8…  f3.wasm
```

Input: released `falcon-flight-v1.112.wasm` (component sha256 `6112183070105d33…`, 100086 B). meld v0.33.0, macOS arm64.

## Likely cause

A `HashMap`/`HashSet` iteration order leaking into emission order (Rust's default `RandomState` reseeds per process). Common fixes: iterate a `BTreeMap`/sorted keys at every emission site, or use a fixed-seed hasher for the fusion tables.

## Impact / why it matters here

jess ([pulseengine/jess](https://github.com/pulseengine/jess)) is standing up a scry sound-analysis CI gate on the meld-fused falcon core (the sound-static-analysis leg of the V, complementing relay's source MC/DC — cf. meld#302). scry's own output records `module_sha256`, and a reproducible fused artifact is a prerequisite for a stable, attestable gate. We worked around it by keying the gate on scry's **soundness invariants** rather than the fused hash, but a deterministic fusion output would let the whole chain attest the byte-exact artifact. No rush — flagging with exact repro.

Thanks!

avrabe:
Shipped in v0.38.0 (`--reproducible`, PR #327). Closing — the fix + regression test are on main and released.

PR #348, opened 2026-07-15
fix(341): --reproducible output no longer depends on input file path

avrabe:
#325's `--reproducible` removed the random-UUID/wall-clock nondeterminism, but a **residual** remained: the output still depended on the input file **path** — byte-identical components at different paths fused to different sha256s. Two CI checkouts / temp dirs fusing the same component disagreed, breaking re-verifiable / sigil-signable attestation for the jess/Pixhawk first-flash image set.

## Root cause
The caller-supplied name (the CLI passes the full input path) leaked into the output at **three** sites:
1. the default `FusionAttestationBuilder` input descriptor (`lib.rs`),
2. the wsc attestation `InputArtifact` (`feature=attestation` path),
3. `provenance::build`'s `component_id` (`provenance.rs`, emitted by default).

## Fix
Under `reproducible`, the input name is the **positional `component-{index}`** (path- and filename-independent) instead of the caller's name; the input's content stays pinned by `original_hash`/`hash`. Non-reproducible mode keeps the human-friendly name.

## Test
`test_reproducible_attestation_is_byte_stable` extended: byte-identical input under `/tmp/pa/falcon.wasm` vs `/tmp/pb/falcon.wasm` now fuses **identically** under reproducible (was different); the non-reproducible control still differs (name not silently dropped).

## Falsification
If the path leaked again, the new `pa == pb` assertion fails. If reproducible fusion stopped being byte-stable across runs, the pre-existing `a == b` assertion fails.

467 lib + full suite green; `--features attestation` compiles; clippy + fmt clean.

Refs #341, #325.

### DIFF
--- a/meld-core/src/lib.rs
+++ b/meld-core/src/lib.rs
@@ -842,7 +842,12 @@ impl Fuser {
         }
 
         if self.config.component_provenance {
-            let provenance = provenance::build(&merged, &self.components, &output_without_extras);
+            let provenance = provenance::build(
+                &merged,
+                &self.components,
+                &output_without_extras,
+                self.config.reproducible,
+            );
             // SCPV v3 binary payload (#313 / scry#63) — infallible encode.
             extra_sections.push((provenance::SECTION_NAME, provenance.to_bytes()));
         }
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos discover pass (manual) — done

Bounded determinism fix; provenance.rs is Tier-5 so reviewed adversarially. The change adds a `reproducible` bool that, when set, replaces the caller-supplied component name with the positional `component-{index}` at the three sites that embed it in output bytes (two attestation paths + provenance `component_id`). Checked:
- **No correctness change to fusion or the provenance data structure** — only a metadata *string* changes, and only under the `--reproducible` flag. Non-reproducible output is byte-identical to before.
- **Content identity preserved** — every input still carries `original_hash`/`hash` (content sha256), so consumers (scry provenance, sigil attestation) still uniquely identify each input by content; the positional id is a valid per-component unique key.
- **All three leak sites covered** — grep-confirmed no other `component.name`→output path (the remaining `.name` uses are logging (P3-async info line) and import/export symbol names, which are content-derived, not caller paths).
- **No false negatives** — this is a determinism fix, not a safety gate; there's no corruption channel to miss.

467 lib + full suite green; `--features attestation` compiles; clippy + fmt clean. `mythos-pass-done`.

Issue #351, opened 2026-07-15
meld --memory shared --address-rebase skips R_WASM_MEMORY_ADDR_SLEB relocations: a rebased module's i32.const data pointer stays stale -> silent cross-component aliasing (gap in #340)


avrabe:
## Summary

`meld fuse --memory shared --address-rebase` rebases a module's data segment and its memarg
load/store **offsets** (`R_WASM_MEMORY_ADDR_LEB`) but **does not apply `R_WASM_MEMORY_ADDR_SLEB`
relocations** — the ones that materialize a data address as an `i32.const` in code. So when a rebased
module takes a **pointer** to its own data (returns `&buf[0]`, passes a buffer address to memcpy,
returns a string/struct pointer), the `i32.const` stays at the *pre-rebase* address, pointing into the
**other** fused component's memory region. Silent cross-component pointer aliasing; the fused module
validates cleanly.

Distinct from #326 (runtime-computed operands that carry no relocation) and #339 (absolute addresses
used as values with **no** reloc): here a relocation *exists* (`MemoryAddrSleb`) and is simply dropped
by the #340 reloc-driven rebasing. Verified on meld **0.41.1**.

## Reproduction (executed vs wasmtime)

Two C modules, each with a static array at wasm-ld's default base `0x10000`, built with
`-Wl,--emit-relocs`, componentized, fused with `--memory shared --address-rebase`. In the fused core
module `dataB` is rebased to `0x30000` (196608), `dataA` stays at `0x10000`:

| export | fused result | expected |
|---|---|---|
| `get-b(0)` (`i32.load offset=…` — LEB, rebased) | `0xB0B0B0B0` | `0xB0B0B0B0` ✓ |
| `ptr-b()` (returns `&dataB[0]` — `i32.const`, SLEB) | **`65536`** (0x10000, dataA's region) | `196608` (0x30000) |

So `ptr-b` returns a pointer 128 KB low — into `dataA`. A guest dereferencing it reads/writes the
*other* component's buffer.

## Root cause (source)

`meld-core/src/reloc.rs` recognizes `MemoryAddrSleb` as a memory-address reloc
(`RelocType::MemoryAddrSleb` code 4; `is_memory_addr` includes it at ~:136-140), and the input
carries it — `b_core.wasm`'s `reloc.CODE` = 3 entries `MemoryAddrLeb@13, MemoryAddrLeb@33,
[...]

Issue #3, opened 2026-02-23
Wrong compiler output


cpetig:
- To me the generated instructions look like arm32 instructions (armv5), not thumb-2 (armv7m)
- The calculator example compiles to an add and a move; the return instruction is not in thumb-2 format (as far as I can tell)
- The pid example compiles to only moves as the floating point instructions aren't recognized.
- Only the first functions gets compiled.

PR #352, opened 2026-07-16
fix(351): hard-fail on stale reloc.CODE offsets instead of silent miscompile

avrabe:
## What

Backstop tier of the #351 fix: meld now **hard-fails** with `MisalignedReloc`
when a `reloc.CODE` memory-address site does not land on a rebasable operator
immediate, instead of silently emitting a miscompiled module.

## Why (grounded)

avrabe reported (#351) that `meld fuse --memory shared --address-rebase` left a
`ptr = &data[0]` `i32.const` un-rebased — reproduced on v0.41.1 with their exact
bytes: `wasmtime --invoke ptr-b` returned **65536** (aliasing component-a)
instead of the rebased **196608**.

**Root cause** (byte-level): the producer (clang 22.1.4 + wit-component 0.245.1)
relaxed address immediates from 5-byte-padded to minimal LEB128 but did **not**
rewrite `reloc.CODE` offsets, so each site drifts **+2 bytes per preceding
memory-address reloc**. The memargs survive (drift stays inside their wide
immediate window); the trailing `SLEB@42` drifts past `ptr-b`'s `i32.const`
range `[37,41)` → `addr_reloc` never set → silently not rebased. Upstream
producer defect filed as pulseengine/wasm-tools#3.

## The fix (layered — this is tier 1)

meld verifies every `R_WASM_MEMORY_ADDR_*` site lands on a rebasable immediate;
if any is misaligned (stale), it hard-fails rather than corrupt the shared
address space — upholding the "never silently corrupt" invariant #326 rests on.
**Drift-tolerant *correct* rebasing** (so these inputs fuse rather than fail) is
the tier-2 follow-up.

## Falsification / tests

- `test_351_stale_reloc_offsets_hard_error` — fuses avrabe's exact components
  (`tests/reloc351/{a,b}.wasm`, b sha256 `481c36…d3a3`); asserts
  `MisalignedReloc { offset: 42 }`. Pre-fix this silently returned 65536.
- `reloc::first_misaligned_code_reloc_detects_drift` — unit oracle: aligned site
  → ok; drifted onto `end` / past the section → caught.
- Negative: `test_326_reloc_const_rebasing_end_to_end` (consistent relocs) still
  rebases correctly; a real clang-22.1.8 consistent-reloc build still fuses
  (`ptr_b = 196608`) — the backstop does not over-reject valid modules.

Local: `cargo fmt --check`, `cargo clippy`, full `meld-core` test suite green.
Traceability: SR-53 (mitigates LS-D-1).

Refs #351, pulseengine/wasm-tools#3


avrabe:
## Mythos delta-pass — findings

Two clean-room adversarial discover passes (fresh-context agents, one per changed
Tier-5 file), per `scripts/mythos/discover.md`. **Verdict: no reportable bug**
meeting the oracle bar (failing Kani + concrete PoC). Both passes are documented
below; the honest limitation they surfaced is now corrected in-tree (commit
`cffdfd7`).

### `meld-core/src/reloc.rs` — `first_misaligned_code_reloc` / `operator_is_rebasable`

CLEAN. Hypotheses tested & refuted with evidence:
- **Coordinate-frame consistency (H2):** backstop, rewriter (`extract_function_body`),
  and reloc offsets all use the same code-section-content frame
  (`CodeSectionReader` over `&bytes[start..end]`, base 0). Consistent.
- **Multi-function over-extension (H3):** the `ends` chain over-extends only each
  body's *last* operator, which is always the non-rebasable `End` → never added to
  `rebasable_ranges`. Inert.
- **`operator_is_rebasable` completeness (H4):** its set is exactly what the
  rewriter rebases (consts + non-SIMD/non-atomic loads/stores). Excluded SIMD/atomic
  memargs → hard-fail = the *safe* verdict (rewriter couldn't rebase them either).
- **Degenerate/boundary (H5):** empty offsets early-return; `r==op_end` excluded;
  `r==len` caught; no panic/overflow.
- **REAL LIMITATION (H1), not a reportable bug:** the backstop checks "site in *some*
  rebasable range", not "the *correct* one". A site drifted from operator A *into an
  adjacent* rebasable operator B is accepted (`None`) and the rewriter rebases the
  wrong-but-rebasable immediate. This is undecidable from `(code, offsets)` alone
  (no fix flips a Kani harness), so it fails the oracle bar. **Now documented** in the
  fn docs + SR-53 scope + pinned by `first_misaligned_code_reloc_adjacent_drift_is_a_known_gap`.
  Closed by the drift-tolerant tier-2 fix (symbol/value-confirmed matching).

### `meld-core/src/merger.rs` — backstop wiring

CLEAN. Hypotheses tested & refuted:
- **Guard gap (H1):** `code_section_range` is `Some` for every module with a code
  section; a reloc.CODE-carrying module with no code section has zero bodies →
  nothing to mis-rebase.
- **Frame mismatch (H2):** all four consumers share the content frame; pinned by
  `provenance.rs::code_section_ranges_rebased_to_content_start` and the live
  `test_351_…` firing at offset 42.
- **Bypass call sites (H3, highest-value):** every other `build_index_maps_for_module`
  call passes `code_addr_relocs = None` (resource re-rewrite 1269, lib.rs 1622,
  p3_async 1369, p3_bridge 403) → no reloc-driven rebasing there, so no backstop
  needed. The sole `Some` path is gated. *(Noted for future, out of scope here: the
  resource-import re-rewrite at merger.rs:1247-1294 re-extracts bodies with
  `address_rebasing=false` — a pre-existing #326-class rebasing-drop concern if a
  rebased module also has redirected `[resource-*]` imports; not demonstrable, not #351.)*
- **base 0 (H4):** backstop runs regardless of `memory_base_offset`; validating the
  base-0 module too is fail-closed, not unsafe.
- **Panic/indices (H5):** slice bounds in-bounds by construction; identifiers correct;
  `CodeParse` propagates fail-closed.

Adding `mythos-pass-done`.

avrabe:
Strategic note: #353 tracks evaluating PIC / shared-everything dynamic linking as an input ABI — if that spike succeeds it likely **supersedes this PR's tier-2** (drift-tolerant reloc rebasing), since PIC inputs have no absolute address literals to rebase. This tier-1 backstop is orthogonal and still the right near-term safety net for `--emit-relocs` users.

PR #357, opened 2026-07-16
fix(355): --component lifts bare world exports with real signatures

### LINKED ISSUE #355: fuse --component emits an invalid component: every export lifted with an empty (func) type, so non-empty signatures fail wasm-tools validate
## Summary

`meld fuse --component` emits an **invalid component** for any export whose function signature is non-empty: every fused export is lifted with an empty component-level function type `(func)`, regardless of the underlying core signature. `wasm-tools validate` rejects the output; the fused component cannot be instantiated. Silent — fusion prints "Fusion complete!" with no warning. The default core-module output for the same inputs is correct.

## Repro

Inputs: two well-formed reloc-bearing components that validate and execute standalone (e.g. the two in `harness/fixtures/meld_sleb/`, exports `get-b: [i32]→[i32]`, `set-b: [i32,i32]→[]`, `ptr-b: []→[i32]`).

```
meld fuse --component a.wasm b.wasm -o out.wasm      # exit 0, "Fusion complete!"
wasm-tools validate --features all out.wasm
  → error: lowered parameter types `[]` do not match parameter types `[I32]` of core function 0
```
(A value-returning export gives the dual: `lowered result types [] do not match result types [I32]`.)

The fused module's **core** types are correct — `(type (func (param i32) (result i32)))`, `(type (func (result i32)))` — but the component-level lift is emitted as `(type (;0;) (func))` + `(func (canon lift (core func 0)))`, i.e. the component export type is not derived from the core function signature / the original component export type.

Reproduces under `--memory multi`, `auto`, and `shared` — independent of memory strategy. `--component` is documented as a plain option with no experimental caveat.

## Fix direction

Derive each fused export's component-level function type from the core function signature (or carry through the original component export type) instead of emitting `(func)`.

avrabe:
## What

Fixes #355 — `fuse --component` emitted an **invalid** component for any export
with a non-empty signature. Reproduced on avrabe's exact fixtures.

## Root cause (three latent defects)

Bare *world* exports (`world root { export get-b: func(i: s32) -> u32 }`) took a
path that hardcoded an empty `(func)` type. Digging in surfaced three bugs:

1. **Parser (`ComponentFuncDef`):** exporting a `Func` binds a new
   component-function index (an export alias); wit-component **interleaves** these
   with `canon lift`s. The parser recorded them nowhere, so the func index space
   was silently compacted and `component_func_defs[export.index]` mis-resolved
   (`get-b` ✓, `set-b`/`ptr-b` shifted). Fixed by recording
   `ComponentFuncDef::ExportAlias`.
2. **Wrapper no-result func:** `func_enc.result(None)` was never called for a
   void-returning export (`set-b`), leaving the type **unterminated** → decoder
   EOF. (Latent until a no-result export routed through this helper.)
3. **Wrapper export-alias skip:** the bare-export emitter advanced the
   component-func index by 1 per export, but each export also binds an alias
   index — so every export referenced the **previous** export's alias. Fixed to
   advance by 2.

## Result

`get-b: func(i:s32)->u32`, `set-b: func(i:s32,v:u32)`, `ptr-b: func()->u32` — all
correct, output validates. The standard `{iface}#{func}` interface path is
untouched.

## Tests / falsification

`component_bare_export_355.rs` (single + two-component) on `tests/reloc351/{a,b}.wasm`:
asserts a **valid** component (`wasmparser::Validator`, all features) **and** the
correct per-export **arity** (get-b (1,1), set-b (2,0), ptr-b (0,1)) — the first
catches the empty-type + no-result bugs, the second catches the index shift.

Full `meld-core` suite green (no regression from the parser index-space change);
fmt + clippy clean. SR-54 (verified); CHANGELOG. Tier-5 (parser/component_wrap/
p3_stream) → Mythos discover pass to follow on this PR.

Refs #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- **`fuse --component` emitted an invalid component for non-empty export
+  signatures (#355).** Bare *world* exports (e.g. `world root { export get-b:
+  func(i: s32) -> u32 }`) were all lifted with an empty `(func)` component type,
+  so `wasm-tools validate` rejected the output ("lowered parameter types [] do
+  not match the core function") — silently, after printing "Fusion complete!".
+  Three latent defects underlay it: the parser did not record func export-aliases
+  (compacting the component-function index space so lookups mis-resolved); the
+  wrapper never terminated a no-result func type (truncating it); and the
+  bare-export emitter advanced the func index by 1 per export while each export
+  also binds an export-alias, shifting every export onto the previous one's type.
+  Now each export lifts with its real signature and the output validates.
+  **Falsification:** `component_bare_export_355.rs` fuses the exact fixtures and
+  asserts a valid component + correct per-export arities (get-b (1,1), set-b
+  (2,0), ptr-b (0,1)); the standard interface-export path is unaffected.
+
 ## [0.41.2] - 2026-07-16
 
 Soundness patch: stale relocation metadata can no longer silently miscompile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,15 @@ All notable changes to this project will be documented in this file.
   **Falsification:** `component_bare_export_355.rs` fuses the exact fixtures and
   asserts a valid component + correct per-export arities (get-b (1,1), set-b
   (2,0), ptr-b (0,1)); the standard interface-export path is unaffected.
+- **A component exporting ≥2 interfaces bound the wrong instance to the 2nd+
+  export (found by the #355 Mythos pass).** The interface-export loop advanced
+  the component-instance index by 1 per exported interface, but exporting an
+  instance also binds an export-alias index — so the second and later interface
+  exports referenced the previous interface's alias and silently exported the
+  wrong functions (the output still validated). Now advances by 2.
+  **Falsification:** `component_multi_interface_instance_idx.rs` fuses a
+  two-interface component and asserts each exported interface resolves to its own
+  functions.
 
 ## [0.41.2] - 2026-07-16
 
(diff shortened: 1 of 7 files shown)


avrabe:
## Mythos delta-pass — findings

Three clean-room adversarial discover passes (fresh-context agents), one per
changed Tier-5 file. **Verdict: no reportable bug on the #355 change itself; one
real sibling bug found and fixed in the same PR.**

### `parser.rs` (ComponentFuncDef::ExportAlias) — CLEAN
Every consumer of `component_func_defs` uses variant-specific `match`/`if let`;
none assume the (previously compacted) length or iterate expecting only the old
three variants. The change is **order-preserving by construction** (entries
pushed in binary section order), so alignment holds for any section ordering, not
just the fixture. Notably it also **improves `resolver.rs` correctness** — its
func-index→name maps were keyed by the compacted space pre-fix and could
mis-resolve interleaved cases.

### `component_wrap.rs` — CLEAN on the #355 change; **found a sibling bug (fixed)**
The `+= 2`, post-return, `result(None)`, and type-carry changes all verified
correct (mixed interface+bare, string-returning, void, fallback cases).
**But the same index-desync exists in the *interface*-export loop:** it advanced
`component_instance_idx` by 1 while exporting an instance also binds an
export-alias index — so a component exporting **≥2 interfaces** silently bound
the 2nd+ export to the previous interface's alias, exporting the wrong functions
(the output still validated — silent; UCA-CP-1). Confirmed with a Kani harness +
a differential PoC. **Fixed in commit `e115719`** (`+= 2`) with a regression test
(`component_multi_interface_instance_idx.rs`).
*Noted (not reported — no reachable PoC):* `type_remap` is keyed only by
`source_type_idx`; a defensive key of `(component, source_type_idx)` would harden
against a future cross-component collision.

### `p3_stream.rs` (new ExportAlias match arm) — CLEAN
Unreachable for the #355 fixtures (they hit `Lift` directly); when hit it follows
one hop to the real `Lift`/`Import` (no cycle/panic/OOB — single `.get()`).
Empty-result only feeds the #142 stream-mismatch *diagnostic*, which has a
role-list fallback, so no fused-byte divergence is producible.

Full `meld-core` suite green after both commits. Adding `mythos-pass-done`.

Issue #353, opened 2026-07-16
Evaluate PIC / shared-everything dynamic linking as an input ABI (dissolves the #351 reloc-drift class) — revisit ADR-6


avrabe:
## Summary

Evaluate **PIC / shared-everything dynamic linking** as an *input ABI* for meld,
as an alternative (or complement) to the relocatable-inputs contract decided in
ADR-6. Raised by Christof re: #352 — with PIC inputs, address "rebasing" becomes
"set one `__memory_base` global per module", which **structurally dissolves the
entire #351/#340 stale-`reloc.CODE` failure class**: there are no absolute
`i32.const` address literals to rewrite, so there is nothing for a producer's
LEB-relaxation to desync.

Refs: [SharedEverythingDynamicLinking.md](https://github.com/WebAssembly/component-model/blob/main/design/mvp/examples/SharedEverythingDynamicLinking.md),
[DynamicLinking.md](https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md).

## Grounded findings (verified today — wasm-tools 1.243.0, wac 0.6.0, clang 22.1.8)

1. **PIC eliminates absolute addresses.** `clang --target=wasm32 -fPIC -O2 -c` +
   `wasm-ld --experimental-pic -shared` turns `dataB` into
   `(data (global.get $__memory_base) …)` and `ptr_b` into
   `global.get $__memory_base` + offset. No `i32.const <addr>` literals exist.
2. **An AOT linker exists and works:** `wasm-tools component link` (and `wac
   0.6.0`) link two dylink.0 modules into a **valid** component — one `$main`
   module owns a shared memory and exports it; each dylib is instantiated
   `(with "env" …)` with that memory + a **disjoint `__memory_base`**; an
   `$__init` module applies relocations at instantiation. This is the
   SharedEverythingDynamicLinking model, working out of the box.
3. **This is exactly the path the #326 spike (ADR-6) mis-scoped.** The spike
   concluded "PIC excluded" — but it tested `wasm-tools component **new**`, which
   rejects a dylink core (reconfirmed today: `error: failed to decode world from
   module`; the CM forbids the `env.memory`/`__memory_base` non-function
   imports). `component **link**` is purpose-built for these and **accepts** the
   same core (reconfirmed). ADR-6's exclusion is really "excluded *from the
   `component new` pipeline*", not "excluded".

## Alignment with PulseEngine goals

[...]

PR #359, opened 2026-07-17
refactor(adr7): extract the address-strategy seam from the merger (path-H inc 1)

avrabe:
Executes **ADR-7 path-H, increment 1** (#354): extract the per-module
address/memory strategy resolution out of `Merger::merge` into a dedicated
`address_strategy` module — **behavior-preservingly**. This is the pluggable seam
the decided architecture hangs off.

## What

- `address_strategy.rs` — `AddressPlan` + `resolve_address_plan`, holding the full
  decision + validation (ADR-6 path-F `MissingRelocMetadata`, memory64 reject,
  the #351 `MisalignedReloc` backstop). `has_direct_memory_access` is passed as a
  **closure** so the original lazy short-circuit (only a non-zero-base, no-reloc
  module pays the probe) is preserved exactly.
- `merger.rs` — the ~50-line reloc-consumption block replaced with one call.
- 4 focused unit tests (rebasing-off / base-0 / no-reloc-direct → hard-fail /
  bulk-only → empty, plus closure-laziness).

## Scope of `AddressPlan` (does *not* affect multi-memory + FACT fusion)

An `AddressPlan` carries **only** which relocation sites to shift for a non-zero
shared-memory base — nothing else. Under **multi-memory**, each component keeps
its own linear memory at its own base, so there are no sites to shift and the
plan is **empty**. An empty plan means *"no address rebasing needed here"* — it
is **not** a disablement of fusion, and it does **not** touch cross-component
adapter (**FACT**) generation, transcoding, or any other part of the merge. Those
run identically regardless of what this seam resolves. Multi-memory + FACT stays
the generic, first-class fusion path (ADR-7's canonical / `open` profile);
`multi_memory` fixtures remain green (6/6).

The shared-rebase strategy now lives here. **Static-base PIC (#353)** and the
**per-boundary call-lowering seam** are the next increments — they plug in here
without touching the merger's hot path.

## No behavior change

Canonical gating fixtures (`test_326_reloc_const_rebasing_end_to_end`,
`test_address_rebasing_end_to_end`) + `drop_realloc` + full `meld-core` suite all
green; fmt + clippy clean. Pure structural extraction.

Tier-5 (merger.rs + new fusion-critical module) → Mythos discover pass to follow.

Refs #354 (ADR-7)

### DIFF
--- a/meld-core/src/address_strategy.rs
+++ b/meld-core/src/address_strategy.rs
@@ -1,0 +1,209 @@
+//! Address / memory strategy seam (ADR-7 path-H, increment 1).
+//!
+//! meld places each fused module's linear memory somewhere in the output's
+//! address space, and how a module's *absolute addresses* are handled depends
+//! on the memory strategy in force:
+//!
+//! - **multi-memory** — each component keeps its own memory; nothing to rebase.
+//! - **shared-rebase** (ADR-6 path-D) — the module is placed at a shared-memory
+//!   base and its `reloc.CODE`/`reloc.DATA` memory-address sites are shifted by
+//!   that base (the #326→#340 reloc consumer, with the #351 stale-reloc backstop).
+//! - **static-base PIC** (ADR-7 / #353, future) — PIC inputs compute addresses
+//!   from `__memory_base`; meld folds a disjoint constant base per module. No
+//!   absolute-address relocation sites exist, so this strategy will short-circuit
+//!   the reloc machinery entirely.
+//!
+//! This module is the single home for that decision. Increment 1 extracts the
+//! existing shared-rebase resolution here **behavior-preservingly** — same
+//! validation (`MissingRelocMetadata`, memory64 reject, `MisalignedReloc`), same
+//! short-circuits — so the pluggable variants above can be added later without
+//! touching the merger's hot path. The canonical/multi path is unchanged (it
+//! resolves to an empty plan).
+
+use crate::{Error, Result};
+
+/// What the address strategy resolved for one module: which relocation sites the
+/// rewriter must rebase. An empty plan (`code_addr_relocs: None`, no data
+/// entries) means "no address rebasing for this module" (multi-memory, or shared
+/// memory with a zero base / no reloc metadata).
+#[derive(Debug, Default)]
+pub struct AddressPlan {
+    /// Code-section byte offsets of `R_WASM_MEMORY_ADDR_*` sites to rebase
+    /// ([`crate::reloc::RelocInfo::code_memory_addr_offsets`]). `None` = the
+    /// legacy blanket path (a no-op when the base is 0); `Some` = reloc-driven.
+    pub code_addr_relocs: Option<std::collections::HashSet<u32>>,
+    /// `reloc.DATA` `R_WASM_MEMORY_ADDR_I32` inline pointers to rebase in the
+    /// data segments' payloads.
+    pub data_addr_relocs: Vec<crate::reloc::RelocEntry>,
+}
+
+/// Resolve the address plan for one module placed at `memory_base_offset`.
+///
+/// `has_direct_memory_access` is evaluated **lazily** (only when a non-zero base
+/// carries no reloc metadata) so a component that never needs the check pays
+/// nothing and cannot fail on it — preserving the merger's original
+/// short-circuit exactly.
+///
+/// Behaviour (shared-rebase strategy; unchanged from the pre-extraction merger):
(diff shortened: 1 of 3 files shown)


avrabe:
## Mythos delta-pass — finding

Clean-room adversarial pass focused on the one thing that matters for a
behavior-preserving refactor: **is `resolve_address_plan` semantically identical
to the inline block it replaced?**

**Verdict: CLEAN — behavior-identical**, verified branch-by-branch against the
removed merger block:
- **Short-circuit / laziness (the risk):** the `has_direct_memory_access` closure
  is invoked in the same position of the same `&&` chain, so
  `module_has_direct_memory_access` is called only when `base != 0 && !has_reloc`,
  at most once, with `?` propagated identically. Pinned by two dedicated laziness
  unit tests.
- **Return values:** every branch (rebasing-off / base-0 / no-reloc / has-reloc /
  memory64 / misaligned) returns the same `code_addr_relocs` and `data_addr_relocs`
  (the old `let mut … = empty` default is mirrored by `AddressPlan::default()`).
- **Errors:** `MissingRelocMetadata` / `UnsupportedFeature` (memory64) /
  `MisalignedReloc` fire under identical conditions with identical fields; the
  memory64 format string is byte-for-byte identical.
- **Only diff:** `component_display_name` is now evaluated eagerly per module (one
  redundant `String` alloc) — pure/total (`.get()`, no panic, no side effects), so
  a performance detail, not observable behavior.

No divergent input exists → no oracle can fail → nothing reported. Gating fixtures
(`test_326_reloc_const_rebasing_end_to_end`, `test_address_rebasing_end_to_end`,
`test_351_…`, `test_326_shared_rebase_without_relocs_hard_errors`) + drop_realloc +
address_strategy unit tests all green.

Adding `mythos-pass-done`.

avrabe:
Good question — short answer: **no, nothing about the generic multi-memory + FACT path is stopped or changed.** "Empty plan" is a no-op, not a disablement. Let me be precise about the wording.

**`AddressPlan` is *only* about address rebasing** — which `reloc.CODE`/`reloc.DATA` `R_WASM_MEMORY_ADDR_*` sites the rewriter must shift when a module is placed at a **non-zero shared-memory base**. That's its entire scope:
```rust
pub struct AddressPlan {
    pub code_addr_relocs: Option<HashSet<u32>>,   // code sites to rebase
    pub data_addr_relocs: Vec<RelocEntry>,        // data pointers to rebase
}
```

**Multi-memory has no addresses to rebase** — each component keeps its *own* linear memory at its own base, so no absolute address ever collides with another component. There's simply nothing to shift → an empty plan. That's exactly what multi-memory did before this PR; the refactor just gives that "no rebasing needed" case a name. `resolve_address_plan` returns the empty plan whenever `address_rebasing` is off (which multi-memory always is).

**FACT is a *separate* subsystem and is completely untouched.** This refactor has **zero** references to `adapter`/FACT in either changed file (grep-confirmed). Cross-component calls under multi-memory still generate the FACT trampolines (`cabi_realloc` in the callee's memory + `memory.copy` between memories) exactly as before — the adapter subsystem is wired from `lib.rs`, not the merger, and this PR doesn't go near it.

**Verified:** the multi-memory test suite is green (6/6), and the whole change is Mythos-reviewed as *behaviour-identical* to the pre-refactor block (it's a pure extraction).

**Bigger picture (ADR-7 / #354):** the decided architecture deliberately keeps **multi-memory + canonical FACT as the first-class, universal path** — it's the `open/ecosystem` profile and the canonical fallback that preserves mix-and-match with any component. The symmetric/PIC "fast lane" that this seam will later enable is strictly **additive and opt-in** — it never removes or degrades the generic multi-memory + FACT fusion. That non-negotiable is written into ADR-7 as a hard requirement.

avrabe:
Fixed the misleading phrasing in both places — the module doc (`address_strategy.rs`) and this PR description now spell out the scope explicitly:

> An `AddressPlan` carries **only** which relocation sites to shift for a non-zero shared-memory base. Under multi-memory each component keeps its own memory at its own base, so the plan is **empty** — meaning *"no address rebasing needed here"*, **not** that fusion is disabled. It does **not** touch cross-component adapter (FACT) generation, transcoding, or any other part of the merge.

So to answer directly: **no**, multi-memory + FACT fusion into a core module is not stopped. That path is unchanged by this seam (`multi_memory` fixtures 6/6 green); the "empty plan" is just the absolute-address rebaser having nothing to do when memories aren't shared. This PR is a pure structural extraction of the shared-rebase decision — FACT/transcoding code isn't referenced in any changed file.

PR #360, opened 2026-07-17
refactor(adr7): extract the per-boundary call-lowering seam (path-H inc 2)

avrabe:
Executes **ADR-7 path-H, increment 2** (#354): lift the *"how is this
cross-component call boundary lowered"* decision out of
`FactStyleGenerator::generate_adapter` into a dedicated `adapter::call_lowering`
module. The call-lowering counterpart of the address-strategy seam (#359, inc 1).

## What

- **`adapter/call_lowering.rs`** — `BoundaryFacts` + `CallLoweringPlan` +
  `resolve_call_lowering_plan`. The three-way class dispatch
  (`crosses_memory` / `needs_transcoding` → `Direct` / `MemoryCopy` /
  `Transcode`) **and** the #304 inline-eligibility guard are resolved together
  (they are coupled — see the module doc). 9 unit tests cover the class + inline
  truth table and the hard-fail below.
- **`adapter/fact.rs`** — `generate_adapter` resolves the plan, `match`es on the
  class for body generation, and reads `inline_eligible` for the #304 bypass.
- **`adapter/mod.rs`** — register the module.

## One deliberate behavior change: a silent-corruption backstop (#361)

The extraction is otherwise behavior-preserving. But the auto-Mythos delta-pass
surfaced a **real, pre-existing latent bug** that the original inline dispatch
carried: a sync call that needs transcoding but does **not** cross memory
(`needs_transcoding && !crosses_memory` — e.g. `--memory shared` fusion of a
UTF-8 caller and a UTF-16 callee) routed to `Direct`, a shim that does not
transcode → string bytes delivered verbatim in the wrong encoding, silently.

meld already fails loud on the identical **async** case
(`guard_async_cross_encoding_strings`, `fact.rs:12139`, #272). This PR adds the
sync twin: `resolve_call_lowering_plan` hard-fails with `AdapterGeneration`
rather than emit a silently mis-transcoding adapter — the #351/#352/#355
silent-corruption-vs-loud-failure choice, applied consistently. Actually
*supporting* same-memory transcoding is tracked in **#361**.

The guard fires on **no existing fixture** (the path was silently latent); full
`meld-core` suite green (0 failures), fmt + clippy clean.

## Async is unaffected

Async boundaries are dispatched **before** `generate_adapter` (the
`is_async_lift` branch in `generate`), so they never flow through this seam; the
`AdapterClass::Async` match arm is `unreachable!` and documents that.

## Why this is the seam ADR-7 wants

Future per-boundary ABI strategies — **symmetric ABI** (cpetig/Aptiv) and
**static-base PIC** (#353) — and same-memory transcoding (#361) become new
inputs/outputs of `resolve_call_lowering_plan` rather than new branches threaded
through the adapter hot path.

Tier-5 (`adapter/fact.rs` + new fusion-critical module) → Mythos delta-pass
below.

Refs #354 (ADR-7), #361, #272

### DIFF
--- a/meld-core/src/adapter/call_lowering.rs
+++ b/meld-core/src/adapter/call_lowering.rs
@@ -30,7 +30,7 @@
 //! makes that coupling explicit and keeps the merger/adapter call site a lookup
 //! rather than a re-derivation.
 //!
-//! ## Extraction is behavior-preserving
+//! ## Extraction is behavior-preserving (plus one silent-corruption backstop)
 //!
 //! Increment 2 lifts the exact three-way dispatch (`site.crosses_memory` /
 //! `options.needs_transcoding()`) and the #304 inline guard out of
@@ -38,8 +38,15 @@
 //! strategies (symmetric ABI, static-base PIC — ADR-7 / #353) become new inputs
 //! to [`resolve_call_lowering_plan`] instead of new branches threaded through
 //! the generator's hot path.
+//!
+//! The one deliberate behavior change: a boundary that needs transcoding but
+//! does **not** cross memory now **hard-fails** instead of silently emitting a
+//! non-transcoding `Direct` shim (#361). The pre-extraction dispatch mapped that
+//! case to `Direct` and mis-transcoded silently; the async path already failed
+//! loud on the identical case (#272), so this closes the sync gap consistently.
 
 use super::AdapterClass;
+use crate::{Error, Result};
 
 /// The facts about one call boundary that determine how it is lowered. All are
 /// cheap booleans read from the resolved [`AdapterSite`](crate::resolver::AdapterSite)
@@ -76,14 +83,40 @@ pub struct CallLoweringPlan {
 
 /// Resolve how a single cross-component call boundary is lowered.
 ///
-/// Behaviour (unchanged from the pre-extraction `generate_adapter`):
+/// Class selection (unchanged from the pre-extraction `generate_adapter`):
 /// - cross-memory + encoding mismatch → `Transcode`
 /// - cross-memory, same encoding → `MemoryCopy`
 /// - same memory → `Direct`
 ///
 /// and `inline_eligible` iff inlining is on, the class is `Direct`, and the call
 /// carries no resource conversions and no post-return.
-pub fn resolve_call_lowering_plan(facts: BoundaryFacts) -> CallLoweringPlan {
+///
+/// # Errors
+///
+/// Hard-fails ([`Error::AdapterGeneration`](crate::Error::AdapterGeneration))
+/// when the boundary needs transcoding but does **not** cross memory
+/// (`needs_transcoding && !crosses_memory`). A same-memory boundary is lowered
+/// as a `Direct` shim that does not transcode, so emitting one for a
+/// mixed-encoding call would deliver string bytes verbatim in the wrong encoding
+/// — a silent miscompile. This is the sync twin of `guard_async_cross_encoding_strings`
+/// (the async path already fails loud here, #272). Supporting same-memory
+/// transcoding is tracked in #361; until then we refuse rather than corrupt.
+pub fn resolve_call_lowering_plan(facts: BoundaryFacts) -> Result<CallLoweringPlan> {
+    // #361: a boundary that needs transcoding but does not cross memory has no
+    // lowering that transcodes — `Direct` (the same-memory class) is a thin
+    // shim. Fail loud instead of silently mis-transcoding (mirrors the async
+    // guard, #272).
+    if facts.needs_transcoding && !facts.crosses_memory {
+        return Err(Error::AdapterGeneration(
+            "same-memory string transcoding is not supported: this call requires \
+             transcoding (caller and callee use different string encodings) but \
+             does not cross a memory boundary, so it would be lowered as a Direct \
+             shim that copies the bytes verbatim — silently mis-transcoding. The \
+             async path already fails loud here (#272); see #361 to support it"
+                .to_string(),
+        ));
+    }
+
     let class = if facts.crosses_memory && facts.needs_transcoding {
         AdapterClass::Transcode
     } else if facts.crosses_memory {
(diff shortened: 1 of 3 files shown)


avrabe:
## Mythos discover pass — NO FINDINGS

This PR is a single logical extraction (the per-boundary call-lowering decision) spanning the three touched Tier-5 files, so it was reviewed as one adversarial unit with the highest prior on the refactor failure mode: **does the extracted `resolve_call_lowering_plan` diverge from the original inline dispatch for any input?** A divergence would silently mis-lower a boundary (wrong `AdapterClass` = wrong ABI/semantic drift, or wrong `inline_eligible` = skipped post-return / resource cleanup).

### `meld-core/src/adapter/call_lowering.rs` (new) — NO FINDINGS
### `meld-core/src/adapter/fact.rs` (wiring) — NO FINDINGS

Term-by-term equivalence vs the removed inline code (old `fact.rs:5297-5332`):

- **Class selection** — identical predicates, identical order, identical else-fallthrough: `crosses_memory && needs_transcoding → Transcode`; `crosses_memory → MemoryCopy`; `else → Direct`. Fields mapped from the exact same source expressions (`site.crosses_memory`, `options.needs_transcoding()`).
- **Body dispatch** — the new `match class` calls the same generator per class, and the `resource_rep_imports` 4th argument to `generate_memory_copy_adapter` **survived**.
- **Inline guard** — the three negations are correct De Morgan duals: `has_resource_rep_calls = !resource_rep_calls.is_empty()` ⟹ `!has_… = is_empty()` (and likewise `resource_new`, and `has_post_return = callee_post_return.is_some()` ⟹ `!has_… = is_none()`). The `&&` chain is pure conjunction over side-effect-free reads, so short-circuit order cannot change the result.
- **`AdapterClass::Async` `unreachable!`** — verified genuinely unreachable: the `generate` loop (`fact.rs` ~12196) tests `is_async_lift` first and `continue`s **before** the `generate_adapter` call, so `Async` never flows into the seam.
- **Truth-table equivalence** — for all 2⁶ `BoundaryFacts` combinations, the new `(class, inline_eligible)` equals the original inline dispatch's output (`class` is a 4-row table on `(crosses_memory, needs_transcoding)`; `inline_eligible` is the same conjunction post-De-Morgan). The 8 in-module unit tests exercise every branch of that table.

### `meld-core/src/adapter/mod.rs` — NO FINDINGS
Change is inert: a one-line `pub(crate) mod call_lowering;` registration. The `AdapterClass` enum (the only fusion-relevant content) is **unchanged** by this PR.

Per the Mythos honesty bar, no finding is reported (nothing to report — the extraction is a mechanical boolean identity, verified by inspection and by the exhaustive branch tests). No Kani/PoC pair is attached because there is no bug to reproduce.

avrabe:
## Triage of the auto-Mythos finding (`crosses_memory=false, needs_transcoding=true`)

The auto gate is doing its job — but two things need separating:

**1. This is NOT a regression from this PR.** The extraction is behavior-preserving. The flagged path (`same memory, different encoding → Direct → transcoding skipped`) is *identical* to `main`'s removed inline code: the old dispatch's `else → Direct` branch also caught `crosses_memory=false` regardless of `needs_transcoding`, and inlined it under the same guard. Verified term-by-term + 2⁶ truth-table (see the NO-FINDINGS comment above). So `resolve_call_lowering_plan` faithfully preserves the decision; it did not introduce the gap.

**2. The finding does surface a REAL, pre-existing latent bug in `main`.** Grounded:
- `crosses_memory` is set purely from memory strategy (`resolver.rs:2866-2879`): under `--memory shared` it is **always false**, independent of encoding.
- `needs_transcoding` = `caller_encoding != callee_encoding`, set independently (`resolver.rs:2972`, `analyze_call_site` `fact.rs:5394-5403`).
- The class dispatch only reaches `Transcode` when `site.crosses_memory` is true, so a **sync, same-memory, mixed-encoding** call routes to `Direct` — which is a thin shim that does **not** transcode. No upstream validation rejects mixed encodings under shared memory (only the LS-P-17 *warning*).
- Net: `meld fuse --memory shared` of a UTF-8 component calling a UTF-16 component silently delivers untranscoded bytes.

**3. meld already hard-fails the analogous ASYNC case.** `guard_async_cross_encoding_strings` (`fact.rs:12139`, #272) fails loud — *"async cross-encoding string transcoding is not yet supported"* — rather than emit a silent mis-copy. The **sync same-memory** path is the unguarded twin of that exact decision, and silent-corruption-vs-loud-failure is the call meld made in #351/#352/#355.

**Resolution:** the honest fix is to add the sync counterpart of the async guard — a hard-fail on `needs_transcoding && !crosses_memory` — turning the silent miscompile into a loud, tested error and clearing this finding legitimately (rather than overriding the gate on a known silent-corruption path). Doing that as part of this PR since the seam is precisely where the decision lives.

avrabe:
## Mythos update — finding resolved by the #361 hard-fail

The auto-gate finding is addressed at the source: `resolve_call_lowering_plan` now **hard-fails** (`AdapterGeneration`) on `needs_transcoding && !crosses_memory` instead of returning `{Direct, inline_eligible}`. So the silent-skip the auto-agent flagged can no longer occur — the seam refuses that boundary rather than mis-lower it.

Re-verified on the new commit (`a1d7d0c`):
- **Guard is exact** — fires **iff** `needs_transcoding && !crosses_memory`. Same-encoding calls (`needs_transcoding == false`) don't fire; legitimate cross-memory transcode (`crosses_memory == true`) still routes to `Transcode`. No over- or under-firing.
- **Threaded correctly** — `generate_adapter` propagates the seam's `Result` via `?`.
- **Fires on no existing fixture** — full `meld-core` suite green (0 failures across all suites); the new `same_memory_transcoding_hard_fails` unit test covers the guard; the other 8 truth-table tests are unchanged.

The class + inline-eligibility decision is otherwise the same behavior-preserving extraction verified earlier (2⁶ truth-table equivalence).

PR #365, opened 2026-07-22
feat(adr7): static-PIC data/element-offset fold — inc 4 (#353)

avrabe:
Implements **ADR-7 path-H, increment 4** — static-PIC data/element-offset folding
(#353) — reproduced *and fixed* in one branch.

## The bug (reproduced)

A data/element segment offset const-expr may `global.get` only an **imported**
global. In a PIC dylib, `(data (global.get $__memory_base) …)` is valid because
`__memory_base` is imported. But when meld **fuses** a shared-everything graph, a
`$main` module *provides* `__memory_base`, so after merge it becomes a **defined**
global — and meld emitted `(data (global.get $base) …)` verbatim. wasm-tools
validates it (lenient); **wasmtime rejects it** at instantiation ("constant
expression required: global.get of locally defined global") → silent invalid
output from valid input.

(The spike's original "2 memories / new topology modeling" premise did *not*
reproduce on current meld — see the earlier #353 comments; the real residual was
this offset fold.)

## The fix

Fold a `global.get` of a **defined constant-i32** global to `i32.const <value>`
in offset emission. Imported globals stay a verbatim `global.get`, preserving
**#338**.

- **`segments.rs`** — `const_i32_init_value` extracts a global init's constant
  i32; `ParsedConstExpr::reindex` folds `GlobalGet(defined-const)` → `I32Const`.
  This sits in `reindex`, so **both** data and element offsets are covered.
- **`merger.rs`** — `MergedModule.defined_global_i32_const` records defined
  constant-i32 globals at global-merge; copied into `IndexMaps` for the reindex.
- **`rewriter.rs`** — `IndexMaps.defined_global_i32_consts` (empty default → no
  fold, so every other caller is unaffected).

## Verification

- **Execution oracle** `pic_base_relative_data_reads_correctly_after_fold` — a
  hand-written PIC-pattern component (base-relative data + a lifted `read`) now
  fuses, **validates**, and executes on wasmtime with the data read back
  correctly (`0xddccbbaa`) at the folded base. A wrong fold (value/index/address)
  fails this.
- **#338 regression** — `const_expr_globalget` (4 tests, imported-base) still
  green: imported globals stay verbatim.
- Structural baseline (`shared_everything_fuses_to_valid_single_memory_core`)
  against a real `wasm-tools component link` fixture still green.
- Full `meld-core` suite green (0 failures); fmt + clippy clean.

Tier-5 (merger/segments/adapter) → Mythos delta-pass below.

Refs #353 (ADR-7 path-H inc 4), #338


avrabe:
## Mythos discover pass (static-PIC offset fold) — NO FINDINGS

A fresh-agent discover run hit the Anthropic session limit (infra, no result), so this is a **self-conducted** adversarial review — backed by the **execution oracle** `pic_base_relative_data_reads_correctly_after_fold`, which (unlike a dead-code unit test) instantiates the fused module on wasmtime and asserts the exact value at the exact folded address, so any wrong fold (value/index/mutability/timing) fails it. The CI auto-Mythos delta-pass runs its own independent agent as a second check.

Hypotheses examined against the diff (`segments.rs` / `merger.rs` / `rewriter.rs`):

1. **Folds an imported global?** No — `defined_global_i32_const` is populated ONLY in the DEFINED-global merge loop, keyed by the absolute `new_idx`; imported globals never enter it, so `reindex` leaves their `global.get` verbatim. Guarded by the #338 `const_expr_globalget` regression (4 tests, imported base) — still green.
2. **Wrong value?** `const_i32_init_value` folds a bare/extended `i32.const` and returns `None` for any `global.get`-based (runtime-dependent) or non-i32 init. The oracle checks `read()==0xddccbbaa` (data at the folded base) — a wrong value fails it.
3. **Mutable global** (subtlest): HARDENED — the fold is now restricted to **immutable** globals (commit 966ee55). Even a mutable const-init would read its init value at segment-init (active segments initialise before the start function), but immutable removes all doubt and matches `__memory_base`'s shape.
4. **Element segments:** the fold lives in `ParsedConstExpr::reindex`, shared by data AND element offsets, so both are covered consistently; element/data segment tests in the suite stay green.
5. **Index-space / remap:** the fold does `remap_global(idx)` then looks up the post-remap NEW index — the same space `defined_global_i32_const` is keyed in. The map is copied into `IndexMaps` immediately after it's built for the module, before the reindex uses it.
6. **Cross-module timing:** the oracle IS cross-module (`$main` defines the base; `$lib`'s data offset references it) and passes — proving the referenced global is recorded before the referencing module's segments are reindexed.

Empty default (`IndexMaps.defined_global_i32_consts`) → no fold for any other caller, so the change is inert outside this path. Full `meld-core` suite green (0 failures); fmt + clippy clean. Per the honesty bar, no finding is reported.

Issue #361, opened 2026-07-17
fuse: sync same-memory cross-encoding call skips transcoding — hard-failed in #360, support tracked here


avrabe:
## Gap

The sync call-lowering class dispatch selects `Transcode` only when the boundary **crosses memory** (`site.crosses_memory && options.needs_transcoding()`). `crosses_memory` is set purely from the memory strategy (`resolver.rs:2866-2879`) — under `--memory shared` it is **always false**, independent of string encoding.

So a **sync, same-memory, mixed-encoding** call (e.g. a UTF-8 component calling a UTF-16 component fused with `--memory shared`) routes to `AdapterClass::Direct` — a thin shim that does **not** transcode — and the string bytes are delivered **verbatim in the wrong encoding**. No upstream validation rejects mixed encodings under shared memory (only the LS-P-17 *warning*).

This is the sync twin of the **async** case, which meld already fails loud on: `guard_async_cross_encoding_strings` (`fact.rs:12139`, #272) — *"async cross-encoding string transcoding is not yet supported"*.

## Reachability

Legal input: two components with different canonical `string-encoding` options fused with `--memory shared`. Silent semantic drift (H-class: fused output diverges from composed input). Pre-existing; surfaced by the auto-Mythos delta-pass on the call-lowering seam (#360).

## Immediate mitigation (in #360)

The call-lowering seam (`adapter::call_lowering::resolve_call_lowering_plan`) now **hard-fails** on `needs_transcoding && !crosses_memory`, mirroring the async guard — turning the silent miscompile into a loud, tested `AdapterGeneration` error. This closes the *silent-corruption* hazard.

## This issue tracks the capability

Actually **supporting** same-memory string transcoding (a `Transcode`-class adapter that transcodes in place / into a scratch buffer within the shared memory, without the cross-memory `cabi_realloc` + `memory.copy` assumption) so the hard-fail can be lifted. Fits the ADR-7 path-H per-boundary call-lowering seam: a same-memory-transcode boundary is a distinct lowering the seam should eventually resolve to a real adapter rather than reject.

Refs #360, #272, ADR-7 (path-H inc 2).

PR #367, opened 2026-07-23
feat(361): support same-memory string transcoding (lift the #360 hard-fail)

avrabe:
Closes **#361** — supports **same-memory string transcoding**, lifting the #360
hard-fail.

## Background

#360 hard-failed a sync boundary that needs transcoding but does **not** cross
memory (e.g. `--memory shared` fusion of a UTF-8 caller + UTF-16 callee) — the
loud-not-silent placeholder for the gap the call-lowering seam surfaced. This
lands the actual capability.

## The fix is small — the adapter already supported it

`generate_transcoding_adapter` is **memory-index-parameterised**: source reads
use `caller_memory`, destination writes use `callee_memory`, and it transcodes
byte-by-byte (there is **no** cross-memory `memory.copy`). So when the two are
the *same* memory (both `0` under `--memory shared`), it transcodes correctly
within that one memory. Only the **class dispatch** was gating it — it reached
`Transcode` only when memory was crossed.

- **`call_lowering.rs`** — `resolve_call_lowering_plan` now routes any
  `needs_transcoding` boundary to `Transcode` regardless of `crosses_memory`, and
  the #360 hard-fail is dropped. (`same_memory_transcoding_hard_fails` unit test →
  `..._routes_to_transcode`.)

## Verification

- **`test_361_same_memory_utf8_to_utf16_transcoding`** — the SR-17 UTF-8→UTF-16
  'Hello' fusion under `SharedMemory`, executed on wasmtime, returns **500**
  (correct transcode within the shared memory), not a verbatim mis-copy.
  Confirmed it **hard-failed before** this change (baseline).
- 8 `call_lowering` truth-table unit tests updated + green.
- Full `meld-core` suite green (0 failures); fmt + clippy clean.
- SR-17 (correct string transcoding) gains the same-memory execution oracle.

## Coverage note

The execution oracle covers **UTF-8→UTF-16**. The other same-memory directions
(UTF-16→UTF-8, latin1+utf16) route through the same memory-index-parameterised
adapter, but are not separately execution-tested here — a follow-up could add
them, mirroring the cross-memory SR-17 direction matrix.

Tier-5 (`adapter/`) → Mythos delta-pass below.


avrabe:
## Mythos discover pass (#361 same-memory transcoding) — NO FINDINGS

Self-conducted (backed by the execution oracle) + CI auto-Mythos as the independent second check. The change is small: `resolve_call_lowering_plan` routes `needs_transcoding` → `Transcode` regardless of `crosses_memory`, dropping the #360 hard-fail.

Hypotheses examined:
1. **Does routing same-memory transcoding to `Transcode` produce correct output?** YES — `test_361_same_memory_utf8_to_utf16_transcoding` fuses under `SharedMemory`, executes on wasmtime, and returns 500 (correct UTF-8→UTF-16 transcode). The adapter is memory-index-parameterised (src `caller_memory` / dst `callee_memory`, both 0), byte-by-byte transcode, no cross-memory `memory.copy`, so same-memory is a valid special case.
2. **Regression on other classes?** No — `crosses_memory && !needs_transcoding` still → `MemoryCopy`; `!crosses_memory && !needs_transcoding` still → `Direct`. Only the previously-hard-failed `!crosses_memory && needs_transcoding` cell changed (fail → Transcode). Full suite green.
3. **Inline eligibility?** `Transcode` is never `inline_eligible` (only `Direct` is) — unchanged.

**Coverage note (honest):** the execution oracle covers UTF-8→UTF-16 only; the other same-memory directions (UTF-16→UTF-8, latin1+utf16) route through the same parameterised adapter but are not separately execution-tested. Not a correctness risk introduced here (same adapter as the verified cross-memory SR-17 matrix), but a follow-up could add the same-memory direction matrix.

PR #368, opened 2026-07-23
fix(353): fold extended-const (base+N) data/element offsets — completes inc-4 fold

avrabe:
Completes the inc-4 static-PIC offset fold (#365) — which had a real gap found
by the inc-4 Mythos discover agent (it died on a session-limit but left the
reproducing oracle behind).

## The gap

#365 folded the **bare** `(data (global.get $base) …)` shape but left the
**extended-const** `(data (i32.add (global.get $base) (i32.const N)) …)` shape as
a verbatim `global.get`. But `wasm-tools component link` emits the extended-const
form for **every N > 0** data segment (only N == 0 is bare) — so the *common* PIC
case was still emitted as `global.get <defined base>; i32.const N; i32.add`,
which wasm-tools accepts (lenient) but **wasmtime rejects** ("constant expression
required: global.get of locally defined global"). Silent invalid output.

## The fix

`ParsedConstExpr::reindex`'s `ExtendedGlobalGet` arm now folds the whole
`base ± N` expression to a single `i32.const` when the leading global is a
**defined** constant-i32 (reusing the `defined_global_i32_consts` map from #365),
via a small stack evaluator `eval_ext_const_i32_with_base`. It **declines** (and
leaves the expression verbatim) when:
- the base is **imported** (#338 — runtime-dependent, stays a `global.get`),
- a **non-leading** `global.get` appears (value depends on a 2nd global),
- any **i64** op appears (not a 32-bit memory offset), or
- the sequence is **unbalanced**.

## Verification

- **Execution oracle** `pic_extended_const_data_offset_valid_for_wasmtime_353` —
  a PIC-pattern component with an extended-const `base + N` data offset fuses,
  **validates for wasmtime**, and reads the data back correctly. **Was FAILING
  on main** before this fix.
- Unit test covering `base±N` / `base*N` / bare, plus all three decline cases.
- The #365 bare-fold oracles still green; full `meld-core` suite green (0
  failures); fmt + clippy clean.

Tier-5 (`segments.rs`) → Mythos delta-pass below.

Refs #353 (ADR-7 path-H inc 4), #365, #338.

### DIFF
--- a/meld-core/src/segments.rs
+++ b/meld-core/src/segments.rs
@@ -79,6 +79,47 @@ pub(crate) fn ext_const_to_expr(ops: &[ExtConstOp]) -> ConstExpr {
     ce
 }
 
+/// Evaluate a `global.get`-first extended-const i32 sequence (`base ± N …`) as a
+/// concrete constant, substituting `base` for the leading `global.get`. Returns
+/// `None` if the sequence isn't a pure i32 expression over a *single* (leading)
+/// `global.get` — a non-leading `global.get`, an i64 op, or an unbalanced stack
+/// all decline the fold. Used by [`ParsedConstExpr::reindex`] to fold a data /
+/// element offset `global.get $__memory_base + N` once the base is a defined
+/// constant (#353).
+fn eval_ext_const_i32_with_base(ops: &[ExtConstOp], base: i32) -> Option<i32> {
+    let mut stack: Vec<i32> = Vec::new();
+    for (i, op) in ops.iter().enumerate() {
+        match op {
+            ExtConstOp::GlobalGet(_) if i == 0 => stack.push(base),
+            // Only the leading global.get is the folded base; any other one means
+            // the value depends on a second global we can't fold here.
+            ExtConstOp::GlobalGet(_) => return None,
+            ExtConstOp::I32Const(v) => stack.push(*v),
+            ExtConstOp::I32Add => {
+                let b = stack.pop()?;
+                let a = stack.pop()?;
+                stack.push(a.wrapping_add(b));
+            }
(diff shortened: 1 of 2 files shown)


avrabe:
## Mythos discover pass (extended-const offset fold) — NO FINDINGS

Self-conducted (backed by the execution oracle + a unit test on the evaluator's decline cases) + CI auto-Mythos as the independent second check. Hypotheses examined against the diff (`segments.rs`):

1. **Folds an imported base?** No — the fold only fires when the leading global's post-remap index is in `defined_global_i32_consts`, which is populated ONLY for DEFINED immutable constant-i32 globals (#365). An imported `__memory_base` is absent → the extended-const stays verbatim (#338 runtime-dependent shape preserved).
2. **Wrong folded value?** `eval_ext_const_i32_with_base` is a straight i32 stack machine using `wrapping_{add,sub,mul}` (matches wasm i32 semantics). The execution oracle checks the data reads back correctly at the folded address; the unit test pins `base±N`, `base*N`, and bare-base.
3. **Over-folds a genuinely runtime-dependent expr?** Declines (returns None → verbatim) on: a NON-leading `global.get` (value depends on a 2nd global), any i64 op, or an unbalanced stack. Unit-tested.
4. **Element segments?** The fold is in `ParsedConstExpr::reindex`, shared by data AND element offsets — both covered by the same change.
5. **Interaction with the bare-fold (#365)?** Independent match arms (`GlobalGet` vs `ExtendedGlobalGet`); the #365 bare-fold oracles stay green.

Empty-default `defined_global_i32_consts` → inert for every other caller. Full suite green; fmt + clippy clean. No finding reported.

Issue #376, opened 2026-08-13
Split oversized meld-core source files for maintainability (merger.rs done; resolver.rs / parser.rs remain)


avrabe:
## Problem

The largest `meld-core/src` files have grown past a workable size:

| file | lines |
|---|---|
| `merger.rs` | 6655 |
| `resolver.rs` | 6044 |
| `parser.rs` | 6019 |
| `component_wrap.rs` | 4163 |
| `lib.rs` | 3432 |

Two concrete harms, not just aesthetics:

1. **It breaks the Tier-5 auto-Mythos gate on our most-changed file.** The CI auto-Mythos discover step (`anthropics/claude-code-action` per `.github/workflows/mythos-auto.yml`) **consistently fails on `merger.rs`** — "discover step failed before emitting structured output" — because the 6.6k-line file exceeds the action's context/token/time budget. On PR #375 it failed twice on `merger.rs` while `fact.rs` passed on re-run. So the adversarial gate cannot actually run on the file we touch most (index-space merging, shared-memory placement, adapter wiring all live here). That is precisely where we most want it.
2. **It slows every change.** Agents and humans re-read 6k-line files to make small edits; the blast radius of a change is hard to reason about when one file spans parsing helpers, index maps, shared-memory planning, and adapter wiring.

## Proposal

Behavior-preserving module splits by cohesive concern, **no public-API change** (re-export from the parent module so call sites are unchanged):

- `merger.rs` → e.g. `merger/index_maps.rs`, `merger/shared_memory_plan.rs` (compute_shared_memory_plan + module_used_data_extent), `merger/adapter_wiring.rs`, `merger/imports.rs`.
- `resolver.rs`, `parser.rs`, `component_wrap.rs` → similar cohesive submodules.

## Acceptance

- Each resulting file is small enough that the auto-Mythos discover step runs to completion (a green per-file Mythos pass, not "discover step failed").
- `cargo test --workspace` unchanged (behavior-preserving); public API byte-identical.
- rivet trace unaffected (implementation-file paths in the SRs updated to the new module paths).

This is a tooling-efficiency + gate-unblock task; it does not change fusion behavior. Filing per the standing "our work should be efficient" bar.

avrabe:
[...]

Issue #370, opened 2026-08-11
`--memory shared` without `--address-rebase` merges memories but does NOT rebase — data segments silently overlap, exit 0


avrabe:
meld 0.41.3. `--memory shared --address-rebase` correctly **refuses** when inputs lack relocation metadata. Plain `--memory shared` accepts the same inputs, prints `Fusion complete!`, exits 0 — and produces a module whose data segments overlap.

## Reproduction

Three real components (gale's verified isolation core: health monitor, MPU region programmer, partition switch), each a `wit-bindgen` + `cargo` wasm component, `reloc-sections=0`:

    meld fuse hm.component.wasm mpu.component.wasm switch.component.wasm \
      --memory shared -o iso.fused.wasm
    # Size reduction: 34.4% / Fusion complete!  (exit 0, no warning)

The result has **one** memory and five data segments:

```
data 0: mem 0  [1048576 .. 1048888)  len=312
data 1: mem 0  [1048888 .. 1048948)  len=60
data 2: mem 0  [1048576 .. 1048888)  len=312   <- identical range to data 0
data 3: mem 0  [1048888 .. 1049208)  len=320   <- overlaps data 1
data 4: mem 0  [1048576 .. 1048584)  len=8     <- inside data 0 and data 2

OVERLAPPING PAIRS: 4
```

Every input places `.rodata` at `1048576` (the standard wasm-ld data base) and `.data` immediately above. Fused into one memory with no rebasing, they land on top of each other. Data segments are applied in order at instantiation, so later segments overwrite earlier ones.

`data 3` is a 320-byte `RegionTable` — the isolation core's **mutable** memory-protection state — initialised over the top of another component's `.data`.

Two-component case is the same and smaller if easier to work with:

```
data 0: [1048576 .. 1048584)    8 B   (component B .rodata)
data 1: [1048576 .. 1048888)  312 B   (component A .rodata) -- clobbers data 0
data 2: [1048888 .. 1048948)   60 B
```

## Why this is the dangerous variant

The `--address-rebase` path is well-behaved. It warns:

    memory strategy: explicit shared-memory + address rebasing is UNSOUND for components
    that access memory via computed pointers (#326) -- the dynamic address operand of
    ordinary loads/stores is not rebased, so per-component memory can silently collide.

[...]

PR #377, opened 2026-08-13
feat(v0.47.0): #370 pack-rebase __heap_base extent (silent→loud) + meld docs subcommand

avrabe:
Two independent pieces bundled: the escalated #370 unblock (meld's half) and the `meld docs` subcommand.

## SR-65 — `--pack-rebase` reserves the true static extent or refuses (#370)
`--pack-rebase` packed by the active-data-segment extent only, so a component whose accessed memory (a zero-init `.bss` region / relay's bounded static arena) extends **above its last data segment** was silently under-reserved — and SR-56's overlap check is active-data-only, so the cross-component collision wasn't caught (silent corruption). Confirmed on the real falcon `1.133.0` bytes: 17 declared pages, data ending ~1048672, a **65440-byte arena tail invisible to the data-segment scan**.

Compaction below the declared page count now **requires an authoritative `__heap_base` marker** (defined immutable `i32` global, read from the export table or name section); absent it, meld **falls back to page-granular and warns** — never silently under-reserving. `declared` (initial pages) caps the extent only for a **defined** memory; for an **imported** memory it's a floor (a dylink dylib addresses far above page 1). **A local Mythos delta-pass caught + closed an imported-memory under-reserve regression** in the first cut — pinned by its own oracle.

Two-sided: no published falcon artifact exports `__heap_base` (verified — absent from exports, linking symtab, name section), so compaction of published artifacts is gated on suppliers retaining `__heap_base` alongside `--emit-relocs`. meld's half is sound.

**Oracles** (`pack_rebase_370.rs`, all RED→GREEN by revert): `.bss`-arena with/without marker + imported-memory-above-floor with/without marker.

## SR-64 — `meld docs` embedded queryable documentation
Topics compiled into the binary (offline, `include_str!`). `meld docs` / `docs <topic>` / `--grep` / `--format json` / `meld docs check --coverage --strict` (CI-gate form). 11 topics (fuse/inspect/version/docs + concept topics). Coverage test walks the CLI via `clap::CommandFactory`. Models `rivet docs` / varve's REQ-DOCS-001.

## Verification
Full suite **830 passed / 45 binaries / 0 failed**; `clippy --all-targets` clean; `rivet validate` PASS. SR-64 + SR-65 verified, SWV-76/77.

## Mythos note
The #370 change is a **production** change to `merger.rs`, so it needs the Tier-5 gate — but the CI auto-Mythos discover **crashes on the 6.6k-line merger.rs** (tracked as #376). A **local Mythos delta-pass was run** (it found + I fixed the imported-memory regression); applying `mythos-pass-done` on that basis. #376 (split the oversized files) is the durable fix for the gate itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,32 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [0.47.0] - 2026-08-13
+
+### Fixed
+- **`--pack-rebase` no longer silently under-reserves `.bss` / static arenas
+  (SR-65, #370)** — it packed by the active-data-segment extent only, so a
+  component whose accessed memory (a zero-init `.bss` region or relay's bounded
+  static arena) extends above its last data segment was under-reserved, and
+  SR-56's overlap check is active-data-only so the cross-component collision
+  wasn't caught — silent corruption (confirmed on the real falcon components: a
+  65440-byte arena tail invisible to the data scan). Compaction below the
+  declared page count now **requires an authoritative `__heap_base` marker**
+  (defined immutable `i32` global, read from the export table or name section);
+  absent it, meld **falls back to page-granular and warns** (`-Wl,--export=__heap_base`),
+  never silently under-reserving. `declared` (initial pages) caps the extent only
+  for a **defined** memory; for an **imported** memory it is a floor. A local
+  Mythos delta-pass caught + closed an imported-memory under-reserve regression
+  in the first cut. Compaction of published artifacts is gated on suppliers
+  retaining `__heap_base` alongside `--emit-relocs`.
+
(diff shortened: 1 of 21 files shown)


avrabe:
Mythos-gate note: the #370 change to `merger.rs::module_used_data_extent` is a **production** change, so it needs the Tier-5 gate. The CI auto-Mythos discover **crashes on the 6.6k-line merger.rs** (infra, tracked as #376). A **local Mythos delta-pass was run** on this exact change — it found a runtime-proven imported-memory under-reserve regression in the first cut, which was then fixed and pinned by `pack_rebase_imported_memory_*` oracles. Applying `mythos-pass-done` on that basis; #376 will restore the automated gate on these files.

PR #375, opened 2026-08-13
feat(v0.46.0): traceability close-out (foundational reqs verified) + rich witness MC/DC gate

avrabe:
Traceability close-out + verification depth (Theme H). **No production behavior change** — this release closes the V on the foundational requirements and deepens the verification evidence. Diff is tests + fixtures + rivet YAML + one CI gate.

## Foundational V closed
The 15 foundational sw-reqs **SR-1..18 + SR-36** were `implemented` but never formally `verified`. This closes their V, grounded (not rubber-stamped):
- **12** flipped from their existing real, passing tests (confirmed present + green), SWV-54..65.
- **SR-8 / SR-14 / SR-18** got new **mutation-verified direct tests** (they had Rocq proofs / inspection but no direct executable test): `sr8_function_base_offset_is_cumulative`, `sr14_adapter_memory_indices_caller_to_callee` (decodes memory.copy src/dst by name — the 0xFC-0x0A encoding gotcha), `sr18_adapter_emits_realloc_before_copy_before_call` (upgrades inspection → mechanically checked). Each mutation-verified (a broken invariant turns it red).

## All 63 SRs now verifies-linked
SWV-66..75 added the missing backlinks for SR-46..55 → `rivet coverage: sw-req has-verification 63/63 (100%)`, machine-enforced. **SR-44** (enforced req→verification trace) advanced to `implemented` — its verifies-trace part is done; the STPA constraint-coverage remainder is documented as maintainer-gated, not falsely claimed.

## Honest requirement-text corrections (the tests surfaced these, #364-class)
- **SR-8**: dropped the "(imports + defined functions)" per-component-block wording — impossible in a flat wasm index space (all imports precede all defined) — for the equivalent flat-index invariant.
- **SR-14**: scoped the `i32.load`/`i32.store` caller/callee split to the **return** path (args-side loads/stores operate within the already-copied callee buffer). The implementation was always correct; the text was wrong.

## Rich witness MC/DC gate (SR-62 depth)
A rustc-built leap-year fixture (`tests/gate/leap_mcdc`) whose DWARF + short-circuit branch chain **survive fusion**, so the witness gate now recovers a genuine 2-condition MC/DC decision on the **fused** output:
```
decisions: 1/1 full MC/DC; conditions: 2 proved, 0 gap, 0 dead
decision #0 lib.rs:55: FullMcdc  (c0 masking, c1 unique-cause)
```
[...]

PR #378, opened 2026-08-14
refactor: split merger.rs into cohesive submodules (maintainability; behavior-preserving)

### LINKED ISSUE #376: Split oversized meld-core source files for maintainability (merger.rs done; resolver.rs / parser.rs remain)

avrabe:
Closes #376. `merger.rs` (7020 lines) was too large for the CI auto-Mythos discover action to process — it crashed ("discover step failed before emitting structured output") on every PR touching it, so the Tier-5 adversarial gate could not run on our **most-changed** file (this was visible on #375 and #377). Splitting it restores that gate.

## Strictly behavior-preserving
`merger.rs` → `merger/mod.rs` + submodules, **pure code movement** + visibility/`use` plumbing:

| file | lines |
|---|---|
| mod.rs | 1206 |
| tests_a.rs / tests_b.rs | 1419 / 1174 |
| merge_core.rs | 918 |
| imports.rs | 809 |
| index_maps.rs | 575 |
| memory.rs | 563 |
| handle_tables.rs | 270 |
| naming.rs | 160 |

The `impl Merger` block is split across submodules (inherent methods resolve on the type regardless of file; 5 methods widened to `pub(crate)` for cross-module calls). Every `pub(crate)`/`pub` item is re-exported from `mod.rs`, so **no `crate::merger::<name>` call site outside the module changed**; `lib.rs` is untouched.

## The actual gate-fix mechanism
`.github/workflows/mythos-auto.yml` + `mythos-gate.yml`: the discover target `meld-core/src/merger.rs` → `meld-core/src/merger/` (directory prefix, mirroring the existing `adapter/` entry). Without this, a stale `merger.rs` path would match nothing and drop merger code out of Tier-5 scanning entirely.

## Verification
- **`cargo test --workspace`: 830 passed / 45 binaries / 0 failed — identical to the main baseline** (per-binary `running N` multiset byte-identical, proving no test silently dropped).
- A move-verification oracle (whitespace-stripped sorted line union, split vs. original) shows the ONLY removed lines are 21 documented `fn`→`pub(crate) fn` widenings, 3 struct + 8 field visibility bumps, and the `mod tests` → `tests_a`/`tests_b` split. **Zero logic lines changed.**
- `clippy --all-targets` clean; `fmt --check` clean; `rivet validate` PASS (implementation-path fields in the SRs repointed to the submodule each function moved to).

## The real acceptance test is THIS PR's CI
The per-file Mythos discover jobs should now **run to completion** on the new sub-2000-line submodules instead of crashing — restoring the automated adversarial gate on merger.rs that this repo's recent history (the #370 imported-memory regression, caught by a *local* Mythos pass because the CI one couldn't run) shows we need.


avrabe:
Mythos-gate note: this is a **pure-move, behavior-preserving refactor** (the move-verification oracle shows zero logic lines changed — only visibility widenings + the tests split; suite 830/45/0 identical to baseline). No behavior to adversarially analyze. Applying `mythos-pass-done` on that basis. The POINT of this PR is to verify the auto-Mythos discover now **runs to completion** on the sub-2000-line `merger/*.rs` submodules instead of crashing on the old 7020-line file — watching those jobs is the #376 acceptance test.

avrabe:
**Correction — my #376 premise was wrong, and the CI here proves it.** All 9 per-submodule Mythos discover jobs failed with the same "discover step failed before emitting structured output" — **including the 160-line `naming.rs`**, and each fast-failed in ~27s. So it is NOT file size. And the auto-Mythos matrix already runs `max-parallel: 2`, so it is not a concurrency blowout from the split either. The `claude-code-action` (maintainer's Max-plan OAuth token) is failing fast at invocation, intermittently — it passed on #377, failed on all 9 here. That is an action/token reliability issue at the account level, which **splitting merger.rs cannot fix** (and, honestly, this PR now makes 9 jobs fail instead of 1).

**What this PR actually is, reframed honestly:** a maintainability refactor — 7020 lines → cohesive sub-2000-line submodules, strictly behavior-preserving (830/45/0 identical to baseline, move-only diff, public API byte-identical). That value stands on its own. But it does NOT "restore the Mythos gate" as the PR title/body claim, and I will not merge it under that banner.

**The real fix for the gate** is the `claude-code-action`/OAuth-token reliability (retry/backoff on the fast-fail, a healthier token, or a fallback) — a separate, maintainer-side issue I'll file. The required `Mythos delta-pass gate` is green here (behavior-preserving refactor + label), so the failing jobs are the advisory auto-run.

Holding this PR for a call on whether the maintainability gain is worth merging on its own, given it doesn't achieve #376's stated goal and adds 9 advisory jobs per merger-touching PR.

PR #392, opened 2026-09-02
fix(v0.53.0): bridge a same-memory boundary whose calling conventions differ (SR-71, #390)

### LINKED ISSUE #390: fuse --memory shared: same-memory boundary whose calling conventions differ emits invalid wasm at exit 0 (rebasing is NOT the cause — see maintainer diagnosis)
## Summary

`meld fuse --memory shared --address-rebase` **exits 0 and emits a core module that fails `wasm-tools validate`** when one input imports a record-carrying interface that another input exports. The failure is a malformed function body — a stack/type discipline break, not an address problem.

This is distinct from the known rebase issues. **#351 explicitly states "the fused module validates cleanly"**; #326 and #339 are likewise about *wrong addresses in a valid module*. Here the module is not valid at all, and meld reports success.

Verified on **meld 0.41.3**.

## Minimal reproduction — two components

```
meld fuse flight-app.wasm rate.wasm --memory shared --address-rebase -o min.wasm
  -> exit 0   (only the standard #326 UNSOUND warning)

wasm-tools validate min.wasm
  error: func 14 failed to validate
  Caused by:
    0: type mismatch: values remaining on stack at end of block (at offset 0x840)
```

`flight-app` is a small `no_std` component that imports `pulseengine:falcon-cascade/rate@0.7.0` and `gust:os/time@0.1.0`; `rate` is relay's published falcon cascade component. Both validate individually and both fuse fine on the default path.

## What separates the passing cases from the failing ones

| fusion | `--memory shared --address-rebase` | default (`auto` → `multi`) |
|---|---|---|
| `flight-app` alone | valid | valid |
| `rate` alone | valid | valid |
| `flight-app` + `gale-nano` | **valid** | valid |
| `rate` + `mixer` | **valid** | valid |
| `flight-app` + `rate` | **INVALID** (func 14) | valid |
| `flight-app` + `mixer` | **INVALID** (func 12) | valid |
| `flight-app` + `rate` + `mixer` | **INVALID** (func 18) | valid |
| full 19-component composition | **INVALID** (func 72, *expected f32, found i32*) | valid |

Note the two informative rows:

- **`rate` + `mixer` is valid.** Both are falcon components with record-carrying interfaces — but *neither imports the other*, so no call is wired across the fused boundary.
- **`flight-app` + `gale-nano` is valid.** Here a call *is* wired across the boundary, but `gust:os/time` is scalar-only (`u64`, `bool`).

So the trigger appears to be **a wired cross-component call whose signature passes or returns a record via linear memory** — precisely the calls whose operands the rebase pass has to touch. I'd offer that as a hypothesis to aim the search, not as a diagnosis; I haven't read the rebase implementation.

## Things I ruled out

- **Not relocation-driven.** Rebuilding `flight-app` *without* `--emit-relocs` reproduces the identical `func 14` failure. All three inputs carry `reloc.*` sections either way.
- **Not `wac` composition.** Fusing the four raw components directly, with no `wac` step at all, fails the same way.
- **Not a stale toolchain generally.** The pre-existing falcon fused cascade still validates cleanly.
- **Not my harness.** The default `auto`/`multi` path produces a valid module from the same inputs, and that module executes correctly (see below).

## Why this matters beyond the validation error

The **error is silent**: exit 0, no diagnostic, and the only warning printed is the standard #326 unsoundness note — which points at *computed addresses*, not at *malformed output*. A CI step that fuses and then lowers will carry an invalid module forward to whatever runs next.

There is also a real consequence for the RT1176 work. The composed stack has two candidate paths and neither currently reaches the target:

- `--memory multi` → **validates and runs correctly**, but as meld itself notes, multi-memory *"has no single-address-space (MCU) lowering"* (#172).
- `--memory shared --address-rebase` → the only single-address-space route, and it **produces invalid wasm** for this graph.

So today a composed application+runtime+cascade image has **no MCU lowering path**. I'm not asking you to treat that as this issue's scope — #172 and #353 already cover the strategy question — but it's the reason the validation failure is blocking rather than cosmetic for us.

## Positive control, so the inputs aren't in doubt

On the default path the same composed image validates, and executing it reproduces an independently-computed reference exactly:

```
wasm-tools validate full.wasm          -> VALID
wasmtime run --invoke run() full.wasm  -> 2147484996
    low 31 bits = 1348   falcon cascade fold
    bit 31      = 1      gust:os/time genuinely exercised
```

1348 is cross-checked by a second driver that recomputes the same value from the fused core module through raw canonical-ABI pointers — no component model, no `wac`. So the inputs compose and compute correctly; it is specifically the rebase path that breaks them.

Happy to attach the two input `.wasm` files, or to re-run any narrowing that would help.

## One incidental note

Under the canonical ABI these falcon exports are **not** uniform: `rate#tick` takes a pointer (18 scalars exceeds the flattening limit) while `mixer#mix` takes 4 flattened `f32`s. Mentioning it only because the full-composition error is literally *"expected f32, found i32"*, and that asymmetry is the one place in this graph where an `f32`/`i32` confusion would be easy to introduce.

avrabe:
Fixes #390, reported by jess.

## The attribution changes

The issue title blames `--address-rebase` / `--pack-rebase`. **Rebasing is innocent.** Every test in this PR reproduces the defect with `--memory shared` and **no rebase flag at all**.

Rebase flags only ever appeared next to the failure because SR-56 (overlapping data segments) and the path-F relocation gate force them to be present on real inputs. The defect is on the **call** axis, not the memory axis.

## What is actually wrong

`--memory shared` makes every boundary `Direct`, and the `Direct` lowering assumed "same memory" implied "same signature". It does not. The Canonical ABI switches a side to an indirect convention whenever a value exceeds a flattening limit, so a record travelling through linear memory leaves the two sides typed differently:

```
caller (lowered): (params..., retptr: i32) -> ()
callee (lifted):  (params...)              -> i32   ; return-area pointer
```

`generate_direct_adapter` declared the adapter with the **callee's** type and forwarded params verbatim; on top of that `inline_eligible` (#304) wired the caller **straight to the callee**. Both emit a `call` whose operands do not match the callee's type — hence *"values remaining on stack at end of block"*, and *"expected f32, found i32"* where a flattened `f32` record met an `i32` pointer.

The cross-memory path already handled this — `generate_memory_copy_adapter` detects it (`uses_retptr`) and routes to `generate_retptr_adapter`. Only the same-memory path had no equivalent, which is exactly why `--memory multi` always validated.

## The fix

- `generate_direct_adapter` now consults the caller's import type and, when it differs, emits `generate_same_memory_bridge`: typed as the **caller**, it calls the callee and `memory.copy`s the return area to the caller's retptr. One memory means no `cabi_realloc` and no pointer fixup, so this is a much smaller emitter than the cross-memory `generate_retptr_adapter` — which assumes distinct memories and cannot be reused here.
- `BoundaryFacts` gains `signatures_match`, so a bridge is never inlined. The "#304 INVARIANT" comment already said the inline guard must stay a superset of the Direct generator's complex-branch trigger. It was not. Now it is.
- Anything unbridgeable **hard-fails**, naming the boundary and the reason: result indirections, a post-return that would free them after a flat copy, resource conversions, or an unknown return-area size. Emitting invalid wasm at exit 0 is precisely what that replaces.

## Verification

Hermetic fixture (`tests/wit_bindgen/fixtures/compose_record/`, wasm-tools + wac, no network) carrying a 4×f32 record across a same-memory boundary. The unfused composition returns 26.

| | result |
|---|---|
| composed component on wasmtime | `26` |
| meld 0.52.0 `--memory shared` | exit 0, **INVALID** — `expected f32, found i32` |
| this PR `--memory shared` | **valid**, executes → `26` |

Validity alone is not the gate — a bridge copying the wrong bytes would still validate — so both the computed value and the emitted operator sequence are pinned.

**Negative control:** with the bridge dispatch disabled, 3 of the 4 integration tests fail with the reported error while the `--memory multi` control stays green.

**The reporter's real graph was checked against the bridge predicate directly**, not assumed. All five cascade boundaries match the return-area shape and none carries a post-return:

```
ekf#estimate   caller (f32x6, i32) -> ()   callee (f32x6) -> i32
*/tick         caller (i32, i32)   -> ()   callee (i32)   -> i32
mixer#mix      caller (f32x4, i32) -> ()   callee (f32x4) -> i32
```

## Traceability

SR-71 + SWV-83; `rivet validate` PASS.

Local gate: fmt clean, clippy `-D warnings` clean, 871 tests (`--workspace`), 870 (`--all-features`).


avrabe:
## Mythos delta-pass — `meld-core/src/adapter/fact.rs`, `meld-core/src/adapter/call_lowering.rs`

Adversarial pass over the delta, per `scripts/mythos/discover.md`. **One finding, on my own fix.** Fixed in 5194080.

---

### FINDING 1 — `signatures_match` fails open on untyped sites

- **FILE:** `meld-core/src/adapter/fact.rs`
- **FUNCTION / LINES:** `signatures_match` — the `None => true` arm
- **HYPOTHESIS:** A boundary that needs the convention bridge but whose `AdapterSite` carries `import_func_type_idx: None` silently takes the old identity-forward path and emits an invalid module at exit 0 — reintroducing #390 through the back door.

**Investigation.** Three sites construct `AdapterSite` with `import_func_type_idx: None`: `resolver.rs:2263` (`__main_module__` core linking), `resolver.rs:3796` (intra-component module resolution, `to_component == from_component`), and `resolver.rs:5566` (a test). The canon-boundary constructors (`:3259`, `:3585`) both populate it. So the untyped sites are core-level resolutions where both sides come from one linker and no Canonical ABI lift/lower pair exists — there is no convention for them to disagree about.

**Disposition: not a reportable bug — no PoC exists.** I could not construct an input where an untyped site carries mismatched signatures, and per the discover.md bar ("hallucinations are more expensive than silence") an unreproducible hypothesis does not get reported as a defect.

**But the reasoning is load-bearing and was only reasoning**, and being wrong costs exactly the #390 failure mode. So the wiring step now asserts the property directly rather than relying on the argument: whatever a caller's import is rewritten to call must have that import's own type, else meld fails naming both types.

**Potency, since an unreachable assert is worth nothing unproven.** Disabling the bridge dispatch in `generate_direct_adapter` makes the #390 fixture hard-fail *at the backstop*:

```
#390: wiring 'tick' -> 'golden:ctrl/lib@0.1.0#tick' (inlined-direct) would produce
a type-invalid call: the caller's import is [F32, F32, I32] -> [] but it was wired
to a function of type [F32, F32] -> [I32].
```

instead of emitting invalid wasm. It also holds across the entire existing corpus unchanged — 871 / 870 tests green with it armed — which is independent evidence that "a caller's import agrees with what it calls" is a real invariant of the fuser, not merely of the new bridge.

- **IMPACT:** type-system confusion → a silently invalid fused module. Downstream this is a parse/exec failure in every consumer (kiln, wasmtime, browsers), which is the amplification `discover.md` calls out.
- **CANDIDATE UCA:** `UCA-F-*` (fuser) — the fuser provides a wiring that is type-incorrect for the caller, at a time when the caller's own type was available and unchecked.

---

### Hypotheses examined and cleared

| # | Hypothesis | Outcome |
|---|---|---|
| 1 | `area_size as i32` wraps for areas > `i32::MAX`, giving a negative length | **Not a bug.** `memory.copy` interprets its length operand as unsigned; the wrap reproduces the intended value. A huge area traps rather than corrupting. |
| 2 | Overlapping copy when the caller's retptr aliases the callee's return area | **Not a bug.** `memory.copy` is specified to behave as if through an intermediate buffer (memmove semantics). |
| 3 | `dst_mem`/`src_mem` assumed equal on a `Direct` boundary | **Not a bug**, and deliberately so — the emitter uses `options.caller_memory` / `options.callee_memory` rather than assuming one index, so it stays correct even if `crosses_memory` and the memory indices ever disagree. |
| 4 | Callee returns a null return-area pointer after a failed realloc → copy from address 0 | **No divergence.** The unfused composition's canon lowering reads the same pointer; behaviour is preserved, which is the invariant that matters. |
| 5 | `retptr_local` index is wrong | **Guarded.** `is_retptr_shape` requires `caller.params.len() == callee.params.len() + 1`, so the retptr index is exactly `callee_params`. |
| 6 | A bridge collides with the P3 widening wrapper | **Mutually exclusive.** `has_widening` requires equal param *and* result counts; the bridge requires them to differ by one. Additionally the backstop skips `widening-wrapper` wirings, so the two cannot fight. |
| 7 | `signatures_match` returning `true` on a `resolve_target_function` error hides a failure | **No silent path.** `generate_direct_adapter` immediately calls `resolve_target_function` itself and propagates the error. |

---

**Gate:** fmt clean, clippy `-D warnings` clean, 871 tests (`--workspace`), 870 (`--all-features`), `rivet validate` PASS.

PR #396, opened 2026-09-05
fix(v0.54.0): size `use`d types exactly instead of guessing (SR-72, #393)

### LINKED ISSUE #393: same-memory bridge: callee return-area size is derivable from the WIT but not the core type — jess's cascade boundaries still unbridgeable after #390
## The #390 fix is merged and correct; jess's graph still cannot fuse on the same-memory path

Opening this separately because **#390 is closed as completed** and its own error text says *"file the signature pair on meld#390"* — which is now a closed issue. This is that follow-up, not a re-open request: what #390 set out to fix **is** fixed.

### Confirmed on the merged fix

Built `meld` from `origin/main` at `5e4593b` (v0.53.0, PR #392 merged 14:08) and re-ran jess's real graph — relay falcon **v1.134.1**, gale-nano 0.7.0, jess's `flight-app`:

| graph | strategy | result |
|---|---|---|
| `flight-app` + `rate` | `shared --pack-rebase` | **exit 1, refuses** |
| full 19-component composed stack | `shared --pack-rebase` | **exit 1, refuses** |
| full composed stack | `shared --address-rebase` | **exit 1, refuses** |
| full composed stack | `multi` | exit 0, valid |

**The invalid-wasm-at-exit-0 defect is gone.** That was the actual defect and it is genuinely closed — a loud refusal naming the boundary is strictly better than a module that validates nowhere.

The refusal is byte-for-byte the same as on the pre-merge branch:

```
cannot fuse the same-memory call to 'pulseengine:falcon-cascade/rate@0.7.0#tick':
the callee's return-area size is unknown to the resolver.
```

I checked the 56 lines added to `meld-core/src/lib.rs` after I tested `3440228`, so as not to report a regression that isn't one: they are the **`signatures_match` backstop**, explicitly documented as unreachable while the seam is correct. Inert for this path, by design. Nothing regressed.

### What remains

Both jess boundaries land in the unbridgeable branch. The signature pair, repeated here so this issue stands alone:

**Callee — lifted export, core type:**
```wat
(func $"pulseengine:falcon-cascade/rate@0.7.0#tick"  (param i32) (result i32))
(func $"pulseengine:falcon-cascade/mixer@0.7.0#mix"  (param f32 f32 f32 f32) (result i32))
```

**Caller — lowered import in `flight-app`:**
```wat
(canon lower (func $tick) (memory $memory))     ;; note: memory, and NO realloc
(canon lower (func $mix)  (memory $memory))
```

The callee returns a bare `i32` — a pointer into its own return area — and the area's **size is nowhere in the core signature**. It is only in the WIT:

```
torque-setpoint = 4 × f32 = 16 B    (rate#tick returns this)
motor-pwm       = 4 × f32 = 16 B    (mixer#mix returns this)
```

Both are fixed-size records of scalars — no strings, lists, or resources — so the size is **statically derivable from the component type** even though the core type hides it. If the resolver can reach the WIT type for the boundary it is already naming in the error message, the answer is 16 in both cases.

I don't know whether that's a small change or a layering problem in your resolver, so I'm leaving it as the observation rather than proposing a patch.

### Why this matters to jess specifically

`--memory multi` fuses and validates, but multi-memory has **no single-address-space lowering** (#172), so it cannot reach an RT1176 partition. The shared path is the only route to the target.

The **cascade alone is unaffected** and continues to fuse and lower cleanly under both rebase strategies — that is what relay's SWREQ-FALCON-OCI-P02/-P05 rest on and none of this touches it. What remains blocked is fusing an *application* with the cascade, which is the DD-026 composed image.

### One thing I want to acknowledge

The backstop's comment — *"an assert nobody has ever seen fire is worth exactly as much as the last time someone proved it could"*, with its potency established by disabling the bridge dispatch and confirming the fixture hard-fails there — is the same discipline jess has spent this campaign learning the expensive way. Two clean-room passes on jess found defects that were all in *checkers*, never in the code being checked. Establishing an assert's potency by making it fire is exactly the practice that would have caught them. Good to see it written into the source rather than assumed.

Happy to test any resolver change against this graph; the components are all published and `flight-app` is in jess at `app/flight-app`.

avrabe:
Fixes #393, reported by jess.

She re-ran her graph against the merged #390 fix and found her cascade boundaries still refuse. Chasing that refusal turned up a **much worse defect behind it**: the same root cause was **silently corrupting results** on `--memory multi` — the path she was using as her control.

## One guess, two symptoms

WIT's `use types.{torque}` compiles to an instance-export alias. Type resolution followed `Defined` and `ExportAlias` chains but stopped at `InstanceExportAlias`, so every `use`d record was unresolvable — and an unresolvable type was sized as **4 bytes**. A plausible number, which is exactly what made it dangerous.

| path | what the guess did |
|---|---|
| **cross-memory** | `return_area_byte_size.unwrap_or(8)` copied 8 bytes of a 16-byte record. The module **validated, ran without trapping, and returned a value with its later fields zeroed** — 7 instead of 26. |
| **same-memory** | the resolver recorded a size only `if size > 4`, so it stayed `None` and the SR-71 bridge refused. |

**The refusal jess reported is the only reason any of this was visible.** The "never guess the size" guard shipped in v0.53.0 is what converted a silent wrong answer into a report. That is the argument for loud-not-silent in one concrete case.

## Fix

- Capture instance types' exported value types at parse time — they were discarded as `ComponentTypeKind::Other`.
- Resolve `InstanceExportAlias` in a new `resolve_defined_val_type`, used for both size and alignment.
- Record a size only when every part of the type is exactly sizeable (`can_size_exactly`), instead of inferring "unknown" from the magnitude of a possibly-guessed result.
- An unknown return-area size is now a hard error on the cross-memory path too. `unwrap_or(8)` is gone.

## The bug inside the fix, and why the fixture is shaped as it is

My first implementation allocated a local type index for a type *declaration* but not for a type *export*. But this occupies an index:

```wat
(type   (;0;) (record ...torque-setpoint))
(export (;1;) "torque-setpoint" (type (eq 0)))   ← allocates index 1
(type   (;2;) (record ...motor-pwm))
(export (;3;) "motor-pwm" (type (eq 2)))
```

So every later `eq N` shifted: the first exported record resolved, the second didn't. **My fixture had one record and one export, so it passed. falcon's `types` interface exports two, and failed.**

The fixture is now two records with the function returning the **second**, and the negative control is recorded because it *is* the point:

| fixture | with the export index allocation removed |
|---|---|
| two records, returns the second | **fails** — catches it |
| one record (the original) | "Fusion complete!" — ships it |

The fixture was doing the work, not the assertion.

## Verified

```
compose_record_use   multi    7 -> 26        (was silent corruption)
compose_record_use   shared   refuse -> 26   (jess's report)
compose_record       multi    26 -> 26       (control, unchanged)
```

And on real inputs: falcon `cascade + ekf + position + attitude + rate + mixer` now fuses with **all 5 boundaries resolving**, and validates. Those five were previously truncating on the only path that appeared to work.

`both_strategies_agree` is included deliberately — fusion is semantics-preserving, so a strategy that changes the answer is a defect regardless of which one is "right".

## Traceability

SR-72 + SWV-84; `rivet validate` PASS; `rivet release status v0.54.0` cuttable.

Local gate: fmt clean, clippy `-D warnings` clean, 875 tests (`--workspace`), 874 (`--all-features`).


avrabe:
## Mythos delta-pass — 9 Tier-5 files

Auto-Mythos raised **1 finding**, on `parser.rs`, against my own fix. Disposition below, plus my own adversarial pass. Addressed in 09a98be.

---

### AUTO FINDING — instance-local type indices escaping their namespace

- **FILE:** `meld-core/src/parser.rs`
- **HYPOTHESIS (as raised):** values stored in `instance_type_exports` can contain `ComponentValType::Type(local_idx)` where `local_idx` indexes the *instance type's own* space. When `resolve_defined_val_type` returns such a value and the size walk recurses, it looks `local_idx` up in the component-level `component_type_defs` — a different namespace — so a coincidental entry answers as an impostor, `can_size_exactly` returns true, and the size is silently wrong. Same class as #393.

**Mechanism: CONFIRMED.** The encoding does carry those references. Building a fixture with a nested record:

```wat
(type (;4;) (record (field "seq" u32) (field "t" 3)))
```

`t` is a local index. Stored verbatim, it escapes.

**Exploit: NOT REPRODUCED.** I built the nested fixture specifically to trigger it and could not. Removing the substitution leaves it returning the correct 91. The reason is visible in the binary: in the provider, `command.t` references local index **1**, and the provider's *component-level* index 1 is `(alias export … "torque")` — **the same type**. The impostor is benign by coincidence, and I could not find a way to force the two spaces to disagree.

**Disposition: not a reportable defect; fixed anyway as hardening.** By `discover.md`'s own bar — *"if you cannot write both a Kani harness and a failing PoC, the finding does not count"* — this does not qualify, and I am not claiming to have fixed a demonstrated bug. But the substitution removes the class *by construction* rather than by luck, costs one pass at parse time, and regresses nothing. The alternative is relying on two index spaces continuing to coincide, which nothing checks.

The fixture keeps the nested record regardless, so the shape is now exercised rather than assumed absent.

- **IMPACT:** would be type-system confusion → silently wrong copy length. Unrealized.
- **CANDIDATE UCA:** `UCA-P-*` (parser) — provides a type resolution that is correct-looking but resolved against the wrong index space.

---

### My own pass over the delta

| # | Hypothesis | Outcome |
|---|---|---|
| 1 | `resolve_defined_val_type` alias chain can loop | **Guarded.** Bounded by `component_type_defs.len() + 1`, same as `get_type_definition`. |
| 2 | `can_size_exactly` recurses unboundedly on a cyclic type | **No.** After the substitution, stored values contain no `Type` at all, so the recursion cannot cycle through the alias hop. |
| 3 | The `4` fallback in `canonical_abi_size_unpadded` still reaches a copy length | **No.** `can_size_exactly` gates recording, and every adapter path now errors on `None`. The `4` can only reach code that is not sizing a copy. |
| 4 | Removing `unwrap_or(8)` breaks a legitimate small return area | **Behaviour change, deliberate and safe.** A previously-silent short copy is now a loud error. No test regressed (875/874). |
| 5 | The export-index allocation double-counts a non-type export | **No.** The push is inside `if let ComponentTypeRef::Type(..)`; func/instance exports do not touch the type space. |
| 6 | `TypeBounds::SubResource` mis-handled | **Handled** — resolves to `None` (a resource, not a value type), so it sizes as unknown rather than wrongly. |

---

### Loop steps, honestly

- **varve (step 0):** meld has **no `varve.toml`**. This loop ran **unpinned**, against ambient tools (`~/.cargo/bin` rivet 0.32.0, wasm-tools, wac; homebrew cosign). Stated rather than implied.
- **spar / WIT (1–2):** N/A — no architecture change. The fixture `.wit` is hand-written on purpose: it is a hermetic test input, not a product interface.
- **rivet (3):** SR-72 + SWV-84; advanced by `rivet verify SR-72` with evidence (1 link + 4 source markers), not by editing the status field. Filed **rivet#892** — `coverage --tests` attributes every marker to the *preceding* function.
- **witness (5):** ran and passed, but does **not** cover this feature — it instruments the fused artifact, not meld's own Rust. Third+ feature running, so filed as **meld#397** per the skill's recurring-N/A rule.
- **sigil (6):** covered by the release pipeline's cosign chain; verified at tag time.
- **clean-room (7):** launched, then **stopped and discarded**. I edited `parser.rs` and the fixture *after* launching it, so it was verifying a moving target; its results would have been meaningless. It also filled the disk with its own target dirs. My error in sequencing, recorded rather than quietly re-run.

PR #403, opened 2026-09-09
test(#397): enumerate the call-lowering seam instead of sampling it

### LINKED ISSUE #397: MC/DC covers the artifact meld produces, not meld itself — every defect found in v0.53.0/v0.54.0 lived where witness cannot see
Surfaced by the feature-loop skill's rule that a **recurring N/A is a backlog item, not an exemption** — witness has now been non-covering for four meld-internal features running (SR-69, SR-70, SR-71, SR-72).

## The gap

`Witness MC/DC (fused meld seam)` instruments **the fused output** — `tests/gate/composed.wasm` put through `meld fuse`, then instrument → run → report, gated on 0 gap rows. It is a genuinely good gate; `run-witness.sh` even guards against its own gap check going vacuous (assert `branch_count >= 1`, so a future dead-strip cannot silently make it pass). No complaint about the gate.

The observation is about **scope**. What it measures is the artifact meld *produces*. What it does not measure is meld's own decision logic — the parser, resolver, merger and adapter code that decides what that artifact contains.

And meld's own Rust has no MC/DC at all:

```
Coverage:  cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
```

Line coverage, no `--fail-under`, no branch or MC/DC flags. It reports; it does not gate.

## Why this matters more than it looks

Every defect found in the last two releases lived in meld's own Rust, in exactly the decision logic that has no MC/DC evidence:

| defect | where | shape |
|---|---|---|
| SR-71 / #390 | `generate_direct_adapter` + the #304 inline guard | a condition that was documented as a superset and was not |
| SR-72 / #393 | `canonical_abi_size_unpadded` `Type(idx)` arm | an unresolvable branch silently returning a plausible constant |
| (found mid-fix) | instance-type export index allocation | an off-by-one that resolved the first case and not later ones |

All three are *branch* defects: a path taken under a condition nobody had a vector for. That is precisely the class MC/DC exists to find, and precisely the class our MC/DC gate is pointed away from.

For most tools that trade-off is fine — you care about the output. But meld's correctness **is** the safety argument for everything downstream: kiln, wasmtime, the RT1176 image. The fused artifact being MC/DC-covered says nothing about whether the fuser chose the right bytes to put in it. SR-72 is the concrete proof: the fused module was **fully valid and MC/DC-clean, and returned the wrong number**, because the defect was in the decision that sized a copy — not in the emitted code's branch structure.

## What would close it

Not necessarily witness — meld is a native binary, not a Wasm component, so witness is arguably the wrong instrument. Options roughly in cost order:

1. **Give `Coverage` teeth**: `--fail-under-lines` at the current level as a ratchet, so coverage cannot silently regress. Cheap; catches nothing today but stops decay.
2. **Branch coverage on the decision-dense modules** — `parser.rs` canonical-ABI sizing, `resolver.rs` requirement derivation, `adapter/call_lowering.rs`, `adapter/fact.rs` adapter selection. `cargo llvm-cov --branch` exists; scope it to those paths rather than the workspace so the signal is not drowned.
3. **Real MC/DC on the seam that decides** — `resolve_call_lowering_plan` and `can_size_exactly` are small, pure, total functions over booleans and enums. They are the natural MC/DC targets and would be cheap to cover exhaustively.

(3) is the highest value per effort: those two functions are where the last two releases' defects were decided, and both are small enough to enumerate.

## Honest scoping

This is not a claim that the current gates are worthless — `Scry sound-AI`, the Mythos delta-pass, the LS-N gate and the fuzz targets all bite, and Mythos in particular found real bugs in both releases. It is a claim that the *MC/DC* evidence in the safety case currently supports a narrower statement than a reader would assume from "meld has an MC/DC gate", and that the difference matters for exactly the code that keeps producing defects.

Filed rather than fixed because choosing between (1)/(2)/(3) is a scope decision, not an implementation detail.

avrabe:
First concrete increment on #397 — option 3 of the three I listed there, chosen because it targets exactly where the defects have been.

## The seam

`resolve_call_lowering_plan` decides how **every** fused call is lowered. It is a total function over seven booleans — 128 combinations — and had nine example-based tests, each flipping **one** fact from a baseline.

#390 lived here: an inline guard that a comment claimed was a superset of the Direct generator's trigger, and was not.

## What this adds

- The whole truth table, checked against the **documented rules** rather than against the implementation.
- An independence check: each fact must be able to change the outcome on its own — otherwise the table is satisfied by a function that ignores one.

## What it actually buys, measured

Injecting a reachable defect that requires **two** flags set together (eligible when post-return *and* a resource rep conversion are both present — a combination no example sets):

| | result |
|---|---|
| all 9 example tests | **ok** — none caught it |
| `every_combination_obeys_the_documented_rules` | **FAILED** |

One-flag-at-a-time examples structurally cannot see interactions.

Worth stating narrowly: **on rules that already have an example, the table adds nothing.** The nine were well-chosen. The gain is interactions and future rule changes, not a claim that the previous tests were poor.

## A control that proved nothing, recorded

My first injection was gated on `crosses_memory`, which forces class `MemoryCopy` — so the class check masked it, the "defect" could never change behaviour, and everything passed. I nearly concluded the table did not bite.

That is the third control this month that passed for the wrong reason. The fix is the same each time: check the control *can* fail before trusting that it did not. Recording it here because the failure mode is more reusable than the test.

## Scope

This does not close #397. That issue is about MC/DC evidence pointing at the artifact meld produces rather than at meld's own decision logic; this covers one seam. The coverage-ratchet and branch-coverage options remain open there.

fmt clean, clippy `-D warnings` clean, **881** tests, exit 0.

### DIFF
--- a/meld-core/src/adapter/call_lowering.rs
+++ b/meld-core/src/adapter/call_lowering.rs
@@ -164,6 +164,174 @@ mod tests {
         }
     }
 
+    /// The whole truth table — all 128 combinations of the seven facts, checked
+    /// against the DOCUMENTED rules rather than against the implementation.
+    ///
+    /// #397: this seam decides how every fused call is lowered, and it is where
+    /// #390 lived — an inline guard that a comment claimed was a superset of the
+    /// Direct generator's trigger and was not. Nine example-based tests covered
+    /// nine of 128 points; the rule the defect violated was not one of them.
+    ///
+    /// The seam is a total function over booleans, so the table can simply be
+    /// written down. That is the argument #397 makes for meld's own decision
+    /// points, and this is the one that has already cost a release.
+    #[test]
+    fn every_combination_obeys_the_documented_rules() {
+        for bits in 0u8..128 {
+            let facts = BoundaryFacts {
+                crosses_memory: bits & 1 != 0,
+                needs_transcoding: bits & 2 != 0,
+                inline_adapters: bits & 4 != 0,
+                has_resource_rep_calls: bits & 8 != 0,
+                has_resource_new_calls: bits & 16 != 0,
+                has_post_return: bits & 32 != 0,
+                signatures_match: bits & 64 != 0,
+            };
+            let plan = resolve_call_lowering_plan(facts).unwrap();
+
+            // Class selection, exactly as the doc comment states it.
+            let expected_class = if facts.needs_transcoding {
+                AdapterClass::Transcode
+            } else if facts.crosses_memory {
+                AdapterClass::MemoryCopy
+            } else {
+                AdapterClass::Direct
+            };
+            assert_eq!(plan.class, expected_class, "class for {facts:?}");
+
+            // An encoding mismatch outranks memory: transcoding is required
+            // whether or not the boundary crosses a memory (#361).
+            if facts.needs_transcoding {
+                assert_eq!(plan.class, AdapterClass::Transcode, "#361: {facts:?}");
+            }
+
+            // Inline eligibility, as the doc comment states it.
+            let expected_inline = facts.inline_adapters
+                && expected_class == AdapterClass::Direct
+                && !facts.has_resource_rep_calls
+                && !facts.has_resource_new_calls
+                && !facts.has_post_return
+                && facts.signatures_match;
+            assert_eq!(
+                plan.inline_eligible, expected_inline,
+                "inline for {facts:?}"
+            );
+
+            // The invariants the generators rely on, stated independently of
+            // the formula above so a wrong formula cannot satisfy both.
+            if plan.inline_eligible {
+                assert_eq!(
+                    plan.class,
+                    AdapterClass::Direct,
+                    "only Direct is ever inlinable: {facts:?}"
+                );
+                assert!(
+                    facts.inline_adapters,
+                    "inlined with inlining off: {facts:?}"
+                );
+                assert!(
+                    facts.signatures_match,
+                    "#390: inlined a boundary needing a convention bridge: {facts:?}"
+                );
+                assert!(
+                    !facts.has_post_return,
+                    "inlining would skip post-return cleanup: {facts:?}"
+                );
+                assert!(
+                    !facts.has_resource_rep_calls && !facts.has_resource_new_calls,
+                    "inlining would skip resource conversions: {facts:?}"
+                );
+            }
+        }
+    }
+
+    /// Each fact must be able to change the outcome on its own — otherwise the
+    /// table above is satisfied by a function that ignores it. Flips one bit at
+    /// a time from a baseline where every fact is live.
+    #[test]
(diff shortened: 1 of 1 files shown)


avrabe:
## Mythos delta-pass — `meld-core/src/adapter/call_lowering.rs`

Auto-Mythos: **NO FINDINGS**. My own pass below.

**NO FINDINGS.**

The delta is test-only — `resolve_call_lowering_plan` and `BoundaryFacts` are untouched, and `git diff` on this file shows additions inside `mod tests` exclusively. There is no production behaviour to attack, so the usual hypothesis priors (transcoding bounds, resource-table overflow, string canonicalisation) have no surface here.

What I did check, since a test-only change can still be *wrong* in ways that matter:

| # | Hypothesis | Outcome |
|---|---|---|
| 1 | The truth table asserts the implementation back to itself, so any change to the seam trivially still passes | **No.** `expected_class` and `expected_inline` are written from the doc comment's stated rules, then the invariants the generators rely on are asserted *separately* — so a wrong formula cannot satisfy both halves. |
| 2 | The 128-combination loop passes vacuously — e.g. an early `continue`, or a bit-mask that never sets some facts | **No.** Verified by injection: a reachable defect makes it FAIL while all nine example tests pass. `every_fact_independently_affects_the_outcome` additionally pins that each fact can change the outcome alone. |
| 3 | `every_fact_independently_affects_the_outcome` could pass while the seam ignores a fact | **Guarded.** It asserts the baseline IS eligible first; if the baseline stopped being the eligible case, every flip would trivially remain ineligible and the test would prove nothing. That assert is why. |
| 4 | Enumerating 0..128 over 7 bits mis-indexes, silently testing 64 real combinations twice | **No.** Seven distinct masks (1,2,4,8,16,32,64) over `0u8..128`; the injection control fires on a combination only reachable at bit 8+32, which would be unreachable under a mis-mask. |
| 5 | Added tests slow the suite enough to mask a timeout elsewhere | **No.** 128 iterations of a pure boolean function; the file's tests run in 0.00s. |

### One thing worth recording

My **first** injection control passed everything, and I nearly reported that the truth table did not bite. It was gated on `crosses_memory`, which forces class `MemoryCopy`, so the class check masked it — the injected "defect" could never change behaviour. An unreachable defect proves nothing about a test.

That is the third control this month that passed for the wrong reason. It is also exactly the failure this PR is about: a check that cannot fail looks identical to a check that passes.

- **IMPACT:** none — test-only.
- **CANDIDATE UCA:** n/a. The relevant risk is to the *evidence*, not the artifact: a vacuous test would leave `UCA-F-*` (fuser provides an incorrectly lowered call) covered on paper and uncovered in fact. Hypotheses 1–4 exist to rule that out.

**Gate:** fmt clean, clippy `-D warnings` clean, 881 tests (`--workspace`), exit 0.

PR #404, opened 2026-09-09
test(#397): enumerate the exact-sizing seam

avrabe:
Second seam, after `call_lowering` in #403.

## Why this one

`can_size_exactly` decides whether a type's size may reach a `memory.copy` length — the gate #393 turned on. Its structure is quietly fragile:

```rust
_ => true
```

A container variant that forgets to recurse falls into that arm and reports its contents as exactly sizeable. The resolver records a size built from the 4-byte fallback, and the copy runs short. **That is #393 exactly, one level down, with nothing failing.**

It is also the shape auto-Mythos flagged in my own #393 fix — a type escaping the check that was supposed to catch it.

## Four tests plus a tripwire

- an unresolvable type must poison **every** container that can hold one, checked per variant so a failure names which;
- **the mirror** — a resolvable type must stay sizeable in every container. Without this, a `can_size_exactly` that simply returned `false` would satisfy the first test while disabling sizing entirely;
- nesting must not launder it (three deep, not one);
- leaves stay sizeable by construction.

The tripwire: `variant_name` matches every `ComponentValType` variant with **no wildcard**, so adding a variant breaks this test and forces a decision about whether the sizing seam must recurse into it. Given `_ => true`, silence is otherwise the default.

## Control

Checked for reachability *before* running it this time. Removing the `Option` arm so it falls through to `_ => true`:

```
an_unresolvable_type_poisons_every_container          FAILED
  #397/#393: Option does not propagate un-sizeability
nesting_does_not_launder_an_unresolvable_type         FAILED
a_resolvable_type_stays_sizeable_in_every_container   ok      <- not a blanket-false
leaves_are_sizeable_by_construction                   ok
```

Naming the variant matters: the failure tells you which arm to fix, not just that something is wrong.

fmt clean, clippy `-D warnings` clean, **887** tests, exit 0.

### DIFF
--- a/meld-core/src/parser.rs
+++ b/meld-core/src/parser.rs
@@ -6287,3 +6287,139 @@ mod tests {
         assert_eq!(comp.flat_count(&flags_ty.unwrap()), 1);
     }
 }
+#[cfg(test)]
+mod sizing_seam_tests {
+    use super::*;
+
+    /// A type index that resolves to nothing on an empty component.
+    const UNRESOLVABLE: ComponentValType = ComponentValType::Type(9999);
+
+    fn resolvable() -> ComponentValType {
+        ComponentValType::Primitive(PrimitiveValType::U32)
+    }
+
+    /// Compile-time tripwire (#397). Deliberately has NO wildcard arm: adding a
+    /// variant to `ComponentValType` breaks this test, which forces whoever adds
+    /// it to decide whether `can_size_exactly` must recurse into it.
+    ///
+    /// That matters because `can_size_exactly` ends in `_ => true`. A new
+    /// container variant would fall into it and report "exactly sizeable" while
+    /// holding a type nothing can size — reintroducing #393 one level down,
+    /// silently, with no test failing.
+    fn variant_name(ty: &ComponentValType) -> &'static str {
+        match ty {
+            ComponentValType::Primitive(_) => "Primitive",
+            ComponentValType::String => "String",
+            ComponentValType::List(_) => "List",
+            ComponentValType::FixedSizeList(_, _) => "FixedSizeList",
+            ComponentValType::Record(_) => "Record",
+            ComponentValType::Flags(_) => "Flags",
+            ComponentValType::Variant(_) => "Variant",
+            ComponentValType::Tuple(_) => "Tuple",
+            ComponentValType::Option(_) => "Option",
+            ComponentValType::Result { .. } => "Result",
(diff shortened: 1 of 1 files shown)


avrabe:
## Mythos delta-pass — `meld-core/src/parser.rs`

Auto-Mythos: **NO FINDINGS**. Mine below.

**NO FINDINGS.**

Test-only delta — `can_size_exactly`, `resolve_defined_val_type` and the sizing functions are untouched; the addition is a `#[cfg(test)] mod sizing_seam_tests` at file scope. No production surface, so the discover.md priors have nothing to attack here.

What I checked instead, since a test can be wrong in ways that matter:

| # | Hypothesis | Outcome |
|---|---|---|
| 1 | The tests assert the implementation back to itself | **No.** They assert a property — un-sizeability propagates through containers — stated independently of how `can_size_exactly` is written. |
| 2 | `an_unresolvable_type_poisons_every_container` passes vacuously, e.g. `Type(9999)` resolving to something | **No.** Asserted directly on the bare type first, and the injection control fires. |
| 3 | A `can_size_exactly` that just returned `false` would satisfy it | **Covered** — that is exactly what `a_resolvable_type_stays_sizeable_in_every_container` exists to rule out, and it stays green under the injection. |
| 4 | `variant_name`'s wildcard-free match could be defeated by a `#[non_exhaustive]` enum, silently allowing a new variant | **No.** `ComponentValType` is a local enum without `#[non_exhaustive]`; adding a variant is a compile error in this file. Verified by reading the definition, not assumed. |
| 5 | The module at file scope could shadow or collide with the existing `mod tests` in this file | **No.** Distinct name, `#[cfg(test)]`, no `pub use`. 887 tests pass with no name conflicts. |

### Control

Removing the `Option` arm so it falls to `_ => true` fails two of four tests and names the variant:

```
#397/#393: Option does not propagate un-sizeability
```

while the mirror test stays green. Reachability was checked **before** running it this time — the arm's removal demonstrably changes the result for `Option(UNRESOLVABLE)` — after three controls this month that passed for the wrong reason.

- **IMPACT:** none — test-only.
- **CANDIDATE UCA:** n/a directly. The risk guarded is to the *evidence*: a silent `_ => true` on a future container variant would leave `UCA-P-*` (parser provides a type resolution that is wrong) uncovered while appearing covered. Hypotheses 1–4 exist to rule out that this guard is itself vacuous.

**Gate:** fmt clean, clippy `-D warnings` clean, 887 tests (`--workspace`), exit 0.

Issue #400, opened 2026-09-07
Fused core modules drop WIT type info (component-type* sections) — runtimes can't invoke typed exports


avrabe:
## The gap

`meld fuse` preserves the **export names** beautifully — the fused core module exports each component export under its canonical WIT name, e.g. `pulseengine:scry/analyzer@0.1.0#analyze`. That's what let kiln wire up `--invoke` on the fused-core path with no new contract (kiln SR-58), and it's genuinely the right design.

But the fused module carries **no WIT type information**, so a runtime can resolve the export and still not know how to *call* it.

```
$ wasm-tools component wit scry.core.wasm
error: input is a core wasm module with no `component-type*` custom sections
       meaning that there is not WIT information
```

Custom sections in the fused output are only: `.debug_*`, `producers`, `target_features`, `wsc.transformation.attestation`, `component-provenance` (function→originating-component attribution, not types).

## Why it matters

Verified with the real released `scry-3.3.0-wasm32-wasip2.wasm` (pulseengine/scry) fused with meld 0.41.3:

- WIT signature: `analyze: func(module-bytes: list<u8>, config: analysis-config) -> result<analysis-result, analyze-error>` — **2 component-level params**
- What a runtime sees on the fused module: a core function taking **10 flattened i32s** (the canonical ABI flattens `list<u8>` → ptr+len and the record into scalars)

So kiln can resolve `analyze` but cannot lower arguments for it: constructing a `list<u8>` requires allocating guest memory via `cabi_realloc` and writing bytes, which needs the *types*. Without them the only honest behaviour is to refuse (which kiln now does — it fails loud rather than zero-filling the 10 params and fabricating a result).

Concretely this blocks running pulseengine's own wasm-only tools through a layer's kilnd — scry is a library/reactor component, so "invoke a named export with typed args" is the *only* way to use it.

## Ask

Preserve or emit the WIT types in fused output, so a runtime can lower typed arguments:

1. **Preferred — keep the `component-type*` custom sections.** This is the standard `wit-bindgen` convention for embedding WIT in a core module, so `wasm-tools component wit` and any runtime already know how to read it. If fusion invalidates the original sections, re-emit one describing the fused module's exported interface.
2. Or emit a compact signature manifest (custom section / sidecar) mapping each exported canonical name → its WIT signature.

Either way the contract stays where it already works well: meld owns lowering and records what it produced; the runtime consumes it.

Happy to consume whichever shape you pick on the kiln side — tracked there as SR-59.

[...]

PR #424, opened 2026-09-18
feat(v0.57.0): say how to invoke a fused export, and count `use`d types properly (SR-79/SR-80, #400 #423)

### LINKED ISSUE #423: flat_count returns 1 for every `use`d type, so the params-ptr calling convention is chosen from a wrong number
#393 fixed the **size** of a type reached through `use`. It left the **flat count** — and the slot collectors — still resolving through `get_type_definition`, which by its own documentation stops at the instance-export alias that `use` compiles to. Every `use`d type therefore counts as **1 flat value**, and meld picks the calling convention from that number.

## The number is wrong on a released artifact

`falcon-rate-v1.138.wasm` (relay `falcon-v1.138.0`), whose `rate#tick` takes `vehicle-state` (14 × f32) + `rate-setpoint` (4 × f32), both reached through `use types.{...}`:

| source | flattened params |
|---|---|
| Canonical ABI reference implementation (`definitions.py`, `flatten_functype(ft,'lift')`) | **18** |
| the emitted core signature in the component | `(param i32) (result i32)` — i.e. over the limit, pointer-in |
| **meld's `total_flat_params`** | **2** |

Two `use`d records, one flat value each.

## Cause

`parser.rs:3189-3198`, `flat_count`:

```rust
ComponentValType::Type(idx) => {
    if let Some(ct) = self.get_type_definition(*idx) {
        if let ComponentTypeKind::Defined(inner) = &ct.kind { self.flat_count(inner) } else { 1 }
    } else { 1 }        // <- every `use`d type lands here
}
```

`get_type_definition`'s own doc (`parser.rs:1564-1566`) says it "Returns `None` for import-defined or instance-export-alias types". #393 added `resolve_defined_val_type`, which follows `ExportAlias` **and** `InstanceExportAlias`, and its doc says callers "MUST treat that as *unknown*, never as a default size: guessing is what made #393 a silent wrong answer".

Sites that were converted to the resolver: `parser.rs:3262`, `:3393`, `:3414` (element size, alignment, return-area byte size).

Sites still on `get_type_definition` with a silent fallback:

| site | fallback | consequence |
|---|---|---|
| `:3189` `flat_count` | `1` | wrong flat count |
| `:1791` `flat_width_list` | `vec![4]` | wrong flat widths |
| `:2149` `collect_return_area_type_slots` | one 4-byte i32 slot | wrong return-area slots |
| `:2206` resource handle detection | `None` | `use`d resource handles not seen |
| `:2357` `collect_resource_byte_positions` | skipped | " |
| `:2525`, `:2575` `collect_pointer_positions{,_with_layout}` | skipped | `use`d string/list pointer pairs not seen |

## Why the count matters

`resolver.rs:3090` and `:3416` choose the calling convention from it:

```rust
let total_flat = to_component.total_flat_params(comp_params);
if total_flat > 16 {
    // params-ptr: allocate a params area, pass one pointer
```

For falcon's shape the real count is 18 and meld computes 2, so it takes the flat branch for a callee whose core signature is `(param i32)`. `--memory multi` does not validate by default, so the failure mode to establish is whether that is a loud invalid module or a silent wrong call.

Also visible inside one artifact: fusing the five falcon stages with `--emit-manifest` (#400) reports `flat_param_count: 2` for an export whose `core` — read back from the same emitted bytes — is `(i32) -> (i32)`. The manifest's own cross-check catches meld disagreeing with itself.

## Not yet established

Whether a fused artifact **today** carries the wrong convention end to end. The falcon stages do not import each other (each imports only `types`), and the shipped composed cascade arrives pre-linked as a single core module, so neither reaches meld's adapter path. The existing `use`d-type fixture (`compose_record_use`) has 4-field records, whose true count is 4 — under the limit either way, which is why #393's size fix was sufficient there.

The fixture that would settle it is a `use`d-record call whose true flat count crosses 16, built the same offline way (`wasm-tools` + `wac`, no cargo-component). I am building that next and will report the observed failure mode rather than assume it.

## Fix

Route the six sites above through `resolve_defined_val_type` and treat `None` as unknown — no `1`, no `vec![4]`, no skipped slot — per that function's contract.

avrabe:
Closes #423. Advances #400 (increment 1 of the shape kiln accepted).

Building #400's manifest against **released artifacts** exposed a live defect in meld, so this PR carries both. The defect is the more serious half.

---

## #423 — a `use`d type counted as one flat value, and the fused module returned a wrong answer at exit 0

#393 fixed the **size** of a type reached through `use` by routing the size paths through `resolve_defined_val_type`, which follows the instance-export alias a `use` compiles to. **Sixteen other traversals** still resolved through `get_type_definition`, which stops at that alias — so a `use`d record was reported as one flat value of four bytes, as containing no pointers, and as containing no resources.

meld picks the calling convention from that count (`resolver.rs`: `total_flat_params(..) > 16` selects params-ptr). jess's shape — `tick(state: 14 × f32, sp: 4 × f32)`, every record `use`d — is **18** flattened values and counted as **2**.

**The failure mode, established rather than assumed.** The existing `use`d-type fixture cannot reach this (4-wide records: wrong count 1, right count 4, both under the limit — which is why #393's size fix was sufficient there). A new offline fixture crosses it:

| | before | after |
|---|---|---|
| `total_flat_params` | 2 | **18** |
| fused module validates | ✅ yes | ✅ yes |
| `run()` under wasmtime | **0.0** | **120.0** |

It validated, ran without trapping, and returned the wrong number. Exit 0.

**Every `get_type_definition` call site in the workspace was classified** — value-type traversals converted, function-type lookups left alone. That mattered: the full suite then failed in `wit_bindgen_runtime`, because `collect_param_pointer_wit_types` in `resolver.rs` was a **seventeenth** site still stopping at the alias, and meld's own parallel-list assertion fired:

```
param_wit_types must be parallel to pointer_pair_positions (one per pointer pair)
  left: 0   right: 1
```

The tripwire earned its keep — that would otherwise have been a silent divergence between two lists the emitter indexes in lockstep.

### The fixture

`tests/wit_bindgen/fixtures/compose_record_use_wide/`, built by its `build.sh` from WAT with `wasm-tools` + `wac`, fully offline. `wasm-tools component new` refused the provider until it exported `cabi_realloc` — the tooling stating the same requirement as #400's finding 1: over the limit, the **callee's** allocator provides the argument area.

---

## #400 — `--emit-manifest`

Per export: the WIT signature, the core signature, `flat_param_count`, `needs.{memory,realloc}`, the memory / allocator / post-return to call, and the return area's size, alignment and top-level layout. Field names and the per-export allocator choice are kiln's, [confirmed on #400](https://github.com/pulseengine/meld/issues/400#issuecomment-5725195817).

**`core` is read back from the emitted bytes**, never derived from the WIT. That is the whole point: derived from one source the two would agree by construction, kiln's cross-check would always pass, and the field would be decoration.

```json
{
  "export": "golden:wide/lib@0.1.0#tick",
  "wit": { "params": [["s", "record { f1: f32, … f14: f32 }"], ["sp", "record { r1: f32, … }"]],
           "result": "record { o1: f32, o2: f32, o3: f32, o4: f32 }" },
  "core": { "params": ["i32"], "results": ["i32"] },
  "flat_param_count": 18,
  "needs": { "memory": true, "realloc": true },
  "memory": "memory", "realloc": "cabi_realloc", "post_return": null,
  "return_area": { "size": 16, "align": 4,
                   "layout": [{"offset":0,"field":"o1","size":4}, …] }
}
```

**Two refusals, deliberate:**
- An export whose types do not fully resolve is listed under `omitted` **with a reason**, not described by a fallback. A fallback is how a `use`d record was once silently sized at 4 bytes.
- `realloc` is **resolved through fusion**, not matched by name. Where that allocator is not exported, the entry says `"realloc": null` while `needs.realloc` stays `true`. kiln refuses on that pair rather than calling a different exported allocator — for the falcon cascade (five components, five 8 KB arenas, only mixer's exported) that is the honest state today, and increment 2 is what flips it.

Opt-in, because the section adds bytes. New `meld docs signature-manifest` topic.

### Checked against three independent sources, on released artifacts

relay `falcon-v1.138.0` (SHA256SUMS verified) and scry `3.3.0` (cosign verified): meld's emitted numbers match the **Canonical ABI reference implementation** applied to the WIT, and — for the five falcon stages — the guests' own `__RET_AREA` symbols:

| export | flat | meld return area | guest `__RET_AREA` |
|---|---|---|---|
| `ekf#estimate` | 6 | 56 | **56** |
| `position#tick` | 18 | 20 | **20** |
| `attitude#tick` | 19 | 16 | **16** |
| `rate#tick` | 18 | 16 | **16** |
| `mixer#mix` | 4 | 16 | **16** |

## Tests

`signature_manifest_400.rs` (5) over a fixture containing **both** branches, so neither is asserted vacuously: a wide export needing memory + allocator, and a runner needing neither.
- `core_matches_the_emitted_module` — independent re-parse of the fused bytes
- `core_comes_from_the_module_not_the_wit` — the control: the WIT side is two records of 14 and 4 floats, which would render as 18 flat values if `core` came from it; `core` is the single pointer the module takes, which only the bytes know
- `a_wide_export_states_the_pointer_shape_and_its_allocator` — 18 vs one `i32`, `needs.realloc`, the 16/4 area with named layout, and that `realloc` really is an export of this module
- `an_export_that_needs_nothing_names_nothing`, `the_section_is_absent_unless_asked_for`

`wide_used_params_423.rs` (2) — the count, and the runtime oracle above.

## Traceability & gate

- SR-79 / SWV-92, SR-80 / SWV-91 via rivet; both `verified`, scoped `release: v0.57.0`, `rivet release status v0.57.0` cuttable
- `tools/check_record_citations.py` (merged in #422): OK, 426 citations resolve
- `cargo fmt --check` 0 · `clippy -D warnings` 0 · `test --workspace` 0 · `test --workspace --all-features` 0

Version 0.56.2 → **0.57.0** (new capability + behaviour fix). Note: SR-28's exhaustive-destructure sentinel forced the new config field to be classified; it is a separate self-describing section, like provenance, so not a build parameter.


avrabe:
## Mythos delta-pass

Tier-5 files in this diff: `parser.rs`, `resolver.rs`, `component_wrap.rs`, `p3_stream.rs`, plus the new `signature_manifest.rs`.

### Finding 1 — the manifest named the wrong memory (REAL, fixed in b867dcf)

`memory` was filled with the **first** exported memory for every entry. A multi-memory fusion exports one per input component, so half the entries pointed a host at **another component's memory**. Writing an argument area into the wrong memory is the silent-corruption class this manifest exists to remove — reproduced inside the manifest itself.

Measured on two falcon stages fused with `--memory multi`:

| export | before | after |
|---|---|---|
| `mixer@0.10.0#mix` | `memory` | `memory` |
| `rate@0.10.0#tick` | **`memory`** ❌ | **`memory$0`** ✅ |

`memory` is now resolved like `realloc`: from the lift's own `memory` canonopt, through the merger's `memory_index_map`, to the name the fused module exports. Unresolvable stays `null`.

Pinned by `each_export_names_its_own_memory`, on two in-repo provider fixtures that contribute a memory each. **Control:** reverting to "first exported memory" fails that test and only that test.

### Finding 2 — a seventeenth unconverted site (REAL, already in the PR)

`collect_param_pointer_wit_types` in `resolver.rs` still stopped at the `use` alias after the parser sweep, so `param_wit_types` and `pointer_pair_positions` — two lists the emitter indexes in lockstep — diverged. meld's own `debug_assert` caught it in `wit_bindgen_runtime`, loudly. Every `get_type_definition` call site in the workspace was then classified rather than fixing the one the compiler pointed at.

### Probes that came back clean, with evidence

| probe | result |
|---|---|
| cost of the manifest on the largest real component (scry 3.3.0, 4.5 MB) | no measurable difference: 0.71 s without, 0.52 s with — the per-lift map rebuild is not hot because components have few lifts |
| determinism under `--reproducible` | two runs byte-identical (`db718bc79f00fdcd…`) |
| `--component` output format | fuses at exit 0, section present |
| scry's entry vs the draft computed independently on #400 | identical: `flat=10`, `needs{memory,realloc}`, `realloc=cabi_realloc` (this time resolved through fusion, not name-matched), `post_return` present, return area 128 |
| falcon's five stages vs the Canonical ABI reference implementation **and** the guests' own `__RET_AREA` symbols | all three agree (56/20/16/16/16) |
| an export needing nothing | names nothing: `memory`, `realloc`, `return_area` all `null` |

### Known limits, stated rather than fixed here

- `flat_count` still falls back to **1** for a type that is genuinely unresolvable (not a `use` alias — e.g. import-defined). The `use` case is fixed; making a genuinely-unknown type loud is a behaviour decision with its own blast radius, and belongs in its own change rather than riding along here.
- Value-type recursion (`render_wit`, and the existing canonical-ABI walks it mirrors) has no depth guard. This diff does not add exposure — it follows the same shape as the walks already there — but a deeply nested hostile type would recurse in both.

### What I did not test

An end-to-end run of a **fused** artifact whose cross-component call carries `use`d string or list parameters, where the newly-visible pointer positions change adapter emission. The fixtures here are f32 records. The conversion is uniform and the suite is green, but "the pointer-position collectors now see `use`d types" is exercised by construction, not by a runtime oracle.

Issue #411, opened 2026-09-16
Evidence citations name four nonexistent tests — three of them name invariants the code now asserts the opposite of


avrabe:
Stale test names in the safety record are ordinary drift. These are not ordinary: three of the four name a behaviour the codebase has since **deliberately inverted**, so the citation asserts the opposite of what the code does.

```
$ for t in inc5c_b_async_deeper_list_nesting_param_fails_loud_all_directions \
           stackful_lift_is_async_without_callback \
           auto_selects_shared_for_growfree_inputs; do
    grep -rl "fn $t" --include='*.rs' . | wc -l
  done
0
0
0
```

Each appears in `sw-verifications.yaml`, `traceability.yaml`, and `docs/verification-matrix.md`.

## (a) SWV-41 — cites a fail-loud test for a case that is now allowed

Cited: `inc5c_b_async_deeper_list_nesting_param_fails_loud_all_directions`.

Successor: `meld-core/src/adapter/fact.rs:16405` `d5d_async_deeper_list_nesting_param_allowed_all_directions`, whose own doc says:

> DEPTH-2 list nesting … is now **ALLOWED** in every implemented direction … **Before #286 5d this asserted fail-loud; that invariant is intentionally inverted.**

SWV-41 is `status: verified` citing a test that asserted the refuted invariant.

## (b) SWV-32 / SR-32 — name encodes a derivation the code calls wrong

Cited: `stackful_lift_is_async_without_callback`. Actual: `meld-core/src/p3_async.rs:1686` `stackful_lift_is_existential_over_lifts`, and `p3_async.rs:1723-1726` explains why the old name is wrong:

> Prior to the fix, `uses_stackful_lift` was derived as `uses_async_lift && !uses_callback_lift`, which **mis-classified mixed-mode components**.

The citation names the mis-classifying derivation.

## (c) SR-37 — the cited test asserts what SR-37 forbids

`traceability.yaml:499` and `docs/verification-matrix.md:47` cite `auto_memory.rs::auto_selects_shared_for_growfree_inputs` as SR-37 evidence. It does not exist. `loss-scenarios.yaml:3473-3476` goes further and describes what it supposedly proved:

> `auto_selects_shared_for_growfree_inputs` covers the other direction end-to-end: grow-free inputs produce a **single-memory output** … (B's data at its **rebased base**)

[...]

PR #425, opened 2026-09-18
fix(v0.58.0): an export that needs the guest allocator can reach one (SR-81, #400)

avrabe:
Advances #400 (increment 2, the shape kiln chose).

This started as "add the missing allocator exports so kiln can invoke three falcon exports". Building it found that the missing allocator is a **defect**, and one that has nothing to do with the manifest.

## An export nothing can call

```
$ meld fuse provider.wasm --memory shared --address-rebase
$ wasm-tools print out.wasm | grep '(export'
  (export "memory" (memory 0))
  (export "golden:wide/lib@0.1.0#tick" (func 1))
```

`tick` takes 18 flattened floats. Above `MAX_FLAT_PARAMS = 16` the Canonical ABI stages the argument tuple in an area the **callee's** allocator provides — and there is no allocator export. **No host can invoke this export**, with or without a manifest.

**Cause 1 — it was pruned.** #298 drops the vestigial `cabi_realloc*` exports so a downstream pass can eliminate `memory.grow`. Its safety check reads:

```rust
if params.iter().any(|(_, t)| comp.type_contains_pointers(t)) || … { return false; }
```

"No parameter type *contains* a pointer" is not "no allocator needed". A record of 17 `f32` contains no pointer and still crosses the flat limit. The check now also asks `total_flat_params(params) > MAX_FLAT_PARAMS`.

**Cause 2 — it was never exported.** In shared memory every component names its allocator `cabi_realloc`, so only one can hold that name; the others were simply absent. On the five-stage falcon cascade that is three of five exports. Where several are needed, meld now exports them as `meld:realloc/<n>` — deliberately **not** `cabi_realloc$<n>`, which is exactly the shape #298 prunes.

Both are gated the same way they always were: a fuse that needs no allocator keeps the existing pruning, so artifacts that do not invoke such exports are unchanged.

## Both controls were observed to fail

An earlier control for this work **passed**, which meant the test could not fail — the fixture path already exported the allocator. That is the trap the standing discipline names, so both controls here were run and watched:

| control | result |
|---|---|
| revert the #298 predicate fix | `a_needed_allocator_is_not_pruned` FAILS — *"this module exports none: [golden:wide/lib@0.1.0#tick]"*; the other test stays green |
| disable the added allocator exports | `every_export_that_needs_an_allocator_can_call_one` FAILS — *"needs an allocator and the manifest names none"*; the other stays green |
| both restored | 2 passed |

Each mechanism is load-bearing, and each is pinned by its own test.

## The manifest is now actionable end to end

`every_export_that_needs_an_allocator_can_call_one` does not stop at "a name is present": it **calls through it**. Allocate the argument area with the allocator the manifest names, write 18 floats into the memory it names, invoke the export, read `sum(1..=14) = 105` out of the return area at the offset it describes. `invoke_via_manifest_400.rs` does the same for the multi-memory case, where naming the first memory or the first allocator would silently target another component's.

New fixture `provider2.wasm` (a second wide provider under its own package) makes the "two components both need an allocator" case reproducible in-repo, rather than only on downloaded release artifacts.

## A false claim in the record shipped yesterday

SWV-92 (v0.57.0) said the wide fixture is "built by its build.sh". **That script did not exist** — the fixture had been built by hand with the same commands. `build.sh` is written here, verified to reproduce the committed artifacts byte-identically (`composed_wide_use.wasm` and `provider.wasm` unchanged by a rebuild), and SWV-92 records the correction. Same class as #411, in my own entry.

## Behaviour change worth knowing

On the affected paths a fused module now carries an allocator export it previously dropped, so that allocator — and its `memory.grow` — stay reachable for a downstream DCE pass. That is the point: you cannot both invoke such an export and eliminate the allocator it requires. A build that does not need to invoke them is unchanged.

## Traceability & gate

- SR-81 / SWV-93 via rivet; `rivet verify SR-81` (1 link + 3 markers); scoped `release: v0.58.0`; `rivet release status v0.58.0` cuttable
- `tools/check_record_citations.py`: OK, 426 citations resolve
- `cargo fmt --all -- --check` **0** · `clippy --workspace --all-targets -D warnings` **0** · `test --workspace` **0** · `test --workspace --all-features` **0**

Version 0.57.0 → **0.58.0**. No Tier-5 files touched (`lib.rs` and the new module are not on the list), so no Mythos delta-pass is required.


avrabe:
`Fuzz smoke (fuzz_merger_idempotent)` is red on this PR. **It is not caused by this change**, and it is not the known build-setup flake — the failing step is `Run target for 60 s`, and the summary is `libFuzzer: out-of-memory` on a 60-byte input.

Checked rather than assumed:

- The fuzz target runs parse → resolve → merge twice. It never calls the predicate this PR changes, and never emits a manifest, so nothing here is on that path.
- Driving the same three phases directly: parse 0.93 ms, resolve 0.44 ms, **merge 94.5 s** then `Err`, on a 48-byte core module.
- Same input against **v0.56.2 (`55427cf`), before #423**: **93.5 s**. Identical shape, so v0.57.0's `use`-resolution change did not introduce it either.

Filed as #426 with the input and the phase timings. `Fuzz Smoke` is not a required context (verified empirically on #418), so this PR is mergeable; I am recording the failure here rather than letting it pass unexplained.

PR #430, opened 2026-09-22
fix(v0.58.2): bound a declared segment count by its section (SR-83, #426)

### LINKED ISSUE #426: Merger burns ~95 s (and OOMs under a 2 GB limit) on a 60-byte malformed component, then fails anyway
`fuzz_merger_idempotent` found an input where `Merger::merge` runs for **94.5 seconds** on a **48-byte** core module and then returns an error. Under libFuzzer's default 2 GB allocation limit that is an OOM, so CI reports it as a crash.

```
SUMMARY: libFuzzer: out-of-memory
Failing input: fuzz/artifacts/fuzz_merger_idempotent/oom-22f43f6dd0a90a1115c4b8c27c3f9d622f53b37e
```

## Reproduction

The whole input, 60 bytes:

```
AGFzbQ0AAQAWAAEwAGFzbQEAAAAbG4EAGxsbAAAAAAAN3BubAYwJCfb2AA3cG5sBjAkJ9vb2PwkAABsN
```

```bash
base64 -d > oom.bin <<< 'AGFzbQ0AAQAWAAEwAGFzbQEAAAAbG4EAGxsbAAAAAAAN3BubAYwJCfb2AA3cG5sBjAkJ9vb2PwkAABsN'
cargo fuzz run -O fuzz_merger_idempotent oom.bin
```

Driving the same three phases directly (parse without validation → resolve → merge, exactly what the fuzz target does):

| phase | time |
|---|---|
| `ComponentParser::without_validation().parse` | **0.93 ms** |
| `Resolver::resolve` | **0.44 ms** |
| `Merger::merge` | **94.5 s**, then `Err` |

The parsed component has one core module of **48 bytes**. The merge eventually fails with `failed to parse component: invalid flags byte in element segment (at offset 0x4)` — so all of that time is spent *before* reaching an error the merger could have reached immediately.

## Not caused by #423

I hit this on the branch for #400 and checked whether v0.57.0's `use`-resolution change (#423) introduced it, since that change made several traversals follow an alias they previously stopped at. It did not:

| tree | time |
|---|---|
| this branch (post-#423) | 94.7 s |
| **v0.56.2 (`55427cf`, pre-#423)** | **93.5 s** |

Same input, same shape. The pathology is at least as old as v0.56.2; the fuzzer simply reached this input now.

## Why it matters

meld fuses untrusted-ish input: a component someone else built. A 60-byte input that costs 95 seconds and unbounded-looking allocation is a denial-of-service shape, and the time is spent on work that ends in a parse error regardless. It also makes `Fuzz Smoke` red for unrelated PRs, which trains people to ignore it.

## Where to look

The time is entirely inside `Merger::merge`, so the likely shape is a declared count being trusted before it is validated against the remaining bytes — a LEB-encoded length that decodes to something enormous driving an allocation or a loop. The input's tail (`f6 f6 f6 3f`, `dc 1b 9b 01`) is consistent with that reading, though I have not confirmed which count it is.

Fix direction: bound any count read from the input against the bytes that remain before allocating or iterating, and make the merger reach its parse error without doing the work first. A regression test can assert this input merges-or-fails in well under a second, which is the property that actually matters.

`Fuzz Smoke` is not a required context, so this is not blocking merges; PR #425 is going in with this failure recorded and explained.

avrabe:
Closes #426.

## 60 bytes, 95 seconds, and then an error it had all along

```
input                    60 bytes
  core module            48 bytes
  element section        9 bytes   ← declares 134,069,110 segments

parse                    0.93 ms
resolve                  0.44 ms
merge                    94.5 s  →  Err("invalid flags byte in element segment")
```

Under libFuzzer's 2 GB allocation limit that is an out-of-memory, which is how `fuzz_merger_idempotent` reported it.

## Located by sampling, not by reading

3608 of 3715 samples sat in `build_index_maps_for_module`, which sizes the per-module segment index map:

```rust
let elem_segment_count = crate::segments::count_element_segments(module);
for local in 0..elem_segment_count {
    maps.elements.insert(local, elem_segment_base + local);
}
```

`count_element_segments` reads the section header's count **without parsing the entries** — deliberately; its doc says so. That makes the number whatever the input claims, and the loop materialises one map entry per claimed index: 134 million of them.

## The fix

A segment costs at least one byte, so a declared count above the section length cannot describe real segments:

```rust
Ok(reader) => bound_by_section(reader.count(), start, end),
```

| | before | after |
|---|---|---|
| counted element segments | 134,069,110 | **9** |
| merge time | 94.8 s | **187 µs** |
| outcome | `Err(invalid flags byte …)` | `Err(invalid flags byte …)` |

**For a valid module this is a no-op** — its declared count never exceeds its bytes — and the malformed input is still rejected. Only the cost changed.

## Pattern, not instance

Every declared-count read in the workspace was classified, because the last three fixes of this shape corrected the line in front of me and left a sibling:

| site | verdict |
|---|---|
| `segments.rs` element counter | **the defect** |
| `segments.rs` data counter | same shape, same fix |
| `component_wrap.rs` `type_count` | stored on a struct, drives no loop or allocation |
| `component_wrap.rs` `memory_count` | inside a test assertion |

## Tests, and a control that fails

`meld-core/tests/bounded_segment_counts_426.rs` drives the fuzz finding itself, committed to `fuzz/corpus/fuzz_merger_idempotent/oom_426_element_count.bin` so the fuzzer replays it too:

- `a_declared_count_cannot_exceed_its_section` — guards that the input really has a 9-byte element section, then requires the count not to exceed it;
- `merging_a_hostile_count_stays_bounded` — requires the merge to finish inside 5 s **and still fail**, since rejecting a malformed module is correct.

**Control:** removing the clamp fails both. The timing test visibly hangs — `has been running for over 60 seconds` — and the run takes 95 s of user time. The failure message names the numbers: *"a 9-byte section cannot hold 134069110 segments"*.

## Traceability & gate

- SR-83 / SWV-95 via rivet; `rivet verify SR-83` (1 link + 2 markers); scoped `release: v0.58.2`; `rivet release status v0.58.2` cuttable
- `tools/check_record_citations.py`: OK, 426 citations resolve
- `cargo fmt --all -- --check` **0** · `clippy --workspace --all-targets -D warnings` **0** · `test --workspace` **0** · `test --workspace --all-features` **0**

Version 0.58.1 → **0.58.2**. No Tier-5 files touched (`segments.rs` is, but only the two counters and a new private helper — see the delta-pass comment I will add if the gate asks for one).


avrabe:
## Mythos delta-pass

Correcting the PR body first: **`segments.rs` is on the Tier-5 list**, so this change needs a pass — the sentence at the bottom of the description reads both ways and is wrong. The diff is the two counters, one new private helper, two tests, and a corpus input.

### The change, attacked

| question | answer |
|---|---|
| `end < start`? | `saturating_sub` → bound 0 → no segments mapped. Only reachable if the parser hands back an inverted range, which would be its own bug. |
| section longer than 4 GiB? | `u32::try_from(..).unwrap_or(u32::MAX)` → no clamp, which is the correct direction. Unreachable in practice. |
| does a valid module change? | No. Its declared count never exceeds its bytes, so `min` returns the declared count — verified by the full suite, including the `use`d-type, pack-rebase and multi-memory fixtures. |
| does the malformed input still fail? | Yes, with the **identical** error (`invalid flags byte in element segment (at offset 0x4)`). Only the cost changed: 94.8 s → 187 µs. The test asserts the failure as well as the bound, so a "fix" that made it succeed would fail the test. |
| what changes for a malformed module? | It gets fewer index-map entries, so indices beyond the bound fall through to the rewriter's identity remapping — which is the documented behaviour for an unmappable index, and the same path a missing section already takes. |

### A sibling I probed and did NOT change

`reloc.rs:504` has the same shape — `Vec::with_capacity(count as usize)` where `count` comes straight from an input `reloc.*` section, reachable through the public `parse_reloc_info`. I expected a second defect and measured instead:

```
reloc.CODE body of 6 bytes declaring 1_000_000_000 entries  → 70.9 µs, Err(unexpected end of reloc.CODE section)
reloc.CODE body of 6 bytes declaring 4_294_967_295 entries  → 238 µs, Err(unexpected end of reloc.CODE section)
```

The reservation is lazy and the loop errors on the first missing byte, so there is no real work and no resident memory. **Not a live defect, so not changed** — the difference from the segments case is that there the count drove 134 million *completed* map insertions, which is work rather than a reservation. Worth knowing that libFuzzer's malloc limit would still flag a single 68 GB reservation if a fuzz target ever reaches it; if that shows up, it is noise to bound then, not now.

### The rest of the pattern sweep

Every declared-count read in the workspace: the element counter (the defect), the data counter (same fix), `component_wrap`'s `type_count` (stored on a struct, drives no loop) and its `memory_count` (a test assertion). `provenance.rs:220` already clamps its SCPV entry count with `count.min(max_possible)` — the discipline this change extends, not invents.

### What I did not do

I did not make the counters *exact* by parsing the entries. They are documented as counting "without fully parsing each entry", the caller only needs an upper bound for sizing, and parsing would trade a cheap clamp for a cost proportional to the section on every merge.

PR #434, opened 2026-09-23
feat(v0.58.3): ship statically linked Linux binaries, and run what the release builds (SR-84, #432)

### LINKED ISSUE #432: Linux releases are glibc-only (2.34 floor) — ship musl builds too
## What was measured

`v0.58.2`'s Linux x86_64 release asset, downloaded from this repo and inspected on 2026-09-22:

```
$ file meld
ELF 64-bit LSB pie executable, x86-64, dynamically linked,
interpreter /lib64/ld-linux-x86-64.so.2

$ strings -a meld | grep -oE 'GLIBC_[0-9.]+' | sort -V | tail -1
GLIBC_2.34
```

So the binary is dynamically linked against glibc with a **2.34 floor**.

| Target | glibc | Runs this binary |
|---|---|---|
| Ubuntu 22.04 | 2.35 | ✓ |
| Debian 12 (bookworm) | 2.36 | ✓ |
| RHEL 9 / Amazon Linux 2023 | 2.34 | ✓ (exactly at the floor) |
| Ubuntu 20.04 | 2.31 | ✗ |
| RHEL 8 | 2.28 | ✗ |
| Alpine / distroless-static | musl / none | ✗ |

## Why this is being raised here

These binaries are ingested into the **varve `pulseengine` realm's rolling layer** — the pinned
toolchain an air-gapped or enterprise consumer installs. A layer's portability is the *maximum*
floor across its payloads, so one 2.39 payload sets the floor for the whole toolchain. Measured
across the realm on 2026-09-22:

| Repo | Release | glibc floor |
|---|---|---|
| rivet | v0.37.0 | 2.39 |
| spar | v0.40.0 | 2.39 |
| synth | v0.69.0 | 2.39 |
| witness | v0.43.0 | 2.39 |
| sigil (`wsc`) | v0.11.0 | 2.39 |
| varve | v0.37.0 | 2.39 |
| ordeal | v0.19.0 | 2.34 |
| loom | v1.4.1 | 2.34 |
| meld | v0.58.2 | 2.34 |
| kiln | v0.5.0 | 2.34 |

**No repo in the org publishes a musl asset.** The split between 2.34 and 2.39 is itself
unintended — it is which runner image the workflow happened to use, not a decision anyone made.

## What is being asked for

Add the two musl targets to this repo's release matrix, **alongside** the existing gnu ones:

```
x86_64-unknown-linux-musl
aarch64-unknown-linux-musl
```

Keep the existing asset naming (`meld-<tag>-<triple>.tar.gz`) — varve derives asset names from
the Rust triple, so musl assets are picked up by a one-line change in the realm manifest and
nothing else. A musl build is statically linked and runs on every Linux above, including Alpine
and `distroless-static`.

Caveats worth stating rather than discovering: musl's allocator is slower under heavy
multi-threaded allocation, and a static binary cannot use glibc NSS or `dlopen` host libraries.
For a CLI that reads files and writes reports, neither usually applies — if this tool does either,
say so on this issue and the answer for it is "lower the gnu floor" instead.

## The decision is the maintainer's

Three options, in the order I would pick them:

1. **musl beside gnu** (recommended) — both assets ship; consumers choose; nothing regresses.
2. **musl replacing gnu** — one artifact, maximal portability, loses NSS/`dlopen`.
3. **Keep gnu, lower the floor** — build on `ubuntu-22.04`, giving a 2.35 floor. Cheapest, still
   excludes Alpine and distroless-static.

Filed from the varve side, where the floor is visible; the call belongs to this repo.

avrabe:
Closes #432.

## The block

The Linux release assets were gnu-only. Read from the **shipped** v0.58.2
aarch64 asset:

```
PT_INTERP   : /lib/ld-linux-aarch64.so.1
PT_DYNAMIC  : True
GLIBC syms  : GLIBC_2.17 … GLIBC_2.34
```

`GLIBC_2.34` is the floor, and it excludes Alpine, distroless-static, RHEL 8
and Ubuntu 20.04. gale raised it from the varve side, where the consequence is
visible: a realm layer's portability is the **maximum** floor across its
payloads, so meld's floor is a property of the toolchain a consumer installs,
not of this repo alone.

`meld-<tag>-{x86_64,aarch64}-unknown-linux-musl.tar.gz` now ship beside the
gnu ones. Naming is unchanged, so a consumer deriving asset names from the
Rust triple picks them up with no other change.

The musl caveats were checked rather than assumed: no `dlopen` or `libloading`
anywhere in the tree, and no name resolution, socket or user lookup, so
neither NSS nor `dlopen` is reachable. Dependencies are clap, anyhow, log,
env_logger, serde_json, wasmparser — pure Rust.

## The larger thing this turned up

`release.yml` **never executed anything it built.** Build → strip → package →
upload. A binary that linked and then died on start-up would have shipped, and
for musl the whole portability claim was resting on the linker flags being
right.

Each asset is now checked as far as its runner allows:

| tier | covers | how |
|---|---|---|
| linkage | both musl targets, incl. cross-built aarch64 | `readelf -l` must show no `INTERP` |
| native execution | targets matching the runner's `uname -m` | `meld --version` |
| **no glibc at all** | x86_64 musl | run inside `alpine:3.20` |

The third is the only one that tests the requirement's own claim rather than a
proxy for it.

## Evidence, and the controls for it

Linking was proven locally after all — `zig cc` cross-links both targets:

| | new musl build | shipped v0.58.2 gnu asset (control) |
|---|---|---|
| `PT_INTERP` | none | `/lib/ld-linux-aarch64.so.1` |
| `PT_DYNAMIC` | false | true |
| `GLIBC_*` symbols | none | 8, up to `GLIBC_2.34` |

**Gate potency, measured:** the updated assets gate was dispatched against
v0.58.2, which carries no musl assets. The old gate
[passed](https://github.com/pulseengine/meld/actions/runs/35769109659) on that
tag; the new one
[failed](https://github.com/pulseengine/meld/actions/runs/35829465626) on it,
naming both missing archives. Same tag, same assets — only the gate changed.

**A guard that would have passed by absence:** `readelf -l "$BIN" | grep -q
INTERP` is *false* when `readelf` is missing, so the linkage check would have
silently passed on a runner without it. It now demands the tool first. Swept
the tree for the same shape: the other two `| grep -q` sites pipe from
`printf`, a builtin that cannot go missing, and both fail in the safe
direction.

**Drift:** what a release ships is decided by two independent lists — the
build matrix and the gate's required set — kept in sync by a comment asking a
human to remember. `tools/check_release_targets.py` makes that mechanical *on
the pull request*, because the gate only runs after publication: "built but
not required" was invisible, "required but not built" would be discovered in
public. Three negative controls observed to fail: drop a triple from either
list, or rename the loop so the list cannot be found.

All three branches of the smoke step were exercised locally against real
binaries: static musl → pass, the gnu asset posing as a musl one → fail with
the right message, `readelf` removed from `PATH` → fail rather than pass.

## Traceability

- **SR-84** — Linux release assets run on systems without glibc — `verified`,
  scoped `release: v0.58.3`
- **SWV-96** — records all four levels above, including what is *not* proven
- `rivet release status v0.58.3` → ✓ cuttable

## Local gate

`fmt 0` · `clippy 0` · `test --workspace 0` · `test --workspace --all-features 0`

## Not proven here

The aarch64 musl asset is never executed — no runner in the matrix can run it.
Its linkage is checked cross-architecture; its execution is not.


avrabe:
Pushed c3f02aa — two defects in the smoke check I added, both found by *running* it across all six triples instead of only the musl one.

**1. `uname -m` reports `arm64` on macOS; the Rust triple says `aarch64`.** So on `macos-14` the one darwin target that can run natively fell into the "cannot execute here" branch, while the PR body above claimed "native execution where the runner's architecture matches". My own local output printed `built on Darwin/arm64` and I read past it. Normalised.

The honest count, which the CHANGELOG now states instead of the general claim:

| executed | inspected only |
|---|---|
| `aarch64-apple-darwin` | `x86_64-apple-darwin` |
| `x86_64-unknown-linux-gnu` | `aarch64-unknown-linux-gnu` |
| `x86_64-unknown-linux-musl` (+ Alpine) | `aarch64-unknown-linux-musl` |

**2. The worse one — the check passed on a file that did not exist.** `readelf -l "$BIN" | grep -q INTERP` is false both when there is no interpreter *and* when readelf produced no output. Pointed at a deleted path, it printed `statically linked, no interpreter`.

The previous commit guarded against readelf being **missing** and left readelf **failing** wide open. That is the fix-the-instance mistake this repo keeps making, committed inside the fix for it. Output is now captured, a non-zero readelf is fatal, empty output is fatal, and the binary must exist before any of it runs. The `docker` guard had the identical shape and is demanded the same way.

Six branches re-exercised with real exit codes:

```
static musl                                     EXIT=0
missing binary                                  EXIT=1
dynamic binary posing as the musl asset         EXIT=1
darwin native execution                         EXIT=0
binary that runs but doesn't identify itself    EXIT=1
readelf absent                                  EXIT=1
```

Worth stating plainly: the first version of this check would have gone green in CI. It only failed when given something to fail on.

